PR #4, opened 2026-01-28
Fix hang on non-session commands

### LINKED ISSUE #3: pi-subagents causes hanging in non-interactive commands
With pi-subagents (0.3.2) installed, non-interactive commands like `pi --version` and `pi --help` print output but don’t exit (hang).

It looks like this is a process lifecycle management issue: the extension starts long-lived handles (e.g. `fs.watch(...)`, and in some cases `setInterval(...)` / `setTimeout(...)`) during extension registration/startup. Those are reaped at `session_shutdown`, but that doesn’t run (or doesn’t run soon enough) for non-interactive commands so the event loop stays alive

simple repro:
   - `pi --version` (hangs)
   - `pi --version --no-extensions` (exits normally)

happy to submit a PR, but wanted to ask preference:

1. minimal change: call `.unref()` on watcher/timers so they don’t keep the process alive
2. larger refactor: lazily start watcher/poller only after a session starts or after the subagent tool is first invoked

Love the extension by the way!

capdevc:
unref poller and watcher to prevent keeping the loop alive.

Fixes https://github.com/nicobailon/pi-subagents/issues/3

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -119,6 +119,7 @@ export default function registerSubagentExtension(pi: ExtensionAPI): void {
 
 			renderWidget(lastUiContext, Array.from(asyncJobs.values()));
 		}, POLL_INTERVAL_MS);
+		poller.unref?.();
 	};
 
 	const handleResult = (file: string) => {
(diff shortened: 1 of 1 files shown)


nicobailon:
nice fix, thanks for the contribution!

PR #10, opened 2026-02-10
Stream JSONL artifacts to disk to prevent OOM in parallel runs

ThewindMom:
## Problem
`pi-subagents` consumes a JSONL event stream from a spawned `pi --mode json` process. During long / tool-heavy runs (especially with frequent `message_update` deltas), this event stream can become extremely large (hundreds of MB).

Previously, the sync runner buffered *all* JSONL lines in memory (`jsonlLines.push(line)`) and only wrote the JSONL artifact at the end of the run. In `subagent parallel`, multiple large buffers accumulate in the parent Node process and can trigger V8 heap OOM:

> FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

## Root cause (first principles)
Artifact capture was effectively implemented as **“collect an unbounded stream in RAM, then write”**. That is fundamentally unsafe for unbounded/streaming workloads, and becomes much worse under parallelism.

## Fix
Stream JSONL artifacts to disk incrementally as lines arrive:
- Open a `WriteStream` for `{runId}_{agent}.jsonl` when JSONL artifacts are enabled.
- Write each JSONL line immediately (`write(line + "\n")`).
- Handle backpressure: pause `proc.stdout` when `write()` returns `false`, resume on `drain`.
- Close the stream cleanly after the subprocess exits.

## Impact
- Prevents Node/V8 OOM in long sync runs and `subagent parallel`.
- Keeps JSONL artifact content/format intact; only the capture strategy changes.
- No API changes.

## Files changed
- `execution.ts`: stream JSONL artifact capture to disk; remove in-memory buffering.
- `CHANGELOG.md`: document fix.
- `types.ts`: clarify comment about large JSONL capture.

### DIFF
--- a/execution.ts
+++ b/execution.ts
@@ -24,7 +24,6 @@ import {
 	writePrompt,
 	getFinalOutput,
 	findLatestSessionFile,
-	detectSubagentError,
 	extractToolArgsPreview,
 	extractTextFromContent,
 } from "./utils.js";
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks for the PR. Applied fix directly in [1050e57](https://github.com/nicobailon/pi-subagents/commit/1050e57682742eadec75ed202a21bcd3957d2a0c)

PR #13, opened 2026-02-11
fix: Windows spawn — prevent ENOENT and task string fragmentation

marcfargas:
## Problem

Two spawn bugs on Windows, both stemming from how `pi` is invoked as a child process:

### 1. ENOENT (original PR scope)

`spawn("pi", args)` without `shell: true` fails on Windows because `pi` is a shell wrapper (`.cmd`/`#!/bin/sh`). Node.js cannot execute it directly → silent ENOENT (exit code -4058).

### 2. Task string fragmentation (discovered during testing)

The naive fix (`shell: true`) introduces a worse bug: `cmd.exe` concatenates args **without quoting** ([Node.js DEP0190](https://nodejs.org/api/deprecations.html#DEP0190)). A task like:

```
"Read C:/dev/project/file.md and provide your architecture review."
```

arrives at the child agent as **separate user messages**: `"Task:"`, `"Read"`, `"C:/dev/..."`, `"and"`, `"provide"`, ... — each word is a new conversational turn. The agent goes off the rails.

**Affects all modes**: single, parallel, and chain (async). Metacharacters (`&`, `|`, `>`) can also cause hangs or command injection via cmd.exe.

## Root Cause

Both bugs trace to the same line — how `spawn` invokes `pi`:

| Approach | ENOENT? | Task splitting? | Metachar injection? |
|----------|---------|-----------------|---------------------|
| `spawn("pi", args)` | ❌ ENOENT | ✅ | ✅ |
| `spawn("pi", args, { shell: true })` | ✅ | ❌ Splits on spaces | ❌ `&` `\|` execute |
| **This fix**: `spawn(node, [cli.js, ...args])` | ✅ | ✅ | ✅ |

## Fix

**Primary path** (Windows): Bypass cmd.exe entirely by spawning `process.execPath` (node) with the pi CLI script (`process.argv[1]`). No shell involved, no escaping needed, args are passed as a proper argv array.

**Fallback** (if `process.argv[1]` is unavailable or doesn't point to a .js/.mjs file): Use shell with manually quoted args.

**Non-Windows**: Unchanged (`spawn("pi", args)` without shell — already works).

### Files Changed

| File | Scope | What |
|------|-------|------|
| `execution.ts` | Sync spawn (single + parallel mode) | Replace bare `spawn("pi")` with platform-aware spawn |
| `subagent-runner.ts` | Async spawn (chain runner) | Same fix — this file also had bare `spawn("pi")` which would ENOENT on Windows |

### Key Code (identical in both files)

```typescript
if (process.platform === "win32") {
    const piCli = process.argv[1];
    if (piCli && (piCli.endsWith(".js") || piCli.endsWith(".mjs")) && fs.existsSync(piCli)) {
        spawnCmd = process.execPath;      // node
        spawnArgs = [piCli, ...args];     // [cli.js, --mode, json, -p, ...]
        useShell = false;                 // no cmd.exe
    } else {
        // Fallback: shell with quoted args
        spawnCmd = "pi";
        spawnArgs = args.map((a) => (a.includes(" ") ? `"${a.replace(/"/g, '\\"')}"` : a));
        useShell = true;
    }
} else {
    spawnCmd = "pi";
    spawnArgs = args;
    useShell = false;
}
```

### Unit test (spawn behavior)

| Input | node+cli.js (fix) | shell+quoted (fallback) | shell+unquoted (bug) |
|-------|-------------------|------------------------|----------------------|
| Spaces in task | ✅ | ✅ | ❌ |
| Long multi-word | ✅ | ✅ | ❌ |
| Quotes in text | ✅ | ✅ | ❌ |
| `&` metachar | ✅ | ✅ | ❌ |
| `\|` metachar | ✅ | ✅ | ❌ |
| Parentheses | ✅ | ✅ | ❌ |
| `%` in text | ✅ | ✅ | ❌ |

### End-to-end test (via subagent tool)

Diagnostic agent (`echo-task`) that reports how many user messages it received:

| Mode | TURN_COUNT | Task intact? |
|------|-----------|-------------|
| Single | 1 ✅ | Full task received verbatim |
| Parallel (run 0) | 1 ✅ | Intact |
| Parallel (run 1) | 1 ✅ | Intact |

Tested on Windows 11, Git Bash, Scoop-installed Node.js + pi.

## `.js`/`.mjs` guard

The `process.argv[1]` check includes an extension guard to prevent spawning `node cli.ts` in tsx dev mode — only `.js` and `.mjs` entry points are used for the direct-node path.

### DIFF
--- a/execution.ts
+++ b/execution.ts
@@ -159,7 +159,30 @@ export async function runSync(
 	}
 
 	const exitCode = await new Promise<number>((resolve) => {
-		const proc = spawn("pi", args, { cwd: cwd ?? runtimeCwd, env: spawnEnv, stdio: ["ignore", "pipe", "pipe"] });
+		// On Windows, spawn("pi", args) fails with ENOENT because "pi" is a cmd/sh wrapper,
+		// and spawn("pi", args, { shell: true }) splits unquoted args on spaces (DEP0190),
+		// fragmenting task strings into multiple user messages.
+		// Fix: spawn node directly with the pi CLI script, bypassing cmd.exe entirely.
+		let spawnCmd: string;
+		let spawnArgs: string[];
+		let useShell: boolean;
+		if (process.platform === "win32") {
+			const piCli = process.argv[1];
+			if (piCli && (piCli.endsWith(".js") || piCli.endsWith(".mjs")) && fs.existsSync(piCli)) {
+				spawnCmd = process.execPath;
+				spawnArgs = [piCli, ...args];
+				useShell = false;
+			} else {
+				spawnCmd = "pi";
+				spawnArgs = args.map((a) => (a.includes(" ") ? `"${a.replace(/"/g, '\\"')}"` : a));
+				useShell = true;
+			}
+		} else {
+			spawnCmd = "pi";
+			spawnArgs = args;
+			useShell = false;
+		}
+		const proc = spawn(spawnCmd, spawnArgs, { cwd: cwd ?? runtimeCwd, env: spawnEnv, stdio: ["ignore", "pipe", "pipe"], shell: useShell });
 		let buf = "";
 
 		// Throttled update mechanism - consolidates all updates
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks! Resolved in [5c9823a](https://github.com/nicobailon/pi-subagents/commit/5c9823a8cf7944be72b2c734f618156f0cd24402)

PR #27, opened 2026-02-17
fix: write long tasks to temp file to avoid ENAMETOOLONG on Windows

marcfargas:
## Problem

When running a chain with a parallel step followed by a synthesizer, the synthesizer's task
includes `{previous}` — the aggregated output of all parallel reviewers. For a typical
4-reviewer design review, this can be 50-100KB+ of text.

This entire task string is passed as a single positional CLI argument to `spawn()`:

```typescript
args.push(\);
```

On Windows (Git Bash / MSYS2), this exceeds OS argument limits and `spawn` fails with:

```
ENAMETOOLONG
```

The error surfaces as chain step 2 (synthesizer) failing to start.

## Fix

When the task string exceeds 8,000 characters, write it to a temp file and pass it via
pi's `@file` syntax instead of as a positional CLI argument:

```typescript
const TASK_ARG_LIMIT = 8000;
if (task.length > TASK_ARG_LIMIT) {
    // write to tmpDir/task.md, pass as @filepath
} else {
    args.push(\);
}
```

- Reuses the existing `tmpDir` lifecycle (created by `writePrompt` for system prompt,
  cleaned up after process exits)
- Creates `tmpDir` if it doesn't already exist (when there's no system prompt)
- The `@file` mechanism is pi's built-in file argument processing — reads the file and
  injects its content as the agent's first user message
- **Zero impact on short tasks** — they continue using the direct positional argument

## Testing

- Verified existing tests still pass (`npm test`)
- Tested with echo-task agent on Windows to confirm task delivery works correctly
- The fix targets the exact scenario: chain step 2 with large `{previous}` content

### DIFF
--- a/execution.ts
+++ b/execution.ts
@@ -4,6 +4,8 @@
 
 import { spawn } from "node:child_process";
 import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
 import type { Message } from "@mariozechner/pi-ai";
 import type { AgentConfig } from "./agents.js";
 import {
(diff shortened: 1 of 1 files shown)


marcfargas:
Closing — this fix is now included as commit 1 of #28 (quality hardening PR).

PR #33, opened 2026-02-22
fix(render): fix ANSI background bleed in text truncation

monotykamary:
## Problem

Text truncation in the subagent result rendering was breaking the TUI background color. The ellipsis characters were appearing without proper ANSI styling, causing background color corruption in the pi interface.

| Before | After |
|--------|------|
| image | image |

## Root Cause

The  function from pi-tui adds an ANSI reset code () before its ellipsis. When this reset was applied, it stripped the active color styling from the ellipsis character, causing it to render without the proper background color.

## Solution

Rewrote  to manually handle ANSI codes:
- Parse through the text character by character
- Track active ANSI styling codes (excluding reset codes)
- Truncate to leave room for ...
- Append ellipsis with the active style applied

Also added proper pre-truncation for:
- Task descriptions (single and chain modes)
- Current tool arguments
- Recent tool arguments

These use terminal width-aware limits to prevent double truncation.

## Changes

- : Complete rewrite with ANSI-aware truncation
- Task preview truncation: Consistent pattern with pre-truncation
- Tool args truncation: Width-aware limits
- Recent output: Let truncLine handle truncation entirely

## Testing

Tested with parallel subagent runs showing long task descriptions. The ellipsis now properly inherits the dim color styling without breaking the TUI background.

### DIFF
--- a/render.ts
+++ b/render.ts
@@ -22,7 +22,45 @@ function getTermWidth(): number {
 
 function truncLine(text: string, maxWidth: number): string {
 	if (visibleWidth(text) <= maxWidth) return text;
-	return truncateToWidth(text, maxWidth - 1) + "…";
+	// truncateToWidth adds \x1b[0m before its ellipsis which breaks TUI background
+	// Instead: manually truncate to leave room for "..." and append it with proper styling
+	const targetWidth = maxWidth - 3;
+	let result = "";
+	let currentWidth = 0;
+	let activeStyle = ""; // Track current style (excluding reset codes)
+	let i = 0;
+
+	while (i < text.length) {
+		// Check for ANSI code
+		const ansiMatch = text.slice(i).match(/^\x1b\[[0-9;]*m/);
+		if (ansiMatch) {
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks for identifying this issue @monotykamary! The screenshots clearly showed the background bleed problem.

I've implemented a fix in 038528b that addresses the same issue with a slightly different approach:

1. **Unicode handling** — Uses `Intl.Segmenter` for proper grapheme segmentation (handles emoji and multi-codepoint characters correctly)

2. **ANSI style stacking** — Tracks all active styles in an array rather than just the last one, so stacked styles like `theme.bold(theme.fg("color", text))` are fully preserved

3. **Pre-truncation improvements** — Incorporated your idea of width-aware pre-truncation for tool args and recent output

The core insight was spot-on: pi-tui's `truncateToWidth` adds `\x1b[0m` before the ellipsis which resets styling. The fix re-applies all tracked styles before appending the ellipsis.

Credited you in the CHANGELOG. Thanks again for the contribution!

PR #50, opened 2026-03-09
fix: resolve skill discovery from settings.packages and cwd

### LINKED ISSUE #48: bug: skill resolution misses settings.packages and task cwd
## Bug

Subagent skill lookup has two related gaps:

1. **`settings.packages` skills ignored** — when resolving skills for a task, only built-in skill paths are searched; package skills declared in project or user `settings.json` under `packages` are never consulted, producing false-negative "Skills not found" errors even when the skill is properly installed.

2. **Wrong cwd for skill lookup** — resolution uses the runtime process cwd instead of the effective task cwd (the `cwd` field on the dispatched task), so relative skill paths and package roots resolve against the wrong directory.

## Expected behavior

- Skill resolution should walk all package roots from both user and project `settings.json` (the same set used at runtime).
- Skill path resolution should be performed relative to the task's effective `cwd`, not the agent process cwd.

## Fix

A fix is available on branch `lajarre:pr/fix-skill-resolution-packages`.

lajarre:
Summary
- include skills declared by local package sources in project and user settings packages
- include cwd package.json pi.skills in discovery
- resolve skills from effective execution cwd (sync and async)
- add runtime cwd fallback for unresolved skills
- fix skill source inference precedence for user/project overlap
- simplify related helper code and test fixtures

Testing
- npm run -s test:integration
- npm run -s test:all

Closes #48

### DIFF
--- a/async-execution.ts
+++ b/async-execution.ts
@@ -15,7 +15,7 @@ import { injectSingleOutputInstruction, resolveSingleOutputPath } from "./single
 import { isParallelStep, resolveStepBehavior, type ChainStep, type ParallelStep, type SequentialStep, type StepOverrides } from "./settings.js";
 import type { RunnerStep } from "./parallel-utils.js";
 import { resolvePiPackageRoot } from "./pi-spawn.js";
-import { buildSkillInjection, normalizeSkillInput, resolveSkills } from "./skills.js";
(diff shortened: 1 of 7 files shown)


nicobailon:
Thanks — I used this as the starting point, rebased the fix onto current `main`, and shipped it there. Closing this stale PR now.

PR #54, opened 2026-03-12
fix: include username in temp directory names to prevent collisions

### LINKED ISSUE #53: Temporary paths collide between users
I use Pi with two different user accounts on the same server and get this:
```
[Extension issues]
  npm:pi-subagents (user) index.ts
    Failed to load extension: EACCES: permission denied, access '/tmp/pi-async-subagent-results'
```

Including the user login name in the path would solve this issue.

akaihola:
Fixes #53

When `pi-subagents` is installed for multiple user accounts on the same machine, both users compete for the same `/tmp/pi-async-subagent-*` directories. The second user to start Pi gets an `EACCES` error because the directory was created by the first user.

Using `os.userInfo().username` makes the paths user-scoped:
- `/tmp/pi-async-subagent-results-<username>`
- `/tmp/pi-async-subagent-runs-<username>`

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -261,8 +261,8 @@ export const DEFAULT_ARTIFACT_CONFIG: ArtifactConfig = {
 
 export const MAX_PARALLEL = 8;
 export const MAX_CONCURRENCY = 4;
-export const RESULTS_DIR = path.join(os.tmpdir(), "pi-async-subagent-results");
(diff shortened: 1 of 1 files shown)


akaihola:
@nicobailon, I rebased this fix on v0.12.0

nicobailon:
Thanks — used this as the starting point and landed the broader fix on main in 22a1b551.

PR #56, opened 2026-03-16
feat: add `.gitignore`

MariaSolOs:
It's a good practice to include a `.gitignore` in open source projects in order to avoid OS-specific files (like `.DS_Store`) being committed to source control.


nicobailon:
Thanks Maria Solano — good catch. I landed the equivalent fix on current `main` by keeping the repo's existing `.gitignore` entries, adding `.DS_Store`, and removing the tracked macOS metadata file. Closing this stale PR now.

Issue #51, opened 2026-03-10
bug: async subagent fails silently with relative cwd (5ms exit code 1)


tburman:
## Bug

Any async subagent call with a relative `cwd` fails in ~5ms with exit code 1 and **zero output**:

```typescript
subagent({
  async: true,
  agent: "web-researcher",
  cwd: "projects/my-project",   // relative CWD triggers the bug
  task: "..."
})
```

Status JSON shows:
```json
{
  "state": "failed",
  "steps": [{ "status": "failed", "durationMs": 5, "exitCode": 1 }]
}
```

Output log is 0 bytes. No error message anywhere.

Sync calls with the same `cwd` work fine.

## Root Cause

**Path**: `async-execution.ts` → `spawnRunner()` → `subagent-runner.ts` → `runPiStreaming()`

The async execution path passes a **relative CWD** to two levels of process spawning:

1. `executeAsyncSingle` sets `runnerCwd = cwd ?? ctx.cwd` (e.g., `"projects/my-project"` — relative)
2. `spawnRunner()` spawns the runner process with `{ cwd: runnerCwd }` — Node resolves this against the **main pi process CWD** → correct absolute path (e.g., `/Users/.../doc-maker/projects/my-project`)
3. The runner config also contains `step.cwd: "projects/my-project"` (same relative string)
4. Inside the runner, `runSingleStep()` calls `runPiStreaming(args, step.cwd ?? ctx.cwd, ...)` — Node resolves this relative CWD against the **runner's CWD** → `/Users/.../doc-maker/projects/my-project/projects/my-project`
5. **That path doesn't exist** → pi process fails immediately → exit code 1, zero output

### Diagram

```
Main pi process (CWD: /Users/.../doc-maker)
  │
  ├─ SYNC: spawn("pi", [...], { cwd: "projects/my-project" })
  │   → resolves to /Users/.../doc-maker/projects/my-project  ✅
  │
  └─ ASYNC: spawnRunner({ cwd: "projects/my-project" })
       → runner CWD = /Users/.../doc-maker/projects/my-project  ✅
       │
       └─ runner: spawn("pi", [...], { cwd: "projects/my-project" })
            → resolves to /Users/.../doc-maker/projects/my-project/projects/my-project  ❌ ENOENT
```

## Why It's Hard to Debug

1. **5ms failure** — too fast to look like a real execution
2. **Zero output** — pi dies before writing anything (CWD doesn't exist)
[...]

PR #62, opened 2026-03-25
fix: reuse current node and pi cli for async subagent spawns

mjh-sakh:
## Summary

Fix async subprocess spawning so background subagents reuse the current Node executable and resolved pi CLI path instead of re-resolving `node` / `pi` from PATH.

Fixes #61.
Related: #51, #13.

## Motivation

Async execution currently has two PATH-dependent launch points:

1. `async-execution.ts` spawns the detached runner with bare `node`
2. `pi-spawn.ts` uses the resolved CLI script only on Windows, but falls back to bare `pi` on macOS/Linux

That means a parent pi session launched through a pinned wrapper like:

```bash
mise exec node@25.8.0 -- pi
```

can run foreground work correctly while async subagents drift onto a different Node / pi runtime. In practice this shows up as confusing async-only startup failures with provider/tool differences, while sync execution in the same cwd still works.

## Changes

- `async-execution.ts`
  - use `process.execPath` instead of bare `node` when spawning the detached async runner
- `pi-spawn.ts`
  - prefer the resolved current pi CLI script on **all platforms**, not only Windows
  - keep the existing `pi` fallback when resolution fails
- `pi-spawn.test.ts`
  - add non-Windows coverage for the resolved-CLI path
  - keep fallback coverage for unresolved CLI lookup

## Why this is safe

- `process.execPath` is the exact Node binary running the parent session
- CLI resolution already guards against `.ts` entrypoints via `isRunnableNodeScript()` and falls back to package/bin lookup
- if CLI resolution fails, behavior remains unchanged (`pi` from PATH)

## Validation

Ran locally:

- `npm test`
- `npm run test:integration`

Also validated the real-world async scenario that motivated the fix:
- async `planner` ✅
- async `reviewer` ✅
- async `scout` ✅

where those runs had previously been failing under a `mise exec node@... -- pi` parent launch.

### DIFF
--- a/async-execution.ts
+++ b/async-execution.ts
@@ -106,7 +106,7 @@ function spawnRunner(cfg: object, suffix: string, cwd: string): number | undefin
 	fs.writeFileSync(cfgPath, JSON.stringify(cfg));
 	const runner = path.join(path.dirname(fileURLToPath(import.meta.url)), "subagent-runner.ts");
 	
-	const proc = spawn("node", [jitiCliPath, runner, cfgPath], {
+	const proc = spawn(process.execPath, [jitiCliPath, runner, cfgPath], {
 		cwd,
 		detached: true,
 		stdio: "ignore",
--- a/pi-spawn.test.ts
+++ b/pi-spawn.test.ts
@@ -33,7 +33,21 @@ function makeDeps(input: {
 }
 
 describe("getPiSpawnCommand", () => {
-	it("uses plain pi command on non-Windows", () => {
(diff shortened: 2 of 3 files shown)


nicobailon:
I replayed this onto current `main`, kept the core fix, added the relevant current-tree test coverage, reran the verification, and merged the updated version from the maintainer side.

Closing this PR because the underlying change is now landed, but not via this original branch. Thanks for the patch — the fix direction was right and it made it in after being applied to the current codebase.

PR #67, opened 2026-03-31
fix(runtime): support direct TypeScript loading

StartupBros:
## Summary
- switch internal relative imports from `.js` to `.ts` for source-loaded runtime compatibility
- replace Node strip-types incompatible parameter properties with explicit field assignments
- declare the Pi core peer dependencies the package imports

## Why
Pi can dogfood packages directly from a local checkout. In that mode there is no compiled `.js` tree, and Node's strip-types loader rejects parameter properties. This PR makes the package load cleanly from source as well as after publish/build.

## Verification
- `pnpm test:all`
- `node --input-type=module -e "import("./index.ts")"`

### DIFF
--- a/agent-management.ts
+++ b/agent-management.ts
@@ -9,11 +9,11 @@ import {
 	type ChainConfig,
 	type ChainStepConfig,
 	discoverAgentsAll,
-} from "./agents.js";
-import { serializeAgent } from "./agent-serializer.js";
-import { serializeChain } from "./chain-serializer.js";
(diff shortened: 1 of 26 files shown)


nicobailon:
I rebased the runtime-loading changes onto current `main`, reran the relevant verification on the rebased branch, and merged the updated branch into `main` from the maintainer side.

Closing this PR because the underlying change is now landed, but not via this original branch. Thanks for the patch — the core idea made it in after being replayed onto the current codebase.

PR #70, opened 2026-04-03
feat: inherit parent session provider for bare model ids

gcunharodrigues:
## Problem

   When agent configs use bare model IDs like `claude-sonnet-4-5`, subagents do not consistently
 inherit the parent session provider.

   In practice, this can make subagent execution depend on registry order or generic CLI model
 resolution, and it also forces users to hardcode `provider/model` in every agent definition.

   ## Change

   This PR centralizes model resolution and makes bare model IDs default to the parent session
 provider while preserving the original model name.

   Examples:
   - agent model: `claude-sonnet-4-5`
   - parent session provider: `github-copilot`
   - resolved subagent model: `github-copilot/claude-sonnet-4-5`

   Explicit fully-qualified model IDs such as `anthropic/claude-sonnet-4-5` are preserved as-is.

   The resolution is now applied consistently across:
   - single execution
   - parallel execution
   - chain execution
   - clarify/TUI flow
   - async/background execution

   ## Why this is better

   - keeps subagents aligned with the parent session provider by default
   - preserves agent-authored model names
   - reduces config duplication in agent files
   - makes agents more portable across providers
   - removes duplicated model resolution logic
   - makes behavior consistent across execution modes

   ## Implementation notes

   - added a centralized `model-resolution.ts`
   - switched to `--model` when a full `provider/model` ID is available
   - kept `--models` as a fallback when only a bare model ID is available
   - added required provider extension wiring for multi-pass providers

   ## Validation

   - `npm run test:unit`
   - `npm run test:integration`
   - `npm run test:e2e`

### DIFF
--- a/async-execution.ts
+++ b/async-execution.ts
@@ -11,6 +11,7 @@ import { createRequire } from "node:module";
 import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
 import type { AgentConfig } from "./agents.js";
 import { applyThinkingSuffix } from "./pi-args.js";
+import { resolveModelFullId } from "./model-resolution.js";
 import { injectSingleOutputInstruction, resolveSingleOutputPath } from "./single-output.js";
 import { isParallelStep, resolveStepBehavior, type ChainStep, type SequentialStep, type StepOverrides } from "./settings.js";
 import type { RunnerStep } from "./parallel-utils.js";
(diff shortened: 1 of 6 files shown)


gcunharodrigues:
Main design intent for this PR:

   - agents keep their existing model ids
   - the parent session provides the default provider for bare model ids
   - explicit `provider/model` ids remain untouched

   This makes agent configs more portable across providers while keeping subagent execution aligned
 with the active parent session.

nicobailon:
Thanks Guilherme — landed on main with the current implementation, so I’m closing this one.

Issue #34, opened 2026-02-24
Feature request: disable built in agents through config


Ninja3047:
It would be really nice if there was a way to disable specific built in agents through some kind of local config so I don't have to maintain a local fork of the extension.

nicobailon:
Thanks! This has been addressed in v0.17.0.

PR #73, opened 2026-04-07
feat: subagents.json overlay for overriding and disabling agents without editing .md files

danielcherubini:
## Summary

Adds a non-destructive overlay config system for tweaking agent frontmatter (`model`, `thinking`, `skills`, `defaultProgress`, etc.) and disabling built-in or user agents without editing any `.md` files on disk. Configuration is read from up to four sources, in order of increasing precedence (later wins, per field):

1. User `~/.pi/agent/settings.json` — `subagents` key
2. Project `.pi/settings.json` — `subagents` key
3. User `~/.pi/agent/subagents.json`
4. Project `.pi/subagents.json`

`disabled` and `disableBuiltins` are unioned across all four. The dedicated `subagents.json` files override the embedded `subagents` key in `settings.json`, so users with an existing `settings.json` can keep everything in one file or split it out.

### Highlights

- **`subagents.json` loader** (new `subagents-config.ts`) with full validation, allowlisted overridable fields, and warnings-never-throws semantics.
- **Field-level overrides** applied in `loadAgentsFromDir` after parsing, so every downstream consumer (agent-management, chain resolution, runtime, model warnings) automatically sees merged config.
- **Per-name `disabled` list** — filtered post-merge in `discoverAgents`/`discoverAgentsAll`.
- **`disableBuiltins: true`** — hides all agents whose `source === "builtin"` in one shot, while leaving user/project agents alone (even if they share a name with a built-in).
- **Embedded `subagents` key in `settings.json`** — same shape, lower precedence than the dedicated file.
- **`overlayWarnings`** added to `discoverAgentsAll` return value, surfaced via `slash-commands.ts` notify when non-empty.
- **README docs + `docs/examples/subagents.example.jsonc`** sample.

### Schema

```jsonc
{
  "agents": {
    "scout": { "model": "anthropic/claude-opus-4-6", "thinking": "high" },
    "reviewer": { "skills": ["code-review", "security-audit"] }
  },
  "disabled": ["delegate", "worker"],
  "disableBuiltins": true
}
```

Overridable fields (allowlist): `description`, `model`, `thinking`, `tools`, `skills`, `extensions`, `output`, `defaultReads`, `defaultProgress`, `interactive`, `maxSubagentDepth`. `name`, `systemPrompt`, and internal metadata are intentionally not overridable — for body changes, use the existing `~/.agents/`/`<project>/.agents/` user-override path.

### Test isolation

The loader honors `PI_SUBAGENTS_HOME` to redirect both `subagents.json` and `settings.json` to a tempdir during tests. The harness was added to all unit tests that touch `discoverAgents`/`discoverAgentsAll` so the contributor's real `~/.pi/agent/` is never read.

## Test plan

- [x] `npm run test:unit` — 179/179 passing
- [x] Tests cover all edge cases: `defaultProgress: false` override, project-wins-per-field merge, disabled+override combo, unknown agent name warnings, invalid JSON, layered precedence across all four sources, `disableBuiltins` union semantics
- [x] Manually verified by symlinking the branch into `~/.pi/agent/extensions/subagent` and using it via real `settings.json` overrides
- [ ] Reviewer to spot-check that the new field on `discoverAgentsAll`'s return type is consumed correctly downstream
- [ ] Reviewer to consider whether the `slash-commands.ts` notify of warnings is the right surfacing point (currently shows count only — can be expanded to a richer panel later)

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,6 +428,56 @@ Skills are specialized instructions loaded from SKILL.md files and injected into
 
 **Missing skills:** If a skill cannot be found, execution continues with a warning shown in the result summary.
 
+## Configuring agents with `subagents.json`
+
+You can override built-in agent fields and disable agents from loading
+without editing the bundled `.md` files. pi-subagents reads two files:
+
+- **Project**: `.pi/subagents.json` (relative to your working directory)
+- **User**: `~/.pi/agent/subagents.json`
+
+Project values take precedence over user values, field-by-field.
+The `disabled` lists from both files are unioned.
+
+### Example
+
+```jsonc
+{
+  "agents": {
+    "scout": {
+      "model": "anthropic/claude-opus-4-6",
+      "thinking": "high"
+    },
+    "reviewer": {
+      "description": "Custom reviewer for the auth subsystem",
+      "skills": ["code-review", "security-audit"]
+    }
+  },
+  "disabled": ["delegate", "worker"]
+}
+```
+
+### Overridable fields
+
+`description`, `model`, `thinking`, `tools`, `skills`, `extensions`,
+`output`, `defaultReads`, `defaultProgress`, `interactive`,
+`maxSubagentDepth`.
+
+The agent's `name`, `systemPrompt` (body), and internal metadata
+cannot be overridden. To replace the system prompt, write a real
+agent file under `~/.agents/` or `<project>/.agents/` (existing
+override path).
+
+### Disabling agents
+
+Names in `disabled` are skipped entirely — they will not appear in
+`/agents list`, are not invokable from chains or directly, and produce
+the standard "unknown agent" error if referenced.
+
+Validation problems (unknown fields, invalid JSON, references to
+non-existent agents) are surfaced as warnings via `discoverAgentsAll`,
+visible in agent-management output.
+
 ## Usage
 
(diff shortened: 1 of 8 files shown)


danielcherubini:
Fixes #34

nicobailon:
I think we should skip this. It adds a whole second customization system for something we already handle pretty cleanly with user/project agent files. The payoff feels low, while the cost is more precedence rules, more docs/tests, and
 more edge cases to maintain. It’s basically extra config complexity to avoid editing files that are already easy to override.

flaviovs:
> (...) something we already handle pretty cleanly with user/project agent files. (...)

@nicobailon could you please explain how we achieve this, i.e. "disable a built-in agent"?

danielcherubini:
I'm going to explain this a bit more.

Having a centralized spot to override certain parts of your agents files is something that is in other harnesses..  but one of the reasons people like Pi, is how minimal it is, and how configurable it is.

So, on that, being able to open up my settings.json and having a single place where i can swap certain things around, like which model for what skill, without duplicating the skill, in the project, just to change the model.. has value.

Finally, whilst its nice that you include some subagents, its also nice to give people a way to remove the built in models, either selectively or all at once. 

Its all about giving people flexibility. 

So what you might see as having a low payoff, I think most others would see as having a high value. Choice, Customization, Flexiblity.

PR #76, opened 2026-04-12
Expose stable chainId and chainDir on chain execution events

gryyny:
## Summary
Adds optional `chainId` and `chainDir` fields to the `subagent:slash:started`
and `subagent:slash:response` event payloads so downstream tooling can
deterministically bind to the active chain run instead of scanning tempdirs.

## Problem
Callers today can infer the active chain directory only by scanning
`<tmpdir>/pi-chain-runs/*` for recent artifacts, which:
- races on concurrent runs
- cannot distinguish between overlapping bestOfN / consensus flows
- forces downstream tooling into filesystem locks and `unverified` fallbacks
  to avoid silent-wrong-answer modes

## Change
- Reuse the existing slash `requestId` as a stable `chainId`
- Add a pure `resolveChainDirPath(runId, baseDir?)` helper so the bridge can
  derive the shared artifact directory without side effects
- Include both as **optional** fields on the existing started/response payloads
- Document the new fields in the README

## Backwards compatibility
Both new fields are marked optional. No breaking changes to the existing
event surface, types, or handler contracts.

## Tests
- New integration test asserts `started` and `response` events carry `chainId`
  and `chainDir`, and that `chainDir` points at a real directory after
  execution
- `npm run test:unit`
- `npm run test:integration`

## Notes
- `npm run test:e2e` currently fails on upstream `main` in this environment due
  a test harness mismatch (`session.agent.setTools is not a function`); this PR
  does not change the e2e surface.
- `package.json` declares `license: MIT`, though there is no separate `LICENSE`
  file in the repo root.

## Why now
Downstream consumers currently ship workaround code (filesystem locks,
ambiguity rejection, explicit `unverified` gate outcomes) that exists only to
paper over this gap. Exposing a stable identifier removes the need for that
consumer-side complexity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -826,6 +826,8 @@ Each chain run creates `<tmpdir>/pi-chain-runs/{runId}/` containing:
 
 Directories older than 24 hours are cleaned up on extension startup.
 
+For slash-command driven chain runs, `subagent:slash:started` and `subagent:slash:response` now include optional `chainId` and `chainDir` fields. `chainId` reuses the stable request/run identifier, and `chainDir` points at the shared chain artifacts directory so downstream tooling can bind to the active chain deterministically.
+
 ## Artifacts
 
 Location: `{sessionDir}/subagent-artifacts/` or `<tmpdir>/pi-subagent-artifacts/`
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks Andrii, this has been addressed.

PR #81, opened 2026-04-15
feat: support disabling builtin agents via settings.json

### LINKED ISSUE #34: Feature request: disable built in agents through config

danielcherubini:
Extends the existing `subagents.agentOverrides` schema with a `disabled` boolean so users can hide a specific builtin agent, and adds a `disableBuiltins` bulk flag at the `subagents` object level to switch every builtin off at once:
```json
  {
    "subagents": {
      "disableBuiltins": true,
      "agentOverrides": {
        "reviewer": { "disabled": true },
        "scout":    { "model": "anthropic/claude-opus-4-6" }
      }
    }
  }
```
`discoverAgents` filters disabled agents out of the runtime list. `discoverAgentsAll` keeps them (with `disabled: true` on the config and `override.scope` attribution) so management UIs can surface and re-enable them. A per-agent override without `disabled` opts the agent out of the bulk `disableBuiltins` switch. Project scope wins over user scope for the bulk flag, so a project can re-enable builtins locally even when the user's global settings disable them.

Motivation: builtin agents ship inside the package, so users can't delete the .md files. Previously there was no way to hide a builtin you don't want to use without touching the package itself.

Closes #34

### DIFF
--- a/agent-manager.ts
+++ b/agent-manager.ts
@@ -115,7 +115,7 @@ export class AgentManagerComponent implements Component {
 
 	private loadEntries(): void {
 		const overridden = new Set([...this.agentData.user, ...this.agentData.project].map((c) => c.name));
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks! This has been addressed in v0.17.0.

PR #90, opened 2026-04-16
fix: make Ctrl+O expand subagent results with full details

asagajda:
## Problem

Two issues with the Ctrl+O expand behavior for subagent results:

1. **`expanded` flag ignored** — `renderSubagentResult()` received it as `_options` but never used it. Tool calls, task text, and progress args were always identically truncated in both views.

2. **Tool calls lost on completion** — `compactForegroundResult()` sets `messages: undefined` to save memory after a subagent finishes, destroying all tool call data. The renderer had nothing to show.

## Changes

- **`types.ts`**: Add `ToolCallRecord[]` — lightweight array of `{name, args}` preserved across compaction.
- **`utils.ts`**: `extractToolCalls()` extracts tool calls from messages before compaction. `compactForegroundResult()` stores them in `toolCalls`.
- **`formatters.ts`**: `formatToolCall()` accepts `expanded` param — skips truncation of bash commands (60→full) and JSON args (40→full) when expanded.
- **`render.ts`**: 
  - Introduce `trunc()` helper that skips `truncLine` when expanded
  - Tool calls only rendered in expanded view (collapsed = compact summary per upstream intent)
  - Task text, progress args all untruncated when expanded

## Result

| View | Content |
|------|---------|
| **Collapsed** (default) | Agent name, truncated task, tool count, usage stats |
| **Expanded** (Ctrl+O) | Full task text, all tool calls with full commands, all details |

## Testing

All 188 existing unit tests pass.

### DIFF
--- a/formatters.ts
+++ b/formatters.ts
@@ -86,10 +86,13 @@ export function buildChainSummary(
 /**
  * Format a tool call for display
  */
-export function formatToolCall(name: string, args: Record<string, unknown>): string {
+export function formatToolCall(name: string, args: Record<string, unknown>, expanded = false): string {
 	switch (name) {
-		case "bash":
-			return `$ ${((args.command as string) || "").slice(0, 60)}${(args.command as string)?.length > 60 ? "..." : ""}`;
+		case "bash": {
+			const cmd = (args.command as string) || "";
+			if (expanded) return `$ ${cmd}`;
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks! This has been addressed directly on main.

PR #98, opened 2026-04-20
fix: harden async subagent startup

### LINKED ISSUE #99: bug: background async launch can report false success
Background async launch can report success even when no detached runner starts.

Reproduction cases:
- The resolved async cwd does not exist.
- The detached child process cannot produce a usable runner process after `spawn()` returns.

Current behavior:
- The caller can receive a successful launch response.
- No usable background run exists.

Expected behavior:
- Validate the resolved async cwd before spawning the detached runner.
- Return a tool error when launch fails.
- Log detached child-process errors for diagnostics.

SeanThomasWilliams:
## Summary
- validate the async cwd before spawning the detached runner
- return a tool error when startup fails
- attach a detached child error handler so spawn failures stay contained
- add regression coverage for missing cwd and spawn failure paths

Closes #99.

## Testing
- `node --experimental-transform-types --import ./test/support/register-loader.mjs --test --test-name-pattern='returns a tool error when an async run uses a missing cwd|returns a tool error when the async runner process cannot spawn' test/integration/async-execution.test.ts`

### DIFF
--- a/async-execution.ts
+++ b/async-execution.ts
@@ -114,22 +114,39 @@ export function isAsyncAvailable(): boolean {
 /**
  * Spawn the async runner process
  */
-function spawnRunner(cfg: object, suffix: string, cwd: string): number | undefined {
-	if (!jitiCliPath) return undefined;
-	
+function spawnRunner(cfg: object, suffix: string, cwd: string): { pid?: number; error?: string } {
+	if (!jitiCliPath) {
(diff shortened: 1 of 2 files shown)


SeanThomasWilliams:
Closing this PR and replacing it with https://github.com/nicobailon/pi-subagents/pull/107. The old branch shared the same commit that also appeared in #102, which made the upstream linkage messy. The replacement PR is a clean branch with only the #99 change.

PR #102, opened 2026-04-20
fix: improve async subagent reliability

### LINKED ISSUE #100: bug: async widget can stay stale and show zero tokens
Async background runs can leave stale entries in the "Async subagents" widget after completion. Completed runs can also show `0 tok` even when the session log recorded usage.

Reproduction:
1. Launch a background run.
2. Wait for completion.
3. Check the widget above the composer.

Current behavior:
- The widget entry can remain visible after the retention window.
- The same entry can show `0 tok` even when the session log contains usage.

Expected behavior:
- Completed runs disappear on time.
- Token totals match the recorded async session usage.

### LINKED ISSUE #101: feat: force top-level subagent runs async from config
`asyncByDefault` only changes the default when a call omits `async`. A top-level call can still force foreground execution with `async: false`, and a top-level chain can still stay foreground when clarify is enabled.

Requested behavior:
Add a config flag that forces depth-0 subagent runs into background mode.

Example:
```json
{
  "forceTopLevelAsync": true
}
```

Expected semantics:
- Top-level single, parallel, and chain runs use `async: true`.
- Top-level clarify is bypassed with `clarify: false`.
- Nested runs keep their normal behavior.

SeanThomasWilliams:
## Summary
- fix stale async widget entries by forcing a TUI rerender and scheduling cleanup when polling detects completion
- fix async token accounting by reading usage from session message entries and the newest session file by mtime
- add `forceTopLevelAsync` for automation setups that never want top-level subagent runs to block
- apply the top-level async override before validation and context/session resolution

## Issues
Closes #100.
Closes #101.

## Dependency
Independent of #98 and #103.

### Stale widget / zero tokens
1. Enable background async subagents.
2. Launch a background run.
3. Wait for completion.
4. Observe that the widget can stay on screen past the retention window and can show `0 tok` even when the session log recorded usage.

### Forced top-level async
1. Set `forceTopLevelAsync: true` in `~/.pi/agent/extensions/subagent/config.json`.
2. Launch a top-level single, parallel, or chain subagent run.
3. Observe that the run is forced into background mode and top-level clarify is bypassed.
4. For chains, validation now runs against the forced-async params before launch.

## Testing
- `git diff --check`
- `node --experimental-transform-types --import ./test/support/register-loader.mjs --test --test-force-exit test/integration/async-job-tracker.test.ts test/integration/session-tokens.test.ts test/integration/top-level-async.test.ts`
- `NODE_PATH=<jiti shim> node --experimental-transform-types --import ./test/support/register-loader.mjs --test --test-force-exit --test-name-pattern='returns a tool error when an async run uses a missing cwd|returns a tool error when the async runner process cannot spawn' test/integration/async-execution.test.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -855,6 +855,33 @@ This aggregated output becomes `{previous}` for the next step.
 
 `pi-subagents` reads optional JSON config from `~/.pi/agent/extensions/subagent/config.json`.
 
+### `asyncByDefault`
+
+`asyncByDefault` makes top-level subagent calls use background execution when the request does not explicitly set `async`.
+
+```json
+{
+  "asyncByDefault": true
+}
+```
+
+This only changes the default. Callers can still force foreground execution by setting `async: false` unless `forceTopLevelAsync` is also enabled.
+
+### `forceTopLevelAsync`
+
+`forceTopLevelAsync` forces depth-0 subagent execution into background mode. This is useful for automation setups that never want the top-level orchestrator to block on child runs.
+
+```json
+{
+  "forceTopLevelAsync": true
+}
+```
+
+When enabled:
+- top-level single, parallel, and chain runs are forced to `async: true`
+- top-level clarify UI is bypassed by forcing `clarify: false`
+- nested subagent calls still follow their own inherited depth and async settings
+
 ### `parallel`
 
 `parallel` controls top-level `tasks` mode defaults and limits.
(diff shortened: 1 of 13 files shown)


SeanThomasWilliams:
Closing this PR. It mixed the #100 widget/token fixes, the #101 config feature, and the #99 startup commit on one branch. Replacement PRs: #105 for #100 and #106 for #101. PR #98 still covers #99.

PR #105, opened 2026-04-21
fix: async widget state and token totals

### LINKED ISSUE #100: bug: async widget can stay stale and show zero tokens

SeanThomasWilliams:
## Summary
- rerender the async widget when cleanup state changes
- schedule cleanup when polling discovers a completed run
- read async token usage from `message.usage`
- prefer the newest session file by mtime when multiple session files exist

Closes #100.

## Testing
- `node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/async-job-tracker.test.ts test/integration/session-tokens.test.ts`

### DIFF
--- a/async-job-tracker.ts
+++ b/async-job-tracker.ts
@@ -7,18 +7,42 @@ import {
 } from "./types.ts";
 import { readStatus } from "./utils.ts";
 
-export function createAsyncJobTracker(state: SubagentState, asyncDirRoot: string): {
+interface AsyncJobTrackerOptions {
+	completionRetentionMs?: number;
+	pollIntervalMs?: number;
+}
+
+export function createAsyncJobTracker(state: SubagentState, asyncDirRoot: string, options: AsyncJobTrackerOptions = {}): {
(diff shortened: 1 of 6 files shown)


nicobailon (approved):
Thanks Sean — this looks good. I verified the issue against the PR, checked the touched async widget/session-token paths, and ran the targeted regression coverage plus adjacent async execution coverage. The cleanup rerender path and the `message.usage` token fix both line up with the bug report.

nicobailon:
Follow-up: commit 68aa0ef9 fixes a related async token gap (missing totals when detached runs lack session files).

PR #106, opened 2026-04-21
feat: force top-level async mode

### LINKED ISSUE #101: feat: force top-level subagent runs async from config

SeanThomasWilliams:
## Summary
- add `forceTopLevelAsync` to the extension config
- force depth-0 runs to `async: true` and `clarify: false`
- apply the override before validation and fork/session setup
- add focused coverage for the override helper

Closes #101.

## Testing
- `node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/top-level-async.test.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -855,6 +855,33 @@ This aggregated output becomes `{previous}` for the next step.
 
 `pi-subagents` reads optional JSON config from `~/.pi/agent/extensions/subagent/config.json`.
 
+### `asyncByDefault`
+
+`asyncByDefault` makes top-level subagent calls use background execution when the request does not explicitly set `async`.
+
+```json
+{
+  "asyncByDefault": true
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks Sean — merged.

Issue #110, opened 2026-04-22
Extension breaks pi agent, when used with llama.cpp


daniel-oelert:
I am trying to use a locally hosted LLM for low-cost mini-subagents. For this I have added a local model with in the pi models.json. When I try to send any message, I get the following error:

```
Error: 400 JSON schema conversion failed:                                                                                                                                                                                                                                   
 Unrecognized schema: {"description":"Agent or chain config for create/update. Agent: name, description, scope ('user'|'project', default 'user'), systemPrompt, systemPromptMode, inheritProjectContext, inheritSkills, model, tools (comma-separated), extensions          
 (comma-separated), skills (comma-separated), thinking, output, reads, progress, maxSubagentDepth. Chain: name, description, scope, steps (array of {agent, task?, output?, reads?, model?, skills?, progress?}). Presence of 'steps' creates a chain instead of an agent."} 
 Unrecognized schema: {"description":"Skill name(s) to inject (comma-separated), array of strings, or boolean (false disables, true uses default)"}                                                                                                                          
 Unrecognized schema: {"description":"Chain step: either {agent, task?, ...} for sequential or {parallel: [...]} for concurrent execution"}                                                                                                                                  
 Unrecognized schema: {"description":"Output file for single agent (string), or false to disable. Relative paths resolve against cwd."}                                                                                                                                      
 Unrecognized schema: {"description":"Skill name(s) to inject (comma-separated), array of strings, or boolean (false disables, true uses default)"}                                                                                                                          
 Unrecognized schema: {"description":"Agent or chain config for create/update. Agent: name, description, scope ('user'|'project', default 'user'), systemPrompt, systemPromptMode, inheritProjectContext, inheritSkills, model, tools (comma-separated), extensions          
 (comma-separated), skills (comma-separated), thinking, output, reads, progress, maxSubagentDepth. Chain: name, description, scope, steps (array of {agent, task?, output?, reads?, model?, skills?, progress?}). Presence of 'steps' creates a chain instead of an agent."} 
 Unrecognized schema: {"description":"Skill name(s) to inject (comma-separated), array of strings, or boolean (false disables, true uses default)"}                                                                                                                          
 Unrecognized schema: {"description":"Chain step: either {agent, task?, ...} for sequential or {parallel: [...]} for concurrent execution"}                                                                                                                                  
 Unrecognized schema: {"description":"Output file for single agent (string), or false to disable. Relative paths resolve against cwd."}                                                                                                                                      
[...]

PR #115, opened 2026-04-25
fix: Fixed schema for #110

daniel-oelert:
Someone mind testing this with different APIs? I am lacking in the subscription department.


nicobailon:
Thanks, already fixed in commit 08b1a59c

PR #119, opened 2026-04-27
fix: treat final drain after assistant output as success

codewithkenzo:
## Summary

- Treat forced runner drain after a valid final assistant response as cleanup success instead of task failure.
- Keep forced drain as failure when the final assistant response has no usable output or reports an error.
- Add sync and async/background regression coverage using the mock Pi fixture.

## Why

#96 added final-message drain logic so stuck child processes do not hang the runner indefinitely. There is one remaining edge case: if the child emits a valid final assistant response and then stays alive until the drain timer terminates it, the forced termination can be reported as a task failure even though the final output was already delivered.

Example shape:

```text
assistant message_end: stopReason="stop", text="matrix-ok"
stderr: Done after 1 turn(s). Ready for input.
subagent.step.failed exitCode:1
```

That can incorrectly mark a successful run as failed and can stop chains after a valid handoff.

## Fix

The sync and async runners now latch whether a final assistant response has already been semantically delivered:

- assistant `message_end`
- `stopReason === "stop"`
- no `toolCall` block
- no `errorMessage`
- non-empty final text

If the final-drain timer later terminates the child, that termination is treated as process cleanup (`exitCode: 0`) only when that valid final response was already delivered. Empty/error final outputs still fail.

## Tests

```bash
env -u PI_SUBAGENT_DEPTH -u PI_SUBAGENT_MAX_DEPTH node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/single-execution.test.ts --test-name-pattern='forced drain'
```

Passes: sync forced-drain success and empty-output failure regressions.

```bash
env -u PI_SUBAGENT_DEPTH -u PI_SUBAGENT_MAX_DEPTH node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/async-execution.test.ts --test-name-pattern='forced drain'
```

Passes: async/background forced-drain success regression.

```bash
node --experimental-strip-types --check execution.ts
node --experimental-strip-types --check subagent-runner.ts
git diff --check
```

Passes.

### DIFF
--- a/execution.ts
+++ b/execution.ts
@@ -212,6 +212,8 @@ async function runSingleAttempt(
 		const HARD_KILL_MS = 3000;
 		let childExited = false;
 		let forcedTerminationSignal = false;
+		let finalAssistantDelivered = false;
+		let finalDrainCleanupWarning: string | undefined;
 		let finalDrainTimer: NodeJS.Timeout | undefined;
 		let finalHardKillTimer: NodeJS.Timeout | undefined;
 		const clearFinalDrainTimers = () => {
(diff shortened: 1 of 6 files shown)


nicobailon:
Applied the fix directly on main in 27a677b3. The sync and async runners now treat forced final-drain termination after valid final assistant output as cleanup success, with regression coverage for sync, async, and empty-output failure cases.

PR #120, opened 2026-04-27
Fix status row overflow on multiline content

salemsayed:
## Summary
- Make shared status row rendering normalize embedded newlines and tabs before rendering
- Truncate row content to the available inner width before padding
- Add regression coverage for clipped, multiline, tabbed, and ANSI-styled row content

## Validation
- `node --experimental-strip-types --test test/unit/render-helpers.test.ts`

Note: A full local `npm test` run also exercised this new test, but currently reports two unrelated existing schema test failures in `test/unit/schemas.test.ts`.


nicobailon:
Thanks!

PR #125, opened 2026-04-28
docs: list /parallel-cleanup alongside other packaged prompts

chenxin-yan:
## Summary

The `/parallel-cleanup` prompt was added in 0.20.0 but never picked up
in the user-facing prompt lists. Wiring is unchanged — the prompt
ships and registers via `package.json` `pi.prompts` directory scan
identically to its peers — so this is a pure documentation fix that
syncs the three sites that enumerate packaged prompts.

## Changes

- `README.md` — add `/parallel-cleanup` row to the prompt table next to `/parallel-review`, `/parallel-research`, and `/gather-context-and-clarify`.
- `skills/pi-subagents/SKILL.md` — add it to the Tool vs Slash Commands bullet list.
- `skills/pi-subagents/SKILL.md` — mention it in the Prompt Template Integration paragraph so the prose stays in sync with the bullet list.

## Test plan

- Docs-only change; no code paths touched.
- Verified `package.json` `pi.prompts` already scans `./prompts` so `/parallel-cleanup` was already discoverable at runtime.
- Verified no test enumerates prompt names by grepping `test/` for all four prompt names — no test updates required.
- Visually confirmed the three updated sites list the four prompts consistently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,7 @@ The package includes reusable prompt templates for common workflows. You do not 
 | `/parallel-review` | Launch fresh-context reviewers with distinct angles, then synthesize what to fix. |
 | `/parallel-research` | Combine `researcher` and `scout` for external evidence, local code context, and practical tradeoffs. |
 | `/gather-context-and-clarify` | Scout/research first, then ask the user the clarification questions that matter. |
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks! I've addressed this on main.

PR #131, opened 2026-05-01
Localize subagent slash flow strings

jerryfan:
I found one focused UX issue: the slash-command path has dense delegation/status text exactly where users choose between direct, chain, parallel, background, and forked subagent runs. If those messages are misread, users can launch the wrong orchestration shape or miss the live-detail/status path.

This PR makes that path optionally localizable while preserving the current English strings as fallbacks.

- No new dependency
- No behavior change without an i18n provider
- English remains the default/fallback
- Locales: ja, zh-CN, es — broad developer workflow locales for delegation/status copy
- Covers slash command descriptions, usage/errors, run status, and result headings

Validation:
- npm pack --dry-run
- npx -y node@22 --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/slash-commands.test.ts

Note: this machine's default Node does not support --experimental-strip-types, so npm test cannot run directly here. With node@22, the full unit suite has existing environment failures unrelated to this change: missing optional @mariozechner/pi-tui for TUI tests and Windows path expectations in pi-args tests.

### DIFF
--- a/src/extension/i18n.ts
+++ b/src/extension/i18n.ts
@@ -1,0 +1,126 @@
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
+
+type Params = Record<string, string | number>;
+type Translate = (key: string, fallback: string, params?: Params) => string;
+
+let translate: Translate = (_key, fallback, params) => format(fallback, params);
+
+function format(text: string, params?: Params): string {
+	if (!params) return text;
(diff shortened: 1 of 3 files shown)


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

PR #133, opened 2026-05-01
fix: report actual subagent result model

### LINKED ISSUE #72: BUG: `result.model` in `SingleResult` always displays the requested/preferred model rather than the actual model the CLI used.
## Root cause

In `execution.ts`, `result.model` is pre-filled with `modelArg` (the preferred model from agent definition or override) at
initialization:

```ts
const result: SingleResult = {
    // ...
    model: modelArg,   // <- set to preferred model upfront
};
```

Later, the actual model from the CLI response is only written when `result.model` is false:

```ts
if (!result.model && evt.message.model) result.model = evt.message.model;
```

Since `result.model` is already set to the preferred value, the actual model from `evt.message.model` is **never** captured. This means:

- The usage summary line (e.g. `3 turns in:12k out:422 claude-sonnet-4-6`) always shows what was *requested*, not what was *used*.
- If the API falls back to a different model (e.g. due to rate limits, restrictions, or provider routing), the mismatch is completely
invisible.
- If no model is specified in the agent definition, `modelArg` is `undefined`, and the actual model *does* get captured — so the bug only
 manifests when a model is explicitly configured.

## Affected files

-  `execution.ts:96` — initializes `result.model` to `modelArg`
- `execution.ts:197` — guard `!result.model` prevents overwrite
- `render.ts:268` / `formatters.ts:21` — display `result.model` in summary, showing stale value

## Suggested fix

1. Don't pre-fill `result.model` with the preferred value — leave it `undefined`.
2. Always update `result.model` from `evt.message.model` (drop the `!result.model` guard).
3. Add a separate `preferredModel` field to `SingleResult` to track what was requested.
4. Surface a mismatch in the usage summary when `model !== preferredModel`.

leohenon:
Fixes #72.

`SingleResult.model` now uses the model reported by the child CLI when available, instead of the requested model. The requested model for the final attempt is kept separately as `preferredModel`.

If the child does not report a model, `model` falls back to the requested model to preserve existing behavior.

Expanded rendering now shows mismatches as:

```text
actual-model (requested preferred-model)
```

Thinking suffixes are ignored for that comparison, so `model` and `model:high` do not render as different requested/actual models.

## Tests

```sh
npm test
npm run test:integration
```

### DIFF
--- a/src/runs/background/subagent-runner.ts
+++ b/src/runs/background/subagent-runner.ts
@@ -101,6 +101,7 @@ interface StepResult {
 	skipped?: boolean;
 	intercomTarget?: string;
 	model?: string;
+	preferredModel?: string;
 	attemptedModels?: string[];
 	modelAttempts?: ModelAttempt[];
 	artifactPaths?: ArtifactPaths;
@@ -550,6 +551,7 @@ async function runSingleStep(
 	exitCode: number | null;
 	error?: string;
 	model?: string;
+	preferredModel?: string;
 	attemptedModels?: string[];
 	modelAttempts?: ModelAttempt[];
 	artifactPaths?: ArtifactPaths;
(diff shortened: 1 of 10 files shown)


nicobailon:

Thanks for the model-reporting fix. This display accuracy issue is not high or critical priority in the current pass, and the PR is stale/conflicted, so I’m closing it with #72.

Issue #136, opened 2026-05-02
Intercom bridge always inactive — checks wrong path for pi-intercom


cad0p:
## Bug

The intercom bridge between parent and child subagents is always inactive. The `subagent({action:"doctor"})` report shows:

```
Intercom bridge
- bridge: inactive (pi-intercom extension was not found)
- pi-intercom: unavailable at ~/.pi/agent/extensions/pi-intercom
```

Child intercom targets in status always show `(if registered)`, and parent↔child messaging fails with "Session not found."

## Root cause

In `src/intercom/intercom-bridge.ts`, `defaultIntercomExtensionDir()` hardcodes:

```typescript
function defaultIntercomExtensionDir(): string {
    return path.join(os.homedir(), ".pi", "agent", "extensions", "pi-intercom");
}
```

Then `diagnoseIntercomBridge()` checks:

```typescript
const piIntercomAvailable = fs.existsSync(extensionDir);  // line 137
```

Pi-intercom is installed as an **npm package** (via `pi install npm:pi-intercom`), not as a directory at `~/.pi/agent/extensions/pi-intercom`. The package lives at `<node_modules>/pi-intercom/` and registers via `package.json`:

```json
{ "pi": { "extensions": ["./index.ts"], "skills": ["./skills"] } }
```

Since the filesystem path doesn't exist, `fs.existsSync` returns `false` and the bridge stays permanently inactive, regardless of pi-intercom being installed and working in the parent session.

## Repro

1. `pi install npm:pi-subagents && pi install npm:pi-intercom`
2. Launch any subagent with intercom in tools
3. Run `subagent({action:"doctor"})` → bridge inactive
4. Try `intercom({action:"ask", to:"<child-target>"})` → "Session not found"
5. Child tries `intercom({action:"ask", to:"<parent>"})` → targets stale session name, message lost

## Impact

- Parent cannot steer or help children mid-flight
- Children blocked on decisions have no way to reach the parent
- Async subagents are unmonitorable except via polling status
- The entire intercom coordination pattern described in the README is non-functional for most users

## Proposed fix

Instead of checking a hardcoded filesystem path, check whether pi-intercom is available as a registered pi package. Options:

**A) Check the settings.json packages list** — pi-intercom appears in `packages: ["npm:pi-intercom"]`. The bridge could scan the resolved package paths for `pi-intercom`.

[...]

PR #137, opened 2026-05-02
fix: check pi-intercom availability via pi.getAllTools() instead of filesystem path

cad0p:
## Problem

The intercom bridge checks `fs.existsSync(~/.pi/agent/extensions/pi-intercom)` to determine if pi-intercom is available. But pi-intercom installed via `pi install npm:pi-intercom` (the standard method) lives in node_modules and never creates that directory. Bridge permanently inactive for all npm package users.

**Repro:**
1. `pi install npm:pi-subagents && pi install npm:pi-intercom`
2. Launch any subagent with intercom in tools
3. `subagent({action:"doctor"})` → `bridge: inactive (pi-intercom extension was not found)`
4. Child targets show `(if registered)` — never appear in session list

**Before:**
```
Intercom bridge
- bridge: inactive (pi-intercom extension was not found)
- pi-intercom: unavailable at ~/.pi/agent/extensions/pi-intercom
```

**After:**
```
Intercom bridge
- bridge: active
- pi-intercom: available
```

## Root cause

`defaultIntercomExtensionDir()` returns a hardcoded filesystem path. Both `resolveIntercomBridge` and `diagnoseIntercomBridge` check `fs.existsSync(extensionDir)`. Only works for file-based extensions, not npm packages.

## Fix

Use the pi SDK's `getAllTools()` to check if the `intercom` tool is registered — the authoritative signal that pi-intercom is loaded. A `hasIntercom` callback is passed from the executor:

```typescript
// Before: hardcoded filesystem path
const piIntercomAvailable = fs.existsSync(extensionDir);

// After: check via pi SDK
hasIntercom: () => deps.pi.getAllTools().some((t) => t.name === "intercom")
```

## Files

- `src/intercom/intercom-bridge.ts` — `hasIntercom` callback replaces `fs.existsSync` check
- `src/runs/foreground/subagent-executor.ts` — passes `hasIntercom` callback at 2 call sites
- `src/extension/doctor.ts` — diagnostic shows `available` vs `not loaded`

### DIFF
--- a/src/extension/doctor.ts
+++ b/src/extension/doctor.ts
@@ -25,6 +25,7 @@ interface DoctorDeps {
 	discoverAgentsAll: typeof discoverAgentsAll;
 	discoverAvailableSkills: typeof discoverAvailableSkills;
 	diagnoseIntercomBridge: typeof diagnoseIntercomBridge;
+	getAllTools: () => { name: string }[];
 }
 
 interface DoctorReportInput {
@@ -54,6 +55,7 @@ const DEFAULT_DEPS: DoctorDeps = {
 	discoverAgentsAll,
 	discoverAvailableSkills,
 	diagnoseIntercomBridge,
+	getAllTools: () => [],
 };
 
 function errorText(error: unknown): string {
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for digging into this. The underlying issue has now been fixed on main and released in v0.23.0 via #136 / commit b9b3013e. The current implementation detects configured npm-installed `pi-intercom` packages while keeping the legacy local extension fallback, so this PR is now superseded. The `getAllTools()` idea is still useful as a possible future doctor diagnostic, but I don't want to swap the bridge activation contract to parent-runtime tool presence.

cad0p:
Ok nice, thanks for fixing!

PR #144, opened 2026-05-04
feat: apply agentOverrides to user agents + fix disableBuiltins priority

smantzavinos:
## Summary

Two related changes that improve the `agentOverrides` settings mechanism:

### 1. `agentOverrides` now applies to user/project agents (not just builtins)

Previously, `agentOverrides` in settings only modified builtin agents. User agents (from `~/.pi/agent/agents/`) were loaded as-is with no way to override their model/thinking per-repo without copying the full `.md` file.

Now `agentOverrides` applies to user agents too, enabling per-repo model configuration:

```json
// .pi/settings.json (project)
{
  "subagents": {
    "agentOverrides": {
      "worker": { "model": "anthropic/claude-sonnet-4", "thinking": "high" }
    }
  }
}
```

**Priority order:**
1. Project `.pi/agents/worker.md` (full file, always wins)
2. Project `.pi/settings.json` `agentOverrides.worker`
3. User `~/.pi/agent/settings.json` `agentOverrides.worker`
4. User `~/.pi/agent/agents/worker.md` `model:` field (base definition)

### 2. `disableBuiltins` now takes priority over `agentOverrides` for builtins

Previously, if `disableBuiltins: true` was set but `agentOverrides` had an entry for a builtin (e.g., `worker: { model: "..." }`), the override would apply and the builtin remained visible — defeating the purpose of `disableBuiltins`.

Now: when `disableBuiltins: true` is set, builtins are disabled even if they have `agentOverrides` entries. Use `disabled: false` in an override to explicitly opt a specific builtin out of bulk disable.

## Use case

Users who define their own agents with the same names as builtins (`worker`, `planner`, etc.) want:
- `disableBuiltins: true` to hide builtins from listings
- `agentOverrides` to set per-repo models on their *user* agents
- Agent prompts/skills to stay centralized (no full file copies per repo)

Without this change, `agentOverrides` only touches builtins, and having overrides with builtin names keeps those builtins visible even with `disableBuiltins: true`.

## Tests

- 7 new tests covering user agent overrides and disableBuiltins + agentOverrides interaction
- All 377 existing tests pass (clean environment)
- Updated 1 existing test to match new behavior (explicit `disabled: false` required to opt out of bulk disable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Added
+- `agentOverrides` in settings now applies to user and project agents, not just builtins. This enables per-repo model/thinking overrides without copying full agent `.md` files. Project `.pi/agents/` files still take highest priority.
+
+### Fixed
+- `disableBuiltins: true` now correctly hides builtins even when `agentOverrides` entries exist for them. Previously, any override for a builtin name would implicitly keep it visible. Use `disabled: false` in an override to explicitly opt a specific builtin out of bulk disable.
+
 ## [0.24.0] - 2026-05-03
 
 ### Changed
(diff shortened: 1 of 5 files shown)


nicobailon:

Closing this draft as superseded by the focused current PR #219 for applying `agentOverrides` to user/project custom agents.

PR #146, opened 2026-05-05
fix: disable single-run output when override is false

### LINKED ISSUE #145: Bug: `output: false` in single-agent runs can create a literal `false` output file
## Context

In repeated real-world use of `pi-subagents`, review-only subagent runs are being launched with `output: false` to suppress artifact files. That usage is documented in the tool contract and is even encouraged in package prompt content such as `prompts/parallel-cleanup.md`.

## Problem

In at least some single-agent runs, `output: false` is being converted into a literal output path named `false`, and the child receives an injected instruction like:

```text
---
**Output:** Write your findings to: /path/to/repo/false
```

That produces accidental repository files named `false` instead of disabling output.

This has happened repeatedly across repositories. In one recent run, a custom `reviewer` agent had frontmatter `output: pi-agent-review.md`, but the caller passed `output: false` to disable output. Runtime override precedence is expected, but the override was mishandled and became `/Users/andrea/Documents/GitHub/pi-manager/false` instead of disabling output.

The issue is particularly confusing because:
- the `subagent` tool schema says `output: false` disables output;
- chain/parallel behavior already has normalization logic for string/boolean false;
- package prompt content currently recommends `output: false` for reviewer tasks;
- the resulting injected instruction makes the child do exactly the wrong thing.

## Reproduction

A minimal reproduction should look roughly like this:

1. Define or use any single agent, for example `reviewer`.
2. Launch it in single-agent mode with `output: false`.
3. Observe the child task text or final repository state.

Expected:
- no output file is configured;
- no `Write your findings to: ...` instruction is injected;
- no file named `false` is created.

Actual:
- the child can receive `Write your findings to: /repo/false`;
- a file literally named `false` can be created in the working directory.

## Key Files

| File | Why it matters |
|------|---------------|
| `src/shared/settings.ts` | Normalizes `output` overrides for chain/parallel behavior and already treats string `"false"` defensively. |
| `src/runs/shared/single-output.ts` | Injects the single-run `Write your findings to: ...` instruction and resolves output paths. |
| `prompts/parallel-cleanup.md` | Currently recommends `output: false` for reviewer tasks, making this path easy to trigger. |
| `test/integration/template-resolution.test.ts` | Covers `false` disabling behavior in chain/template resolution, which suggests single-mode coverage may be missing. |
| `test/integration/parallel-execution.test.ts` | Includes a defensive test for string `false` in top-level parallel runs. |

## Constraints & Considerations

- The fix should preserve the documented meaning of `output: false`: disable output.
- Runtime `output` overrides should still take precedence over agent frontmatter defaults.
- The fix should cover both boolean `false` and any stringified `"false"` that may leak through parsing layers.
- It would be good to audit all single-agent code paths, not just chain/parallel ones, because the latter already appear more defensive.
- Package prompts that recommend `output: false` should be safe to follow.

## Open Questions

- Is the bug specific to single-agent execution, or can it appear in additional caller paths that eventually use single-run output injection?
- Is the boolean being coerced to a string before `resolveSingleOutputPath(...)`, or is the wrong value being passed into the injected-task layer later?
- Are there existing integration tests for single-agent `output: false`, and if not, should one be added alongside the fix?
- Should `injectSingleOutputInstruction(...)` or its caller defensively reject any non-path falsey/boolean sentinel before building task text?

## Expected Outcome

`output: false` should reliably disable output in single-agent runs, never resolve to a filesystem path, never inject `Write your findings to: ...`, and never create a file named `false`.

Su3h7aM:
## Summary

This PR fixes a bug where `output: false` could be mishandled in single-agent execution and end up behaving like a literal output path instead of disabling file output.

In the broken path, a single run could inject an instruction such as `Write your findings to: /repo/false`, which in turn could create an unwanted file literally named `false`.

Fixes #145.

## What changed

- normalize single-run output settings so stringified `"false"` is treated the same as disabled output;
- apply the normalized value when resolving output paths for foreground single-agent runs;
- keep override precedence intact when a caller explicitly disables an agent's default output file;
- refactor single-output helpers slightly to centralize normalization and IO error formatting;
- add regression coverage for unit and async single-run paths.

## Why this matters

The tool contract already documents `output: false` as the way to disable output. Parallel and template-related paths were already defensive here, but single-run behavior could still leak a string `"false"` through parsing layers.

This change makes single-agent behavior consistent with the documented contract and prevents accidental repository artifacts.

## Testing

- added unit coverage for `resolveSingleOutputPath("false", ...)`;
- added async integration coverage to confirm single-agent background runs do not write `false` or the agent default output when output is disabled.

### DIFF
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -35,7 +35,7 @@ import { createForkContextResolver } from "../../shared/fork-context.ts";
 import { resolveCurrentSessionId } from "../../shared/session-identity.ts";
 import { applyIntercomBridgeToAgent, INTERCOM_BRIDGE_MARKER, resolveIntercomBridge, resolveIntercomSessionTarget, resolveSubagentIntercomTarget, type IntercomBridgeState } from "../../intercom/intercom-bridge.ts";
 import { formatControlIntercomMessage, formatControlNoticeMessage, resolveControlConfig, shouldNotifyControlEvent } from "../shared/subagent-control.ts";
-import { finalizeSingleOutput, injectSingleOutputInstruction, resolveSingleOutputPath, validateFileOnlyOutputMode } from "../shared/single-output.ts";
+import { finalizeSingleOutput, injectSingleOutputInstruction, normalizeOutputSetting, resolveSingleOutputPath, validateFileOnlyOutputMode } from "../shared/single-output.ts";
 import { compactForegroundDetails, getSingleResultOutput, mapConcurrent, readStatus, resolveChildCwd } from "../../shared/utils.ts";
 import {
 	buildSubagentResultIntercomPayload,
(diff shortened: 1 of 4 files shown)


nicobailon:
Landed the single-run output false fix on main in c13713e0 (`fix: normalize string false output overrides`). Closing this PR as superseded by the mainline commit.

PR #148, opened 2026-05-05
Fix slash result finalization and add subagents admin command

dt-benedict:
## Summary

- Fix slash-launched subagent result cards staying visually stuck in the running state by finalizing the hidden result snapshot before clearing the live status/redrawing.
- Keep slash-command final results visible in non-UI contexts, while preserving live visible cards + hidden final snapshots in TUI mode.
- Add `/subagents` as the primary administration command for the Subagents Manager; `/agents` remains as an alias.
- Extend the manager detail screen with metadata and a direct `m` shortcut to update user/project subagent models via the existing model picker.
- Refresh `package-lock.json` to match current package metadata/dev dependency versions so `npm ci` works cleanly.

## Testing

```bash
npm run test:all
```

Result:

```text
# tests 212
# pass 212
# fail 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+- Added `/subagents` as the primary administration command for the Subagents Manager. The existing `/agents` command remains as an alias.
+- Agent detail screens now expose metadata fields and a direct `m` shortcut that opens the model picker and saves the selected model for user/project agents.
+
 ## [0.11.12] - 2026-03-28
 
 ### Changed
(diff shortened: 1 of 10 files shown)


dt-benedict:
Added one more fix after noticing configured agent models can be ignored for forked subagent sessions: subagent launches now use `--model` instead of `--models`, because `--models` only scopes model cycling and does not override a restored session model. Added regression coverage in `pi-args.test.ts`; full `npm run test:all` passes (214 tests).

dt-benedict:
Closing this PR because current main has since absorbed the two bug fixes (slash result finalization and --model subagent selection) and removed the old manager overlay, which caused conflicts. I'll open a fresh smaller PR against current main for the remaining /subagents administration command adapted to the new architecture.

PR #149, opened 2026-05-05
Add lightweight /subagents admin command

dt-benedict:
## Summary

- Adds `/subagents` as a lightweight administration command on top of current `main`.
- Lets users inspect selected subagent metadata directly in the conversation.
- In interactive/RPC UI contexts, lets users update the selected subagent model from Pi's available model list.
- Saves user/project agents back to their markdown files; saves builtin model changes through settings-backed `subagents.agentOverrides`.

This intentionally avoids reintroducing the removed manager overlay; it uses Pi's existing dialog primitives instead.

## Usage

- `/subagents` — select a subagent, view metadata, optionally change model.
- `/subagents <agent>` — show metadata for a named subagent.
- `/subagents <agent> model` — jump straight to model selection for that subagent.

## Testing

Focused:

```bash
node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/slash-commands.test.ts
```

Full suite with isolated HOME/USERPROFILE to avoid local user-agent overrides affecting builtin-discovery tests:

```bash
tmp_home=$(mktemp -d) && HOME=$tmp_home USERPROFILE=$tmp_home npm run test:all
```

Result:

```text
# pass 317
# fail 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- Added `/subagents` as a lightweight administration command for inspecting subagent metadata and updating configured models via Pi's available model list. User/project agents are updated in markdown; builtin agents are updated through settings-backed `subagents.agentOverrides`.
+
 ## [0.24.0] - 2026-05-03
 
 ### Changed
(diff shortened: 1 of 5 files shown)


nicobailon:

A replacement admin UI/command is not planned as a high-priority item right now. The old manager UI was intentionally removed, and I’m keeping the active queue focused on runtime issues.

PR #152, opened 2026-05-06
fix: allow explicit nested subagent delegation

Gabrielgvl:
## Summary\n- restore opt-in nested delegation for child agents that explicitly include `subagent` in their tools allowlist\n- keep normal child sessions non-orchestrating by default\n- adjust child prompt boundaries and docs to distinguish safe child mode from explicit nested delegation\n- add unit coverage for nested-delegation env wiring and prompt rewriting\n\n## Why\nThe docs/changelog describe bounded recursion (main → subagent → sub-subagent), but current child-mode registration unconditionally returns when `PI_SUBAGENT_CHILD=1`, so even explicitly configured manager-like agents cannot delegate. This makes `maxSubagentDepth` ineffective for real nested delegation.\n\n## Validation\n- PASS: `node --experimental-strip-types --test test/unit/pi-args.test.ts test/unit/subagent-prompt-runtime.test.ts`\n- PARTIAL: `npm test` currently fails in this local clone on unrelated existing environment/flaky issues:\n  - packaged planner/worker/oracle defaultContext assertion\n  - post-exit stdio guard clean-exit timing\n  - index-child-registration tests cannot resolve peer package `@mariozechner/pi-coding-agent/index.js` in this temp clone\n

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Restore explicit nested delegation for child agents that include `subagent` in their `tools` allowlist while keeping normal child sessions non-orchestrating by default.
+
 ## [0.24.0] - 2026-05-03
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ Use the optional prompt shortcuts below when you want the pattern to be repeatab
 
 Packaged `planner`, `worker`, and `oracle` default to forked context when a launch omits `context`; pass `context: "fresh"` when you intentionally want a fresh child run.
 
-Child-safety boundaries are enforced at runtime. Spawned child sessions do not register the `subagent` tool, do not receive the bundled `pi-subagents` skill, and receive explicit boundary instructions that they are not the parent orchestrator and must not propose or run subagents. Forked child context filtering also removes parent-only subagent artifacts (including old hidden orchestration-instruction messages, slash/status/control messages, and prior parent `subagent` tool-call/tool-result history) while preserving ordinary prose and unrelated tool calls/results.
+Child-safety boundaries are enforced at runtime. Spawned child sessions do not register the `subagent` tool unless the child agent explicitly includes `subagent` in its own `tools` allowlist, do not receive the bundled `pi-subagents` skill, and receive boundary instructions that match their delegation permissions. Forked child context filtering also removes parent-only subagent artifacts (including old hidden orchestration-instruction messages, slash/status/control messages, and prior parent `subagent` tool-call/tool-result history) while preserving ordinary prose and unrelated tool calls/results.
 
 ## Optional shortcuts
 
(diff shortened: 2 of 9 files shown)


albertgwo:
+1, I hit this independently today and arrived at the same root cause from a different environment, so I'll add my reproduction as corroborating evidence for the maintainer.

## Symptom

Running `super-reviewer` from a top-level pi session produced a "single-agent simulation" pass instead of the documented four-persona consensus fanout. The agent's own output included this self-report:

> The parent guard for this child run prohibits launching nested subagents, and `subagent` is not in this child's tool set. The standard four-persona fanout (`ghostmonk-reviewer`, `abchandra-reviewer`, `tnem-reviewer`, `ppoptart12-reviewer`) was therefore not dispatched.

`super-reviewer.md` declares `tools: read, write, bash, subagent` in frontmatter, and the user's `~/.pi/agent/extensions/subagent/config.json` sets `maxSubagentDepth: 3`, so on paper the depth-1 fanout should be allowed.

## Ground-truth check

I added a small `agent_start` hook that dumps `pi.getActiveTools()` and `pi.getAllTools()` to a log file, then dispatched a tiny probe agent declaring `tools: read, write, bash, subagent`.

Before the fix in this PR (pi-subagents 0.24.0 as installed):

```json
{"depth":"1","maxDepth":"3","inheritProjectContext":"0","inheritSkills":"0",
 "active":["read","write","bash","intercom","contact_supervisor"],
 "all":["read","bash","write","contact_supervisor","intercom"]}
```

`subagent` is missing from `all`, not just from `active`. The tool was never registered, which matches the early return at `src/extension/index.ts:223` that this PR addresses.

After applying the change in this PR locally:

```json
{"depth":"1","maxDepth":"3","inheritProjectContext":"0","inheritSkills":"0",
 "active":["read","write","bash","subagent","intercom","contact_supervisor"],
 "all":["read","bash","write","subagent","contact_supervisor","intercom"]}
```

Subsequent dispatches from the probe successfully invoked `subagent({ action: "list" })` and received the executable agent list as a normal `toolResult`. The depth guard in `subagent-executor.ts` is reachable from the child, as the PR description states, so `maxSubagentDepth` is now actually enforcing the documented bound rather than being dead code.

## Why I think the PR's shape is right

The unconditional early-return on `PI_SUBAGENT_CHILD=1` was stricter than the documented contract in two ways. It made `maxSubagentDepth` ineffective for legitimate orchestrator agents (`super-reviewer`, `wave-review-orchestrator`, `review-compiler`), and it forced the model to misattribute its own restraint as "subagent isn't in my tool set," which is what showed up in user-visible output. Gating registration on a new `PI_SUBAGENT_ALLOW_NESTED` env that is only set when the child's own `tools` allowlist contains `subagent` keeps the safe default for leaf workers and scouts while making the orchestrator path actually orchestrate. Updating the soft prompt boundary in lockstep avoids the two layers contradicting each other.

Happy to share the debug hook patch and the probe agent definition if it would help reproduce. CI on this branch shows no checks reported, so if there is anything I can do to help unblock validation, let me know.

nicobailon:

Explicit nested delegation has been reworked and shipped in v0.25.0 with child-safe fanout boundaries and nested status/control. This earlier approach is superseded, so I’m closing it.

PR #160, opened 2026-05-09
Normalize output boolean strings and surface child exit errors

marcomarandiz:
## Summary
- Treat provider-coerced `output: "false"` as disabled output for foreground single, top-level async, and async single runs.
- Treat `output: "true"` like boolean true, using the agent default output path when available.
- Return a clear error for child processes that exit non-zero without stderr, including SIGTERM/SIGKILL hints.

## Why
Some tool schemas/providers can coerce booleans to strings. When `output:false` arrives as `"false"`, subagents currently create a file literally named `false` and inject misleading output instructions. Also, killed children can bubble up as a generic `Failed` instead of the useful exit-code signal.

## Validation
- `npm test`
- `npm run test:integration -- test/integration/parallel-execution.test.ts test/integration/single-execution.test.ts test/integration/async-execution.test.ts`

### DIFF
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -494,7 +494,8 @@ export function executeAsyncSingle(
 		};
 	}
 
-	const outputPath = resolveSingleOutputPath(params.output, ctx.cwd, runnerCwd);
+	const normalizedOutput = params.output === "false" ? false : params.output === "true" ? agentConfig.output : params.output;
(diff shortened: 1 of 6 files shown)


nicobailon:
Landed the relevant output boolean-string normalization on main in c13713e0 (`fix: normalize string false output overrides`), with the final-output cleanup landed separately in 9a831683. Closing this PR as superseded by the mainline commits.

PR #169, opened 2026-05-12
Add option to disable bundled slash commands

davidvornholt:
Hi! 👋

This adds a small `slashCommands: false` config option for folks who want to keep the `subagent` tool available but opt out of the extension's integrated slash commands.

What changed:
- skips registering the bundled slash commands when `slashCommands === false`
- documents the new config option in the README
- adds a unit test to cover the opt-out path

Tested with:
- `npm test -- --runInBand`


nicobailon:

Disabling bundled slash commands is not a high-priority maintenance item right now. I’m closing this to keep the active PR queue focused on runtime correctness issues.

PR #176, opened 2026-05-13
perf: throttle subprocess UI updates to reduce TUI lag during chain execution

anh-chu:
When running subagent chains, the TUI lags badly — spinners barely move and time counters don't update consistently.

## Root cause

The subagent executor in `execution.ts` calls `onUpdate` on every tool start/end and message event from the child process (~24 events per chain step). Each call triggers a full `renderSubagentResult()` rebuild of the TUI component tree, plus a full terminal re-render via `doRender()`. Combined with the 80ms spinner animation interval, the rendering pipeline gets overwhelmed.

### 1. Throttled UI updates (`src/runs/foreground/execution.ts`)
- Added `throttledFireUpdate()` with a 100ms minimum interval between UI updates
- The 4 high-frequency event handlers (`tool_execution_start`, `tool_execution_end`, `message_end`, `tool_result_end`) now use the throttled version
- Final result updates and control events (needs_attention, active_long_running) remain unthrottled via the original `fireUpdate()`

This reduces UI updates from the subprocess by ~3× during active execution.

### 2. Render memoization (`src/tui/render.ts`)
- `renderSubagentResult()` now accepts an optional `context` parameter
- When result details haven't changed (same object reference) and the result is not actively running, the previously built Container component is reused
- Running results still rebuild fully (spinner needs fresh `Date.now()` values), but the throttle limits how often this happens

### 3. Context passthrough (`src/extension/index.ts`)
- `renderResult` passes the render context through to `renderSubagentResult` so memoization state is available via `context.state`

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -467,7 +467,7 @@ DIAGNOSTICS:
 
 		renderResult(result, options, theme, context) {
 			syncResultAnimation(result, context);
-			return renderSubagentResult(result, options, theme);
+			return renderSubagentResult(result, options, theme, context);
 		},
 
 	};
--- a/src/runs/foreground/execution.ts
+++ b/src/runs/foreground/execution.ts
@@ -400,6 +400,16 @@ async function runSingleAttempt(
 			progress.durationMs = Date.now() - startTime;
 			emitUpdateSnapshot(getFinalOutput(result.messages) || "(running...)");
 		};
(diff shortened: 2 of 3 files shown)


nicobailon:

Thanks for investigating TUI update throttling. This is not high or critical priority for the current pass, and the PR is conflicted, so I’m closing it. The broader rendering/flicker problem remains captured in #186.

Issue #147, opened 2026-05-05
Fork-context subagents fail with "thinking blocks must remain as they were" on Anthropic models (esp. after parent session compaction)


tdietert:
## What happened

Spawning a `defaultContext: "fork"` subagent (in this case the builtin `oracle`) from a compacted parent pi session fails immediately with a 400 from the Anthropic API:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.5.content.1: `thinking` or `redacted_thinking` blocks in the latest assistant message cannot be modified. These blocks must remain as they were in the original response."}}
```

The child failed in ~1.4s with zero tool calls — it died on its very first turn while replaying the inherited parent transcript.

Re-running the same task with a **fresh-context** agent (`reviewer`) succeeded immediately and finished normally, which confirms the failure is specific to the fork path, not the agent definition or the prompt.

## Why this is happening (best guess)

When extended thinking is enabled on Anthropic models, the API returns `thinking` / `redacted_thinking` content blocks that are cryptographically signed. The API requires those blocks to be passed back **byte-identical** on subsequent turns of the same conversation — any re-serialization (different JSON key order, whitespace normalization, stripping/adding fields) invalidates the signature and the API rejects the whole request.

`pi-subagents`' fork pipeline appears to either:

1. Re-serialize the parent transcript when handing it to the child, breaking the signature; or
2. Strip/transform thinking blocks in a way that's not API-accepted; or
3. **Most likely in this case:** replay thinking blocks from the parent's *latest* assistant message that no longer match the server-side state, because the parent session was **compacted** — compaction rewrites history but the residual thinking block on the latest assistant turn is no longer the one Anthropic signed.

The parent session in this repro had a compaction marker in its system prompt (`<session_state source="compaction">`).

## Provider / model context

This is **Anthropic-specific**. The signed thinking-block contract is an Anthropic API requirement when extended thinking is enabled — see https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#preserving-thinking-blocks. Other providers (OpenAI, Gemini, etc.) don't impose the same byte-identical replay requirement on reasoning content, so fork against those providers wouldn't surface this bug. Anyone using fork-context subagents on a Claude model with extended thinking enabled in a long-lived (compacted) session is exposed.

## Expected behavior

Fork-context should either:

[...]

PR #177, opened 2026-05-13
fix: sanitize forked session tool IDs and thinking signatures for Anthropic compatibility

anh-chu:
When a subagent worker runs in forked context with an Anthropic model, the inherited session file contains tool call IDs with colons (e.g. `read:0`, `todo_md:7`) and invalid thinking signatures (e.g. `"reasoning"`). Anthropic's API rejects these with:

- `400: String should match pattern '^[a-zA-Z0-9_-]+$'` (tool_use.id)
- `400: Invalid signature in thinking block` (thinking signature)

## Changes

- Add `sanitizeForkedSession()` that scans forked session JSONL entries
- Replace characters outside `[a-zA-Z0-9_-]` with `_` in tool call IDs
- Strip `thinkingSignature` fields from assistant messages (the value is not a valid cryptographic signature)
- Preserve ID consistency between `toolCall.id` and `toolCallId` references
- Only affects forked subagent sessions; the orchestrator session is unchanged

### DIFF
--- a/src/shared/fork-context.ts
+++ b/src/shared/fork-context.ts
@@ -1,6 +1,75 @@
 import * as fs from "node:fs";
 import { SessionManager } from "@earendil-works/pi-coding-agent";
 
+const VALID_TOOL_ID = /^[a-zA-Z0-9_-]+$/;
+
+function sanitizeForkedSession(sessionFile: string): void {
+	const lines = fs.readFileSync(sessionFile, "utf-8").split("\n");
+	const mapping = new Map<string, string>();
+	const seen = new Set<string>();
+
+	function getSanitized(id: string): string {
+		if (VALID_TOOL_ID.test(id)) return id;
+		if (mapping.has(id)) return mapping.get(id)!;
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks for digging into this. The Anthropic fork-context bug in #147 is real, but I would not merge this draft as-is.

The sanitizer currently walks the whole message payload and rewrites any nested id field, so a tool argument like id: issue:147 could be changed even though it is not a tool-call id. It also can collapse read:0 into an existing valid read_0, and stripping thinkingSignature from redacted thinking leaves the Anthropic replay path with a redacted block that no longer has data.

The useful next version would be the smallest fork-context fix that only remaps actual tool-call ids and matching tool results, handles normal and redacted thinking deliberately, and adds a focused test for the forked session JSONL shape.

nicobailon:
I'm closing this for now while we clean up the PR queue. The Anthropic fork-session issue may still be worth revisiting, but this draft needs a narrower current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #187, opened 2026-05-15
feat: expose thinking level parameter to /run, /chain, and /parallel

rubdos:
Add support for overriding agent thinking level at invocation time via inline configuration. Users can now specify thinking level directly in slash commands:

```
  /run researcher[thinking=extended] Analyze this complex problem
  /chain scout[thinking=on] -> planner[thinking=extended] Plan and execute
  /parallel agent1[thinking=false] + agent2[thinking=high] Parallel tasks
```

and agents can spawn subagents with a specified thinking level.

This allows control over reasoning depth per-invocation without modifying agent configurations. Specifically, it allows using models for subagents that support a disjunct set of thinking levels.

---

My use case: I have my worker agent default to `mistral-medium-3.5:high`, and I'd like to be able to switch around to no thinking, or to another model on xhigh or no thinking level.

This is mostly vibe-coded (and vibe-tested by running it and letting agent test it out). I'll be happy to fix up this PR until mergeable.

### DIFF
--- a/src/agents/agent-management.ts
+++ b/src/agents/agent-management.ts
@@ -196,6 +196,11 @@ function parseStepList(raw: unknown): { steps?: ChainStepConfig[]; error?: strin
 			if (typeof s.progress === "boolean") step.progress = s.progress;
 			else return { error: `config.steps[${i}].progress must be a boolean.` };
 		}
+		if (hasKey(s, "thinking")) {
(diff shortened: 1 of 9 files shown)


nicobailon:

Thanks for the thinking-level work. This PR does not directly solve the current high-priority provider-compatibility problem in #212, and it is not high or critical priority on its own, so I’m closing it during triage.

Issue #182, opened 2026-05-14
@earendil-works/pi-tui missing from dependencies — causes Cannot find module on pi upgrade


irizzant:
## Problem
After running `pi upgrade`, `pi` crashes on startup with:

```
Error: Failed to load extension "/.../pi-subagents/src/extension/index.ts":
Failed to load extension: Cannot find module '@earendil-works/pi-tui'
```

## Root cause
`@earendil-works/pi-tui` is listed in `devDependencies` and in `peerDependencies` (as `optional: true`), but it is **not** in `dependencies`.

Because it is an *optional* peer dependency, `npm` does not install it automatically. However, `src/extension/index.ts` imports it unconditionally at runtime, so it is effectively required.

## Environment
- `pi-subagents` version: `0.24.2`
- `pi` version: `0.73.1`
- Node version: `v24.6.0`

## Workaround
Manually install the missing package globally:
```bash
npm install -g @earendil-works/pi-tui
```

## Suggested fix
Move `@earendil-works/pi-tui` from `devDependencies` to `dependencies` (or make the peer dependency non-optional) so that `npm install -g pi-subagents` brings it in automatically.

XXXBusy:
I can take this as a small fixed-scope npm packaging fix if paid help is useful.

Scope:
- reproduce the missing `@earendil-works/pi-tui` runtime dependency from a clean install/upgrade path
- move or guard the dependency so the extension can load reliably after `pi upgrade`
- add a smoke check around extension loading or document the exact manual verification if the repo does not have a test harness for this path

Fixed price: **$50** after the PR is accepted/merged. Payment can be GitHub Sponsors, PayPal, crypto, or invoice.

I would keep it narrowly focused on the runtime dependency/load failure, not broader extension changes.

nicobailon:
Fixed on main in c16ab0ea (`fix: ship pi tui as runtime dependency`). `@earendil-works/pi-tui` is now a runtime dependency instead of an optional peer/dev-only dependency. Validated with `npm run test:all` and `git diff --check` before pushing fd8e225e.

PR #188, opened 2026-05-15
fix: move @earendil-works/pi-tui from optional peer dep to direct dependency

aprasadh:
## Fix for #182

`@earendil-works/pi-tui` was listed as an optional peer dependency, so `npm install -g pi-subagents` skipped it. Since `src/extension/index.ts` imports it unconditionally at runtime, this caused a `Cannot find module '@earendil-works/pi-tui'` crash at startup.

### Changes
- **Moved `@earendil-works/pi-tui`** from `devDependencies` → `dependencies` (ensures it's installed as a required runtime dependency with `npm install -g`)
- **Removed `@earendil-works/pi-tui`** from `peerDependencies` and `peerDependenciesMeta.optional` (removing the contradictory optional peer dep entry)

### Verification
- All 378 unit tests pass with this change
- No code changes needed — the runtime import was always unconditional

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,8 +54,7 @@
   "peerDependencies": {
     "@earendil-works/pi-agent-core": "*",
     "@earendil-works/pi-ai": "*",
-    "@earendil-works/pi-coding-agent": "*",
-    "@earendil-works/pi-tui": "*"
+    "@earendil-works/pi-coding-agent": "*"
   },
   "peerDependenciesMeta": {
     "@earendil-works/pi-agent-core": {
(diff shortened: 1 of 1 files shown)


ian (approved):

shaharmor:
This should be merged asap. It's blocking usage of latest Pi versions with subagents

nicobailon:
Landed the equivalent fix on main in c16ab0ea (`fix: ship pi tui as runtime dependency`). Closing this PR as superseded by the mainline commit.

PR #189, opened 2026-05-17
The tiniest PR

tmustier:
Hey Nico — tiny thing I noticed in Ghostty/cmux with pi-subagents: the turn indicator shows up smushed (see screenshot). 
image

But fear not, for I have a brilliant fix should you choose to accept it!

We can add:
```
 
```

(a space 🫶 )

image

### DIFF
--- a/src/tui/render.ts
+++ b/src/tui/render.ts
@@ -840,7 +840,7 @@ function renderSingleCompact(d: Details, r: Details["results"][number], theme: T
 	const isRunning = r.progress?.status === "running";
 	const contextBadge = d.context === "fork" ? theme.fg("warning", " [fork]") : "";
 	const stats = statJoin(theme, [
-		r.usage?.turns ? `⟳${r.usage.turns}` : "",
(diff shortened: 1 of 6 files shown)


tmustier:
LOL my +1/-1 space caused tests to fail because of course... 
Now we're fixed and green

tmustier:
Also fixed CI so I don't leave it red - optional second commit.

So there are 2 commits:
  1. `Add space after turn indicator`, which is adding the space and updating the render-fork-badge test to match
2. **Fix unrelated Windows test breakage** which fixes 4 unrelated tests that aren't related but were failing

The second commit, optionally, fixes some Windows-only test failures that are (i promise!) unrelated:
- async-execution.test.ts: fixed a race between the control event being written and the test reading status.json before the run cleared its live state.
- fork-context-execution.test.ts: avoid reading mock pi calls from other tests.
- render-widget.test.ts: allow Windows or Unix path separators.
- single-execution.test.ts: normalize Windows backslashes before checking the extension path.

nicobailon:
Sorry for the delay! This has been addressed, thanks! Will release shortly.

PR #192, opened 2026-05-18
Allow child subagents to dispatch when their tools declare subagent

albertgwo:
## Purpose

Allow child subagents to dispatch their own subagents when the child agent definition explicitly declares the `subagent` tool. This re-enables the canonical use of agents like `wave-review-orchestrator` whose role is to run a parallel reviewer fanout, without weakening the strict child boundary for every other child.

## Summary of changes

- **Env-var gate** (`src/runs/shared/pi-args.ts`): adds `PI_SUBAGENT_FANOUT_CHILD`, set on the spawned child whenever `input.tools?.includes("subagent")` is true. The agent's own tools declaration is the authority; no new opt-in flag.
- **Tool registration** (`src/extension/index.ts`): the early return that blocks `registerTool` in every child now skips only when `PI_SUBAGENT_FANOUT_CHILD` is not set, so fanout children get the `subagent` tool registered.
- **Boundary text** (`src/runs/shared/subagent-prompt-runtime.ts`): adds `CHILD_FANOUT_BOUNDARY_INSTRUCTIONS`, a fanout-aware boundary used in place of the strict text when the gate is set. The softer text keeps "you are a child, parent owns final orchestration, do not propose dispatching agents not in your tools list", and drops only the unconditional "do not propose or run subagents" clause that contradicts the agent's own tools declaration.
- **History stripping** (`stripParentOnlySubagentMessages`): inherited `subagent` tool-call and tool-result messages are now preserved for fanout children. Parent-only custom message types (`subagent-notify`, control notices, slash result) are still stripped for both fanout and non-fanout children.
- **Tests**: 8 new cases across `test/unit/index-child-registration.test.ts`, `test/unit/pi-args.test.ts`, and `test/unit/subagent-prompt-runtime.test.ts`. Full suite passes: 387 tests, 0 failing.
- **Backward compatibility**: the strict `CHILD_SUBAGENT_BOUNDARY_INSTRUCTIONS` remains the default for any child whose tools list does not declare `subagent`, and `maxSubagentDepth` continues to bound global recursion in both modes.

## Changes to review

- **Gate authority**: the env var is set from `input.tools?.includes("subagent")` rather than a new frontmatter flag. The agent's tools list already exists as the load-bearing declaration; coupling fanout permission to it avoids API surface bloat. If you would prefer an explicit `allowSubagentFanout: true` frontmatter flag, that is a one-line swap on the gate site in `pi-args.ts`.
- **Softer boundary wording** in `CHILD_FANOUT_BOUNDARY_INSTRUCTIONS`. Specifically the line "You may dispatch the subagents listed in your `tools` for the duration of this task. Do not propose dispatching agents not in that list." Confirm this matches the contract you want the child to follow.
- **History stripping change** in `stripParentOnlySubagentMessages`: for fanout children, inherited subagent tool-call and tool-result blocks pass through. This is the right default when the child is meant to continue an in-flight fanout, but it does mean a fanout child sees more of the parent's earlier orchestration than a strict child would. Confirm this is acceptable.
- **No depth-cap interaction**: the patches do not touch `maxSubagentDepth` or `checkSubagentDepth`. A fanout child at the configured cap still gets blocked by the existing depth check when it attempts to dispatch.

### DIFF
--- a/UPSTREAM.md
+++ b/UPSTREAM.md
@@ -1,0 +1,63 @@
+# Upstream fork notes
+
+This fork of [`nicobailon/pi-subagents`](https://github.com/nicobailon/pi-subagents) carries one patch series, `fanout-children`, that re-enables subagent dispatch for child agents whose own `tools` declaration includes `subagent`. The strict child boundary remains the default for every other child.
+
+## Why the fork exists
+
+pi-subagents 0.24.3 unconditionally:
+
+1. Skips registering the `subagent` tool in any child process (`src/extension/index.ts` early return).
+2. Prepends `CHILD_SUBAGENT_BOUNDARY_INSTRUCTIONS` to every child's system prompt, including the clause "Do not propose or run subagents."
+3. Strips all `subagent` tool-call blocks and tool-result messages from inherited history (`stripParentOnlySubagentMessages`).
+
+Together these three sites make it impossible for any child subagent to run its own fanout, even if the agent definition explicitly declares the `subagent` tool. This blocks the canonical use of agents like `wave-review-orchestrator`, whose entire role is to dispatch persona and toolkit reviewers in parallel.
+
+Upstream maintenance has paused, so the fix lives here instead.
+
+## What the `fanout-children` branch changes
+
+The branch adds a new env var, `PI_SUBAGENT_FANOUT_CHILD`, set by the parent in `buildPiArgs` whenever the child agent's `tools` list includes `subagent`. Three sites consult that env:
+
+| Site | Old behavior | New behavior |
+| --- | --- | --- |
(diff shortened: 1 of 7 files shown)


nicobailon:
Thanks for the PR. I reworked this on main rather than merging the branch as-is, and the nested fanout support has now shipped in v0.25.0. The released version keeps fanout gated to agents whose resolved builtin tools explicitly include `subagent`, uses child-safe registration, adds parent-visible nested status/control by id, carries nested result summaries, and includes the follow-up hardening around retryable result enrichment, scoped child-safe status, and nested route/path sanitization. Closing this since the feature is now covered by the release.

PR #197, opened 2026-05-20
completion-guard: skip mutation expectation when agent declares only read-only tools

JayceFreeman:
Three failure shapes share one root cause: the guard reads task prose to
infer 'is this an implementation task?' when the agent's frontmatter
`tools:` already declares whether it CAN implement. Architect (16 runs,
$21.66) and reviewer-codex-xhigh (3 runs, $14.48) both declare
`tools: read, grep, find, ls` — yet the prose-regex chain fires anyway
because their tasks contain implementation verbs.

Floor: if the agent declares only read-only tools and no MCP tools,
mutation is impossible regardless of prose. Skip the regex chain.

Concrete changes:

- long-running-guard.ts: export READ_ONLY_TOOL_NAMES (read, grep, find,
  ls, web_search, fetch_content, get_search_content, intercom,
  contact_supervisor) and isReadOnlyToolName(). Unknown tool names are
  conservatively NOT in the set — the safe direction.
- long-running-guard.ts: extend MUTATING_BASH_PATTERNS with
  `/\bf\.write\s*\(/` to catch the python heredoc shape
  `with open(p) as f: f.write(...)` (worker false-positive case from
  the handoff §8(c)).
- completion-guard.ts: add `tools?` and `mcpDirectTools?` to
  CompletionMutationGuardInput. Add declaresOnlyReadOnlyTools() helper
  and an early return at the top of expectsImplementationMutation().
- foreground/execution.ts: pass `agent.tools` and
  `agent.mcpDirectTools` into the guard call.
- background/subagent-runner.ts: same plumbing from `step.tools` /
  `step.mcpDirectTools`.

Tests (test/unit/completion-guard.test.ts):
- Architect with read-only tools never expects mutation, even for
  implementation-y task text (§8(a)).
- Reviewer with read-only tools is not flagged by 'must fix' prose
  (§8(b)).
- Worker with the new f.write pattern + 'Changed files:' output does
  not trigger (§8(c)).
- Worker with mutation tools but no edits and no f.write still
  triggers (§8(d)).
- Backward compat: omitting tools falls through to the existing prose
  path; unknown tool names are treated as possibly mutating; MCP tools
  disqualify the short-circuit; negative case for \bf\.write\( does
  not match `selfwrite(`.

Deferred (named, separate slices):

- The 'soft failure preserves artefact' corollary (single-output.ts,
  subagent-executor.ts:1946-1950 gate on exitCode === 0). Only matters
  for TRUE guard trips; this change removes the false-positive class
  first.
- Restructure expectsImplementationMutation as
  `canMutate(tools) && askedToMutate(agent, task)` — the revelation
  refactor. Code comment marks the spot.
- tools: false override collapses to undefined upstream in agents.ts.
  Edge case; documented in the new field doc comment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@ node_modules/
 .fallow/
 
 package-lock.json
+
+# pi-lens local cache (committed by mistake earlier; ignored going forward)
+.pi-lens/
--- a/src/runs/background/subagent-runner.ts
+++ b/src/runs/background/subagent-runner.ts
@@ -641,6 +641,8 @@ async function runSingleStep(
 				agent: step.agent,
 				task,
 				messages: run.messages,
+				tools: step.tools,
+				mcpDirectTools: step.mcpDirectTools,
 			})
 			: undefined;
 		const completionGuardTriggered = completionGuard?.triggered === true && !run.observedMutationAttempt;
--- a/src/runs/foreground/execution.ts
+++ b/src/runs/foreground/execution.ts
@@ -663,7 +663,13 @@ async function runSingleAttempt(
 
 	let fullOutput = getFinalOutput(result.messages);
 	const completionGuard = result.exitCode === 0 && !result.error
-		? evaluateCompletionMutationGuard({ agent: agent.name, task, messages: result.messages })
+		? evaluateCompletionMutationGuard({
+			agent: agent.name,
+			task,
+			messages: result.messages,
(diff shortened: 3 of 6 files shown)


nicobailon:
Landed the completion-guard fix on main in 42ae7f1b (`fix: respect read-only completion guards`). Closing this PR as superseded by the mainline commit.

PR #201, opened 2026-05-20
Raise default parallel concurrency

andreasbauer80:
## Summary
- raise default top-level and chain parallel concurrency from 4 to 5
- raise default top-level parallel max tasks from 8 to 10
- update schema descriptions, README, changelog, and unit expectations

## Tests
- `node --experimental-strip-types --test test/unit/parallel-utils.test.ts test/unit/recursion-guard.test.ts test/unit/schemas.test.ts`
- `npm run test:unit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Added
 
+### Changed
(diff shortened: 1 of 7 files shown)


nicobailon:

Raising default concurrency/max tasks is not planned. The safer default is to keep fanout conservative and let users opt into larger runs via configuration.

PR #202, opened 2026-05-21
feat: add staggerMs + staggerJitter config for parallel subagents

kierr:
A burst of parallel subagents hammers providers and triggers velocity rate limits, so I added these to spread things out:

- `staggerMs` — delay between each worker start (default 0, no delay)
- `staggerJitter` — randomness fraction (default 0.2 = ±20%)

Worker N sleeps for N × staggerMs ± staggerJitter × staggerMs before its first task. `mapConcurrent` now takes an options bag for stagger config.

I've set defaults to preserve existing behavior (i.e. no stagger) but I recommend sane defaults to prevent thundering herd - particularly when the parallel subagent limit is set high (5 with no stagger is OK, but 50... 429!)

A small note: #43 added something similar to resolve a different issue (file lock contention) but it was later dropped to rely on pi's lock retry behavior instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -807,12 +807,16 @@ Forces depth-0 single, parallel, and chain runs into background mode and bypasse
 {
   "parallel": {
     "maxTasks": 12,
-    "concurrency": 6
+    "concurrency": 6,
+    "staggerMs": 100,
+    "staggerJitter": 0.2
   }
 }
 ```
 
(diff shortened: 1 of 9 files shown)


nicobailon:

Thanks for the rate-limit mitigation idea. Staggered fanout is useful in some setups, but it is not high or critical priority for this pass. Current callers can control blast radius with `concurrency`, so I’m closing this for now.

PR #204, opened 2026-05-21
fix: ignore legacy skill dirs in agent discovery

affsantos:
## Summary

Legacy `.agents/**/*.md` discovery currently treats Pi skill files under `.agents/skills/**` as executable subagents when they contain `name` and `description` frontmatter. That makes valid skill packages like `.agents/skills/foo/SKILL.md` show up in `subagent list` as runnable agents.

This keeps the legacy `.agents` compatibility behavior, but excludes `.agents/skills/**` from agent discovery. The intent is to preserve custom legacy agents under `.agents/` while respecting Pi's documented skill location under `.agents/skills/`.

## Why

Pi documents both canonical and legacy skill locations, including:

- `~/.pi/agent/skills/`
- `~/.agents/skills/`
- `.pi/skills/`
- `.agents/skills/`

Since skills also use markdown/frontmatter, broad `.agents/**/*.md` agent discovery can accidentally classify skill docs as executable agents. This is confusing in `subagent list`, where skill instruction packages can appear to be runnable subagents.

## Changes

- Add an optional directory-skip predicate to recursive markdown discovery.
- Skip `skills` subtrees only when scanning a legacy `.agents` root for agents.
- Add coverage that `.agents/skills/*.md` and `.agents/skills/*/SKILL.md` are ignored while ordinary legacy `.agents/*.md` agents still load.

## Validation

- `node --experimental-strip-types --test test/unit/agent-frontmatter.test.ts`

Note: `npm test -- test/unit/agent-frontmatter.test.ts` starts the full unit suite in this checkout and hits unrelated missing peer dependencies for TUI-related tests (`@earendil-works/pi-tui`, `@earendil-works/pi-coding-agent`). The targeted test file passes with Node's strip-types runner.

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -535,7 +535,11 @@ export function removeBuiltinAgentOverride(cwd: string, name: string, scope: "us
 	return filePath;
 }
 
-function listMarkdownFilesRecursive(dir: string, predicate: (fileName: string) => boolean): string[] {
+function listMarkdownFilesRecursive(
+	dir: string,
+	predicate: (fileName: string) => boolean,
+	shouldSkipDir: (dirPath: string) => boolean = () => false,
+): string[] {
 	const files: string[] = [];
 	if (!fs.existsSync(dir)) return files;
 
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for putting this together. I'm closing this because the legacy skill-discovery bug is now fixed on main by #303.

That fix covers the same .agents/skills case from this PR and also covers user legacy skills and package roots that expose broad agent directories, while keeping real SKILL.md agent files loadable. This branch is now conflicting with main, so I do not want to ask for a rework just to land overlapping behavior.

PR #206, opened 2026-05-22
fix: handle deferred session file writes from createBranchedSession

mevatron:
SessionManager.createBranchedSession() now defers writing the branched session file when the path contains no assistant messages (matching the newSession() contract). The file is only written when _persist() is called with the first assistant response.

This caused forked subagent sessions to fail with:
  'Session manager returned a forked session file that does not exist'

Fix: When createBranchedSession() returns a path but the file doesn't exist, write it ourselves using the source manager's getBranch() method to extract the correct branch entries. After createBranchedSession(), the source manager's internal state has been switched to the branched session, so getBranch() returns the correct path.

- Add writeBranchedSessionFile() fallback that writes the session file with a proper header and branch entries
- Import SessionEntry type from pi-coding-agent SDK
- Update tests to cover the deferred-write scenario with both mock and real SessionManager instances

### DIFF
--- a/src/shared/fork-context.ts
+++ b/src/shared/fork-context.ts
@@ -1,5 +1,5 @@
 import * as fs from "node:fs";
-import { SessionManager } from "@earendil-works/pi-coding-agent";
+import { CURRENT_SESSION_VERSION, SessionManager, type SessionEntry } from "@earendil-works/pi-coding-agent";
 
 type SubagentExecutionContext = "fresh" | "fork";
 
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for digging into the right failure. I would not merge this as-is yet.

The fallback makes the missing child session file exist, but it writes a branch that has no assistant message. When the child opens that file and later writes its first response, Pi treats the existing file as already written and then appends the whole in-memory session again. That leaves duplicate session headers and duplicate parent history in the child session file.

The smallest useful revision would keep forked startup working without writing a no-assistant file that Pi later re-appends, and add a regression that opens the fallback child file, appends the child user turn and first assistant response, and confirms the history is written once.

nicobailon:
I'm closing this for now while we clean up the PR queue. The deferred-session bug is real, but this branch needs a revised fix before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #219, opened 2026-05-27
Apply agentOverrides to user/project custom agents (frontmatter wins per field)

### LINKED ISSUE #218: agentOverrides silently ignored for custom user/project agents
# `agentOverrides` silently ignored for custom user/project agents

## Summary

`subagents.agentOverrides.<name>` in `settings.json` applies only to builtin agents. When a custom user-scope (`~/.pi/agent.<flavor>/agents/<name>.md`, `~/.agents/<name>.md`) or project-scope (`<repo>/.pi/agents/<name>.md`) agent has the same name as an entry in `agentOverrides`, the override is silently dropped and the agent falls through to `defaultModel` / its frontmatter / etc.

The setting key (`agentOverrides`, not `builtinAgentOverrides`) and the surrounding prose in the README don't telegraph the builtins-only restriction, and the loader emits no warning when an override matches no builtin — so the override looks correct, the agent runs on the wrong model, and the only way to notice is the model identifier in the TUI run line.

## Repro

```jsonc
// ~/.pi/agent/settings.json
{
  "defaultModel": "anthropic/claude-opus-4-7",
  "subagents": {
    "agentOverrides": {
      "reviewer":    { "model": "anthropic/claude-sonnet-4-6" },  // builtin → applied
      "implementer": { "model": "anthropic/claude-sonnet-4-6" }   // custom  → silently dropped
    }
  }
}
```

```yaml
# <repo>/.pi/agents/implementer.md
---
name: implementer
description: TDD-disciplined implementer
tools: bash, read, edit, write
# no `model:` field — relying on agentOverrides
---
<system prompt...>
```

Dispatch `subagent({ agent: "implementer", task: "…" })`. Subagent runs on `claude-opus-4-7` (the default), not `claude-sonnet-4-6`. The TUI run line is the smoking gun:

```
✓ implementer (claude-opus-4-7) · 12 tool uses · 1m21s
```

## Root cause

In `src/agents/agents.ts`, `applyBuiltinOverrides(...)` is invoked exclusively against `loadAgentsFromDir(BUILTIN_AGENTS_DIR, "builtin")`. User and project agent arrays pass through untouched:

```ts
// discoverAgents
const builtinAgents = applyBuiltinOverrides(loadAgentsFromDir(BUILTIN_AGENTS_DIR, "builtin"), ...);

const userAgentsOld = scope === "project" ? [] : loadAgentsFromDir(userDirOld, "user");
const userAgentsNew = scope === "project" ? [] : loadAgentsFromDir(userDirNew, "user");
const userAgents = [...userAgentsOld, ...userAgentsNew];

const projectAgents = scope === "user" ? []
    : projectAgentDirs.flatMap((dir) => loadAgentsFromDir(dir, "project"));
```

`discoverAgentsAll(...)` mirrors the same pattern.

The override helper itself is generic — it merges purely by agent name. Only the calling convention restricts it to builtins.

## Why this matters

The case that surfaced this: **one repo, multiple Pi harnesses**. A team checks `implementer` / `code-reviewer` / `spec-reviewer` personas into `.pi/agents/` so the prompt is version-controlled and identical across the team. Individual developers run different Pi distributions — some via direct Anthropic, some via Bedrock — and the model identifiers differ between harnesses (`anthropic/claude-sonnet-4-6` vs a bedrock-prefixed variant).

Without `agentOverrides` reaching custom agents, today's options are all bad:

- **Pin a model in the agent's frontmatter.** Commits a provider choice into the shared repo. Breaks the other harness.
- **Duplicate the agent into per-harness user-scope directories** (`~/.pi/agent.<flavor>/agents/<name>.md`). Defeats the point of having a single shared persona.
- **Replace the persona by overriding a similarly-named builtin and stuff the whole system prompt into `agentOverrides.<builtin>.systemPrompt`** as an escaped JSON string. Ugly, hard to review, loses semantic agent naming.

Letting `agentOverrides` reach custom agents — with frontmatter as the per-agent floor — removes the whole category of friction. The persona body lives in `.pi/agents/<name>.md` (shared, version-controlled), the model lives in per-harness `settings.json` (local, per-developer).

Same change also surfaces a quieter issue: today, an `agentOverrides` entry whose name doesn't match any builtin loads with zero feedback. After this fix, "no match anywhere" is still silent, but the override at least reaches the agent the user named — which is presumably what they meant.

## Proposed behavior

Apply `agentOverrides[name]` to any agent with a matching name — user, project, or builtin — with one safety rule for custom agents: **frontmatter wins per-field**. The override fills only fields the agent's frontmatter left unset.

- Builtin agents: unchanged. Builtins ship with no frontmatter values, so per-field fill-in collapses to the existing overwrite behavior. Existing tests pass without modification.
- Custom agent that pins `model:` (or any other field) in frontmatter: unchanged. Override silently doesn't touch the pinned field. Already-covered by the existing "frontmatter wins on shadowing project agent" test (renamed to make the rule explicit).
- Custom agent that leaves a field unset, paired with a matching `agentOverrides` entry: now inherits the override that previously sat idle. This is the new behavior.
- `disableBuiltins`: still touches only builtins. Custom agents are never bulk-disabled.

## Breaking-change surface

Effectively zero for the documented contract. The only behavior change a user could observe:

- Custom agent named `<X>`, with no `<field>:` in frontmatter, plus `agentOverrides.<X>.<field>` in `settings.json` → that field is now populated from the override instead of staying unset / falling through to `defaultModel`.

In every plausible scenario this is what the user already wanted (otherwise why is the override entry there?), so I'd rate this as a documentation-clarification-grade change rather than a semver-minor one. Happy to call it out in the CHANGELOG either way.

## Implementation

PR linked separately. ~85 LOC: one new helper (`applyCustomAgentOverride`) implementing per-field fill-in, one thin wrapper (`applyCustomAgentOverrides`) handling the project-over-user precedence, both wired into `discoverAgents` and `discoverAgentsAll`. 6 new unit tests; existing 13 tests in `agent-overrides.test.ts` unchanged in behavior (one renamed for clarity). No other files touched in `src/`.

Side note picked up while testing: `test/unit/agent-overrides.test.ts` overrides `HOME` per-test but doesn't unset `PI_CODING_AGENT_DIR`, so the suite reads the developer's real config when that env var is set. Workaround is `env -u PI_CODING_AGENT_DIR node --test ...`. Worth fixing separately — flagging here so the PR's "tests pass" claim is reproducible.

Happy to send the PR — wanted to file the issue first so the naming question (`agentOverrides` vs renaming the helper) and the changelog-classification can be settled before code lands.

jjuraszek:
# Apply `agentOverrides` to user/project custom agents (frontmatter wins per field)

Fixes nicobailon/pi-subagents#218.

## What this changes

`subagents.agentOverrides.<name>` now resolves against any agent with a matching name — user-scope, project-scope, or builtin. Previously it resolved only against builtins; matches against custom agents were silently dropped.

One safety rule for custom agents: **frontmatter wins per-field**. The override fills only fields the agent's frontmatter left unset. Builtins continue to behave exactly as before because they ship with no frontmatter values, so per-field fill-in collapses to the existing overwrite behavior.

## Why

Detailed motivation in the linked issue. The short version: teams ship a custom persona in `.pi/agents/<name>.md` so the prompt is shared and version-controlled, while individual developers run different Pi distributions (direct Anthropic, Bedrock, ...) whose model identifiers differ. There's no clean way today to keep the persona body shared and let each harness pick its own model — `agentOverrides` looks like it should do exactly that, but doesn't reach custom agents.

## Implementation

Single source file: `src/agents/agents.ts`. New helpers, sibling to the existing `applyBuiltinOverride` / `applyBuiltinOverrides`:

```ts
function applyCustomAgentOverride(agent, override, meta): AgentConfig {
    // For each field: if override sets it AND agent's frontmatter didn't, fill it in.
    // If frontmatter set it, keep frontmatter.
    // If override matched but didn't fill anything, return the original agent unchanged
    // (so the `override` annotation stays absent, matching existing test semantics).
}

function applyCustomAgentOverrides(agents, userSettings, projectSettings, ...): AgentConfig[] {
    // Per-agent: project override (if project settings.json exists) wins over user override.
    // Same precedence as builtin path.
}
```

Wired into both discovery entry points:

```diff
-const userAgents = [...userAgentsOld, ...userAgentsNew];
-const projectAgents = scope === "user" ? [] : projectAgentDirs.flatMap(...);
+const userAgents = applyCustomAgentOverrides([...userAgentsOld, ...userAgentsNew], ...);
+const projectAgents = applyCustomAgentOverrides(
+    scope === "user" ? [] : projectAgentDirs.flatMap(...),
+    ...,
+);
```

`disableBuiltins` deliberately stays out of the custom path — bulk-disabling builtins shouldn't bulk-disable a user's hand-written agents.

The helper name `applyBuiltinOverrides` is now slightly misleading since a sibling function does the same for non-builtins. A rename (`applyAgentOverridesBuiltin` / `applyAgentOverridesCustom`, or merging the two) is a clean follow-up; kept out of this PR to keep the diff focused on behavior.

## Behavior summary

| Agent shape | `agentOverrides[name]` entry | Before | After |
|---|---|---|---|
| Builtin, no override | — | bundled defaults | unchanged |
| Builtin, override matches | sets `model` | override applied | unchanged |
| Custom agent, no override entry | — | as-loaded | unchanged |
| Custom agent, override matches, frontmatter has the field | sets `model` | override silently dropped | unchanged — frontmatter still wins |
| Custom agent, override matches, frontmatter left field unset | sets `model` | override silently dropped, falls through to `defaultModel` | **override now applied** |
| Custom agent + `disableBuiltins: true` | — | custom agent unaffected | unchanged |

Only the bold row is new behavior.

## Tests

6 new unit tests in `test/unit/agent-overrides.test.ts`:

- `fills in unset fields on a custom project agent from project agentOverrides`
- `fills in unset fields on a custom user agent from user agentOverrides`
- `applies user agentOverrides to a custom project agent when project settings have no entry`
- `prefers project agentOverrides over user agentOverrides on a custom project agent`
- `leaves a custom agent untouched when no agentOverrides entry matches its name`
- `disableBuiltins does not disable custom agents`

The pre-existing test asserting "frontmatter wins when a project agent shadows a builtin" stays green; renamed to `frontmatter wins per-field over agentOverrides for a shadowing project agent` to spell out the rule that's now generalized.

All 19 tests in `agent-overrides.test.ts` pass. Pre-existing failures in unrelated test files (`fork-context`, `nested-control`, `render-helpers`, `widget-nested-render`) are identical before and after this PR — verified with `git stash; git checkout main; node --test ...` and back.

## Test isolation caveat

`agent-overrides.test.ts` resets `HOME` per-test but doesn't unset `PI_CODING_AGENT_DIR`. If a developer has that env var pointing to a real `~/.pi/agent.*/settings.json` with `agentOverrides` entries (typical when running tests from inside a Pi session), several existing tests fail because the suite reads real config instead of the temp fixture. Workaround:

```bash
env -u PI_CODING_AGENT_DIR node --test test/unit/agent-overrides.test.ts
```

All 19 pass under that invocation. Worth tightening `beforeEach` to also clear `PI_CODING_AGENT_DIR`, but I left it out of this PR — it's an orthogonal cleanup and would muddy the diff.

## CHANGELOG

Entry added under `[Unreleased] / Added`:

> Apply `subagents.agentOverrides[name]` to user-scope and project-scope custom agents in addition to builtins. Frontmatter wins per-field — overrides only fill fields the agent's frontmatter left unset, so existing custom agents that pin their own model/thinking/etc. are unaffected. Lets shared persona files (`.pi/agents/<name>.md`) stay version-controlled while per-harness `settings.json` supplies the local model. `disableBuiltins` continues to apply only to builtins.

Happy to move it under `Changed` or `Fixed` if you prefer — the behavior change is small enough that all three are defensible.

## Out of scope

- Renaming `applyBuiltinOverrides` → `applyAgentOverridesBuiltin` (mechanical follow-up).
- Warning the user when an `agentOverrides[name]` entry matches zero agents anywhere.
- Tightening `beforeEach` in `agent-overrides.test.ts` to clear `PI_CODING_AGENT_DIR`.

Each is small and unrelated; happy to send separate PRs.

## Verification

Reproduced locally against a real project-scope custom agent (`gridstrong/.pi/agents/implementer.md`, no `model:` in frontmatter) with a user-scope `agentOverrides.implementer.model = "anthropic/claude-sonnet-4-6"`.

**Before:** `discoverAgents(...)` returns `{ source: "project", model: undefined, override: undefined }`, subagent dispatch runs on `claude-opus-4-7` (the default).

**After:** `discoverAgents(...)` returns `{ source: "project", model: "anthropic/claude-sonnet-4-6", override: { scope: "user", ... } }`, subagent dispatch runs on `claude-sonnet-4-6`. Builtin overrides for `worker`/`reviewer` continue to apply identically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- Apply `subagents.agentOverrides[name]` to user-scope and project-scope custom agents in addition to builtins. Frontmatter wins per-field — overrides only fill fields the agent's frontmatter left unset, so existing custom agents that pin their own model/thinking/etc. are unaffected. Lets shared persona files (`.pi/agents/<name>.md`) stay version-controlled while per-harness `settings.json` supplies the local model. `disableBuiltins` continues to apply only to builtins.
+
 ## [0.25.0] - 2026-05-21
 
 ### Added
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -460,6 +460,105 @@ function applyBuiltinOverrides(
 	});
 }
 
+// Custom (user/project) agents pre-fill fields from their frontmatter, so
+// agentOverrides[name] only fills fields the frontmatter left unset. This
+// keeps frontmatter as the per-agent source of truth and lets settings.json
+// supply per-harness defaults (model, thinking, ...) for shared personas.
+// Bulk disableBuiltins intentionally does not touch custom agents.
+function applyCustomAgentOverride(
+	agent: AgentConfig,
+	override: BuiltinAgentOverrideConfig,
+	meta: { scope: "user" | "project"; path: string },
+): AgentConfig {
+	const next: AgentConfig = { ...agent };
+	let anyFilled = false;
+
+	const fill = <T>(field: keyof AgentConfig, currentlyUnset: boolean, value: T) => {
+		if (!currentlyUnset) return;
+		(next as Record<string, unknown>)[field as string] = value;
+		anyFilled = true;
+	};
+
+	if (override.model !== undefined) {
+		fill("model", agent.model === undefined, override.model === false ? undefined : override.model);
+	}
+	if (override.fallbackModels !== undefined) {
+		fill(
+			"fallbackModels",
+			agent.fallbackModels === undefined,
+			override.fallbackModels === false ? undefined : [...override.fallbackModels],
+		);
+	}
+	if (override.thinking !== undefined) {
+		fill("thinking", agent.thinking === undefined, override.thinking === false ? undefined : override.thinking);
+	}
+	if (override.systemPromptMode !== undefined) {
+		fill("systemPromptMode", agent.systemPromptMode === undefined, override.systemPromptMode);
+	}
+	if (override.inheritProjectContext !== undefined) {
+		fill("inheritProjectContext", agent.inheritProjectContext === undefined, override.inheritProjectContext);
+	}
+	if (override.inheritSkills !== undefined) {
+		fill("inheritSkills", agent.inheritSkills === undefined, override.inheritSkills);
+	}
+	if (override.defaultContext !== undefined) {
+		fill(
+			"defaultContext",
+			agent.defaultContext === undefined,
+			override.defaultContext === false ? undefined : override.defaultContext,
+		);
+	}
+	if (override.disabled !== undefined) {
+		fill("disabled", agent.disabled === undefined, override.disabled);
+	}
+	if (override.systemPrompt !== undefined) {
+		fill("systemPrompt", agent.systemPrompt === undefined, override.systemPrompt);
+	}
+	if (override.skills !== undefined) {
+		fill(
+			"skills",
+			agent.skills === undefined,
+			override.skills === false ? undefined : [...override.skills],
+		);
+	}
+	if (override.tools !== undefined) {
+		const toolsUnset = agent.tools === undefined && agent.mcpDirectTools === undefined;
+		if (toolsUnset) {
+			const { tools, mcpDirectTools } = splitToolList(override.tools === false ? [] : override.tools);
+			next.tools = tools;
+			next.mcpDirectTools = mcpDirectTools;
+			anyFilled = true;
+		}
+	}
+	if (override.completionGuard !== undefined) {
+		fill("completionGuard", agent.completionGuard === undefined, override.completionGuard);
+	}
+
+	if (!anyFilled) return agent;
+	next.override = { ...meta, base: cloneOverrideBase(agent) };
+	return next;
+}
+
+function applyCustomAgentOverrides(
+	agents: AgentConfig[],
+	userSettings: SubagentSettings,
+	projectSettings: SubagentSettings,
+	userSettingsPath: string,
+	projectSettingsPath: string | null,
+): AgentConfig[] {
+	return agents.map((agent) => {
+		const projectOverride = projectSettings.overrides[agent.name];
+		if (projectOverride && projectSettingsPath) {
+			return applyCustomAgentOverride(agent, projectOverride, { scope: "project", path: projectSettingsPath });
+		}
+		const userOverride = userSettings.overrides[agent.name];
+		if (userOverride) {
+			return applyCustomAgentOverride(agent, userOverride, { scope: "user", path: userSettingsPath });
+		}
+		return agent;
+	});
+}
+
 export function buildBuiltinOverrideConfig(
 	base: BuiltinAgentOverrideBase,
 	draft: Pick<AgentConfig, "model" | "fallbackModels" | "thinking" | "systemPromptMode" | "inheritProjectContext" | "inheritSkills" | "defaultContext" | "disabled" | "systemPrompt" | "skills" | "tools" | "mcpDirectTools" | "completionGuard">,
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks for the PR. The fix is still useful, but I would not merge this branch as-is because it no longer merges with current main.

Current main has changed the same agent loader path since this was opened, including package-agent discovery, extra user agent dirs, and subagentOnlyExtensions in overrides. Could you rebase and keep the change focused on applying agentOverrides to the updated custom-agent discovery path, preserving those current fields?

When rebasing, please also update the Unreleased changelog entry so it credits Jacek Juraszek (@jjuraszek) for this PR, matching the current changelog style. The focused tests on your branch cover the intended model case; after the refresh this should be a small review.

nicobailon:
I'm closing this for now while we clean up the PR queue. The agent override fix still looks useful, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #226, opened 2026-05-28
Auto-add unique runId/index suffix to default output paths

smithyyang:
**Problem:** When subagent output comes from agent config default (e.g. ``output: /tmp/research.md``), parallel tasks write to the same path, and the model may follow hardcoded prompt filenames instead of the runtime output override.

**Fix:** When output is from agent default (no task-level override), auto-append ``_<runId>_<index>`` to the filename at runtime.
- ``/tmp/research.md`` → ``/tmp/research_abc123_0.md``
- Unique per task, chain step, or async run
- Explicit ``output: "/tmp/custom.md"`` stays exact (not modified)
- Progress tracks in ``.progress_<runId>/progress.md`` under the same directory

**Affects:** foreground single, foreground parallel, async single, async chain — all execution paths.

**Validation:** 3-researcher parallel run produced ``/tmp/research_b90ae4e0_0.md``, ``_1.md``, ``_2.md`` without conflict.
No ``~/`` pollution. All files in ``/tmp/``.

**Follow-up:** progress naming could use ``progress_<runId>.md`` for consistency (currently uses hidden subdirectory).

### DIFF
--- a/agents/researcher.md
+++ b/agents/researcher.md
@@ -7,7 +7,7 @@ systemPromptMode: replace
 inheritProjectContext: true
 inheritSkills: false
 output: research.md
-defaultProgress: true
+defaultProgress: false
 ---
 
 You are a research subagent.
--- a/agents/scout.md
+++ b/agents/scout.md
@@ -7,7 +7,7 @@ systemPromptMode: replace
 inheritProjectContext: true
 inheritSkills: false
 output: context.md
-defaultProgress: true
+defaultProgress: false
 ---
 
 You are a scouting subagent running inside pi.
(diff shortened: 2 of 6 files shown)


nicobailon:
Thanks for sending this. I’m closing this older branch because the follow-up #227 became the better place for this work: it carries the authoritative output-path instruction and wider async/chain coverage that this branch is missing.

This branch also no longer merges cleanly with current main, and it changes bundled defaultProgress as a workaround rather than fixing the progress path directly. If the output-path work comes back, the useful shape would be a fresh smaller PR against current main with focused regressions.

nicobailon:
One extra note for a future revival: I do not see a merged replacement carrying this change, and the current Unreleased changelog does not credit @smithyyang for this output-path work. A fresh PR should include that small changelog entry too.

PR #227, opened 2026-05-28
Make default output paths unique and authoritative across all run modes

smithyyang:
## Summary

This PR fixes several output-path and UI issues across foreground, async, parallel, and chain subagent execution paths.

It makes runtime output paths authoritative and unique when they come from agent defaults, while preserving exact user-specified output paths.

### 1. Prevent output collisions for default agent outputs

When an agent default output path is used, the runtime now appends a unique suffix:

```txt
/tmp/research.md -> /tmp/research_<runId>_<index>.md
```

This prevents multiple subagents from writing to the same file in parallel/chain runs.

Covered paths:

- foreground single
- foreground top-level parallel
- foreground chain sequential steps
- foreground chain parallel steps
- async single
- async chain
- async parallel

Explicit user output paths remain exact and are not modified:

```ts
subagent({ agent: "researcher", output: "/tmp/custom.md", task: "..." })
```

### 2. Fix async parallel index collision

Previously, async parallel default outputs all used index `0`, causing two researchers to write to the same file:

```txt
/tmp/research_<runId>_0.md
/tmp/research_<runId>_0.md
```

This PR passes the real flat step index, producing:

```txt
/tmp/research_<runId>_0.md
/tmp/research_<runId>_1.md
```

### 3. Fix async single default-output suffixing

Async single runs pre-normalized the output before calling `executeAsyncSingle`, so the lower layer could not tell whether the path came from an agent default or an explicit user override.

This PR adds an `outputFromAgentDefault` flag so async single can suffix default outputs correctly without changing explicit outputs.

### 4. Make runtime output paths authoritative in prompts

Agent prompts may mention static filenames such as `research.md`, `context.md`, or `plan.md`. Runtime-generated paths must override these static filenames.

This PR strengthens task output instructions from:

```md
**Output:** Write your findings to: /tmp/research_xxx_0.md
```

to:

```md
**Output:**
Write your findings to exactly this path: /tmp/research_xxx_0.md
This path is authoritative for this run.
Ignore any other filename mentioned elsewhere, including filenames in the base agent prompt, system prompt, or task instructions.
```

It also injects the same runtime output override into the child system prompt, so the system prompt and task prompt agree on the authoritative output path.

### 5. Disable default progress files for built-in writer agents

The built-in `researcher`, `scout`, and `worker` agents now set:

```yaml
defaultProgress: false
```

This avoids automatic progress file creation unless the caller explicitly asks for progress.

### 6. Disable chain clarify UI by default

Sequential foreground chains previously opened the clarify UI by default. This PR changes chain clarify behavior so the UI only opens when explicitly requested:

```ts
clarify: true
```

Default behavior now matches single/parallel runs: execute directly unless clarify is requested.

### 7. Show model badge in compact multi-result view

Compact multi-result rendering now includes the model/thinking badge for each result row.

## Validation

Manual validation performed locally:

- foreground parallel with 3 researchers created unique files:
  - `/tmp/research_<runId>_0.md`
  - `/tmp/research_<runId>_1.md`
  - `/tmp/research_<runId>_2.md`
- async parallel with 2 researchers now injects and saves to `_0.md` and `_1.md` respectively
- async single now injects and saves to `_0.md`
- foreground chain no longer opens clarify UI by default
- foreground chain step 1 wrote to `_0.md`, and step 2 received `_1.md` in its task instructions
- task prompt contains authoritative output instruction
- system prompt injection path is wired through foreground `runSync` and async step construction
- TypeScript syntax checks passed for changed files

## Notes

This PR intentionally preserves backwards compatibility:

- explicit `output` paths are used exactly as provided
- `output: false` disables output
- `output: true` means “use the agent default” and receives the runtime suffix

### DIFF
--- a/agents/researcher.md
+++ b/agents/researcher.md
@@ -7,7 +7,7 @@ systemPromptMode: replace
 inheritProjectContext: true
 inheritSkills: false
 output: research.md
-defaultProgress: true
+defaultProgress: false
 ---
 
 You are a research subagent.
--- a/agents/scout.md
+++ b/agents/scout.md
@@ -7,7 +7,7 @@ systemPromptMode: replace
 inheritProjectContext: true
 inheritSkills: false
 output: context.md
-defaultProgress: true
+defaultProgress: false
 ---
 
 You are a scouting subagent running inside pi.
--- a/agents/worker.md
+++ b/agents/worker.md
@@ -8,7 +8,7 @@ inheritSkills: false
 tools: read, grep, find, ls, bash, edit, write, contact_supervisor
 defaultContext: fork
 defaultReads: context.md, plan.md
-defaultProgress: true
+defaultProgress: false
 ---
 
 You are `worker`: the implementation subagent.
--- a/src/extension/schemas.ts
+++ b/src/extension/schemas.ts
@@ -19,7 +19,7 @@ const OutputOverride = Type.Unsafe({
 		{ type: "string" },
 		{ type: "boolean" },
 	],
-	description: "Output filename/path (string), or false to disable file output",
+	description: "Output filename/path (string), or false to disable file output. If explicitly setting a path, prefer an absolute /tmp/... path; omit output to use the agent default auto-unique output path.",
 });
 
 const OutputModeOverride = Type.String({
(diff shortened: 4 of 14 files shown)


nicobailon:
Thanks for putting this together. I’m closing this PR because it is too stale to merge against current main. GitHub marks it conflicting, and applying the branch now would roll back a lot of newer runtime work and tests rather than just add the output-path fixes.

There are useful ideas here, especially making runtime output paths unambiguous, but the branch also misses `output: true` in several paths where the PR says it should behave like an agent default. Some pieces have since been handled differently on main, including duplicate parallel output detection and the compact model badge.

A fresh smaller PR on current main would be the right shape for any remaining output-path behavior, with coverage for omitted output versus `output: true`, explicit paths, and the relevant foreground, async, and chain paths.

nicobailon:
One extra changelog note after checking current main: there is not a merged replacement for this branch, so there is no replacement changelog entry or contributor credit yet.

If any of the remaining output-path work comes back in a fresh PR, it should include an Unreleased entry and credit youngshine (@smithyyang).

PR #229, opened 2026-05-28
fix: resolve subagent context per agent defaultContext

### LINKED ISSUE #228: Per-agent defaultContext ignored when mixing fork-default and fresh-default agents
## Summary

When `context` is omitted on a `subagent(...)` call, pi-subagents currently promotes the **entire** invocation to `fork` if **any** requested agent has `defaultContext: "fork"`.

That causes read-only agents configured with `defaultContext: fresh` (for example `scout`, `reviewer`) to inherit the full parent transcript when batched with `worker` or `oracle` in parallel or chain mode.

## Reproduction

1. Configure `scout` with `defaultContext: fresh`.
2. Configure `worker` with `defaultContext: fork`.
3. Run parallel subagent call without explicit `context`:

```json
{
  "tasks": [
    { "agent": "scout", "task": "Find relevant files" },
    { "agent": "worker", "task": "Implement fix" }
  ]
}
```

4. Observe both tasks run with forked/inherited parent context.

Root cause: `applyAgentDefaultContext()` in `src/runs/foreground/subagent-executor.ts`.

## Expected behavior

When caller omits top-level `context`:

- Each agent/task/step should use **its own** `defaultContext`.
- Explicit `context: "fresh"` or `context: "fork"` should override all agents in that call.
- Parallel scout + worker should fork only the worker task, not the scout.

## Proposed fix

- Remove whole-invocation fork promotion.
- Resolve context per flat task index via agent `defaultContext`.
- Fork session files and `wrapForkTask()` only for indices whose resolved context is `fork`.
- Update tool docs/schema descriptions accordingly.

## Impact

Users relying on implicit whole-invocation fork when mixing fork-default and fresh-default agents in one call would need to pass explicit `context: "fork"` for that behavior.

fitchmultz:
Fixes #228

## Summary

- Remove whole-invocation fork promotion when any agent in a parallel/chain call has `defaultContext: fork`.
- Resolve context per flat task index from each agent's `defaultContext` (or explicit caller `context`).
- Fork session files and `wrapForkTask()` only for indices whose resolved context is `fork`.
- Keep invocation-level fork metadata when any agent uses fork (intercom/TUI badge behavior).
- Update the local Pi development baseline to `@earendil-works/*` `0.77.0` and `typebox` `1.1.39`.
- Add Pi 0.77 tool prompt metadata (`promptSnippet` / `promptGuidelines`) for parent and child-safe `subagent` tool registrations.
- Switch the local TS test loader to `jiti/register` so tests keep working on Node releases where `--experimental-transform-types` is removed.
- Keep Pi runtime packages as optional wildcard peers so future Pi versions are not blocked by npm peer ranges.

## Why

Mixed scout + worker invocations were inheriting the full parent transcript for read-only scouts, inflating input tokens and cost even when scout frontmatter already specified `defaultContext: fresh`.

The Pi 0.77 refresh keeps the branch current with the latest package and tool-metadata guidance while preserving forward-open install behavior.

## Test plan

- [x] `npm test` / `npm run test:unit` (472 unit tests)
- [x] `npm pack --dry-run`
- [x] `npm publish --dry-run`
- [x] `pi install git:github.com/fitchmultz/pi-subagents@fix/per-agent-context-policy` verified `pi-subagents@0.25.1`

## Release / publish note

`pi-subagents@0.25.1` is ready from this branch, but npm publish is maintainer-gated. Publishing with the available `fitchmultz` npm token fails with:

```text
npm error code E403
npm error 403 Forbidden - PUT https://registry.npmjs.org/pi-subagents - You do not have permission to publish "pi-subagents".
```

Current npm maintainer is `nicopreme <nico.bailon@gmail.com>`, so the upstream maintainer needs to publish `pi-subagents@0.25.1` after merge.

## Related issue

https://github.com/nicobailon/pi-subagents/issues/228

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ AGENTS.md
 node_modules/
 .fallow/
 
+context.md
 package-lock.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+## [0.25.1] - 2026-05-28
+
+### Changed
+- Updated the local pi development baseline to `@earendil-works/*` `0.77.0` in package metadata.
+- Declared all pi-bundled runtime packages (`@earendil-works/pi-*`, `typebox`) as optional wildcard peers and kept them only as development dependencies, so npm peer ranges do not block future pi releases.
+- Added explicit `subagent` tool prompt snippet/guidelines so Pi 0.77+ tool metadata attribution has concise delegation guidance separate from the full schema description.
+- Switched the local TypeScript test loader to `jiti` so validation does not depend on Node's removed `--experimental-transform-types` flag on newer Node releases.
+
 ## [0.25.0] - 2026-05-21
 
 ### Added
--- a/docs/upstream-issue-per-agent-context.md
+++ b/docs/upstream-issue-per-agent-context.md
@@ -1,0 +1,45 @@
+# Issue: Whole-invocation fork promotion ignores per-agent `defaultContext: fresh`
+
+## Summary
+
+When `context` is omitted on a `subagent(...)` call, pi-subagents currently promotes the **entire** invocation to `fork` if **any** requested agent has `defaultContext: "fork"`.
+
+That causes read-only agents configured with `defaultContext: fresh` (for example `scout`, `reviewer`) to inherit the full parent transcript when batched with `worker` or `oracle` in parallel or chain mode.
+
+## Reproduction
+
+1. Configure `scout` with `defaultContext: fresh`.
+2. Configure `worker` with `defaultContext: fork`.
+3. Run parallel subagent call without explicit `context`:
+
+```json
+{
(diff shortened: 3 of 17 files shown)


nicobailon:
I would not merge this as-is. The issue is real and the per-agent context direction makes sense, but the branch now conflicts with main and it changes fork failure behavior in a risky way.

With fork session creation delayed until each child asks for a session file, a mixed fresh + fork run can start the fresh child before learning that the forked child cannot branch because the parent session or leaf is missing. The previous behavior failed before launching any children. Please rebase and make those fork checks happen before any foreground parallel or chain children start.

nicobailon:
I'm closing this for now while we clean up the PR queue. The per-agent context work is still worth revisiting, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #230, opened 2026-05-29
docs(agents): mark contact_supervisor as runtime-bridge-injected in bundled agent prose

JayceFreeman:
The bundled `scout` / `planner` / `reviewer` / `oracle` / `researcher` agents reference `contact_supervisor` in their **Supervisor coordination** prose, but it is intentionally absent from their `tools:` frontmatter. The intercom runtime bridge injects `contact_supervisor` at dispatch time (`intercom-bridge.ts`), so the runtime behavior is correct — only the prose was silent about *why* the tool is not declared, which reads as a lying contract against the frontmatter.

This marks each first mention as bridge-injected (one line per file, `+5/-5` total):

> use the bridge-injected `contact_supervisor` (provided at runtime, so it is not in this agent's `tools:` list) ...

No behavior change — documentation only.

### DIFF
--- a/agents/oracle.md
+++ b/agents/oracle.md
@@ -15,7 +15,7 @@ Your primary job is to prevent the main agent from making hidden, conflicting, o
 
 Before you do anything else, reconstruct the key inherited decisions, constraints, and open questions from the forked conversation, codebase state, and task. Those decisions form your baseline contract. Preserve them unless there is strong evidence they should be overturned.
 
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks, this direction makes sense: the bridge does add `contact_supervisor` at runtime, and this wording explains why it is missing from the five agent frontmatter lists.

I would not merge it as-is yet because `agents/context-builder.md` has the same supervisor coordination wording and also omits `contact_supervisor` from its `tools:` list. Please include that file too so the bundled agent docs stay consistent.

One more small thing before this can be merge-ready: please add an `Unreleased` changelog entry for the docs clarification and credit Jayce Freeman (`@JayceFreeman`) for #230. The PR branch does not touch `CHANGELOG.md` yet.

nicobailon:
I'm closing this for now while we clean up the PR queue. The docs direction is fine, but this branch still needs the missing bundled-agent wording and changelog before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #231, opened 2026-05-29
feat(run-history): RunEntry v2 (model/cwd/tool_calls/error_excerpt) + lock-free rotation; extend BuiltinAgentOverrideConfig (output/defaultReads/defaultProgress)

JayceFreeman:
Two related changes on the same `agents.ts` / `run-history.ts` surface. 477/477 unit tests pass (incl. the new ones).

## RunEntry v2 + lock-free rotation

`RunEntry` gains optional `model` / `cwd` / `tool_calls` / `error_excerpt`, sourced at each of the **4 `recordRun` call sites** from the values actually used to run the child — never the parent `ctx.cwd`:

| field | source |
|---|---|
| `model` | `SingleResult.model` |
| `cwd` | the childs executed cwd — `taskCwd` (parallel chain), `stepCwd` (sequential chain), `resolveParallelTaskCwd(...)` (parallel path), `effectiveCwd` (single path) |
| `tool_calls` | `progressSummary.toolCount` (preserves `0`; not dropped by truthiness) |
| `error_excerpt` | `SingleResult.error`, truncated to 300 chars, only on error |

**Rotation is now lock-free.** The old read-path trim raced under concurrent processes. New design: `recordRun` appends one JSON line via `O_APPEND` (kernel-atomic across processes on a local POSIX fs; entries are small → single `write()`), then a best-effort `maybeRotate` trims to `ROTATE_KEEP` by writing a per-process-unique temp and **atomic-renaming** it into place. The file is a complete valid snapshot either way — no duplication, no corruption. The only loss this admits is an append landing inside a rotates read→rename window; acceptable for telemetry. No new dependency, no lock file. A **16-process × 150-write storm test** guards no-corruption / no-double-write / bounded-size.

## BuiltinAgentOverrideConfig: output / defaultReads / defaultProgress

These three fields are honored on `AgentConfig` but were silently dropped by the override schema, parser, applier, **and the write path** — so a `settings.json` override setting them was a no-op (a lying contract). Now threaded through all of: schema (`BuiltinAgentOverrideBase` + `Config`), `parseBuiltinOverrideEntry` (typed validation errors), `applyBuiltinOverride` (the `false` sentinel clears the builtin default), and the **save path** (`cloneOverrideBase` / `cloneOverrideValue` / `buildBuiltinOverrideConfig`) so values round-trip through save+reload rather than being dropped on persist.

### Tests
+6 run-history unit tests (incl. multi-process storm) and +7 agent-overrides tests (apply, `false`-sentinel clear, malformed-value rejection, builder round-trip, save+reload persistence).

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -45,6 +45,9 @@ export interface BuiltinAgentOverrideBase {
 	systemPrompt: string;
 	skills?: string[];
 	tools?: string[];
+	output?: string;
+	defaultReads?: string[];
+	defaultProgress?: boolean;
 	mcpDirectTools?: string[];
 	completionGuard?: boolean;
 }
@@ -61,6 +64,9 @@ interface BuiltinAgentOverrideConfig {
 	systemPrompt?: string;
 	skills?: string[] | false;
 	tools?: string[] | false;
+	output?: string | false;
+	defaultReads?: string[] | false;
+	defaultProgress?: boolean;
 	completionGuard?: boolean;
 }
 
@@ -185,6 +191,9 @@ function cloneOverrideBase(agent: AgentConfig): BuiltinAgentOverrideBase {
 		systemPrompt: agent.systemPrompt,
 		skills: agent.skills ? [...agent.skills] : undefined,
 		tools: agent.tools ? [...agent.tools] : undefined,
+		output: agent.output,
+		defaultReads: agent.defaultReads ? [...agent.defaultReads] : undefined,
+		defaultProgress: agent.defaultProgress,
 		mcpDirectTools: agent.mcpDirectTools ? [...agent.mcpDirectTools] : undefined,
 		completionGuard: agent.completionGuard,
 	};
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks for this. I would not merge this branch as-is now because it is based on the May 29 mainline and conflicts with the current code in agents.ts and chain-execution.ts. It also predates the newer chain and status work, so resolving it would need a careful current-main rework rather than a straight merge.

The useful pieces still look real: current main still does not persist output, defaultReads, and defaultProgress through builtin overrides, and run history still only records the old basic fields.

The smallest mergeable path would be a fresh PR against current main, ideally split into the override persistence fix and the run-history metadata/rotation fix. For the run-history piece, please also keep recording cheap in the common case and keep reads bounded for an already-oversized history file.

nicobailon:
I'm closing this for now while we clean up the PR queue. The run-history and override pieces may still be useful, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #232, opened 2026-05-29
Fix EPERM crash on Windows ACL corruption + enable DIRS mutable container for fallback paths

NOirBRight:
## Problem

Pi crashes on startup on Windows Azure AD/Entra ID machines when the NTFS ACL on the global temp directory is corrupted (typically after wake-from-sleep). `ensureAccessibleDir` in `src/extension/index.ts` throws an uncaught EPERM on `mkdirSync`, which propagates up through Pi's extension loader and kills the entire Pi process.

A secondary issue: `RESULTS_DIR` and `ASYNC_DIR` are exported as `export const` (primitive string) bindings, making them read-only in all ES module consumers. A fallback path set by `ensureAccessibleDir` cannot propagate to these consumer modules.

## Changes

1. EPERM-tolerant `ensureAccessibleDir` - catches EPERM/EACCES, attempts recovery, falls back to pid-scoped path
2. DIRS mutable container - `export const DIRS = { results, async, chain, artifacts }` replaces direct RESULTS_DIR/ASYNC_DIR usage; object properties are mutable while the binding remains const (ES module compatible)
3. All 11 consumer module sites migrated to DIRS.results/DIRS.async
4. 9 new unit tests pass

## Files changed (11)
- src/extension/index.ts, src/shared/types.ts, src/extension/doctor.ts
- src/runs/background/async-execution.ts, async-job-tracker.ts, async-resume.ts, run-id-resolver.ts, run-status.ts, stale-run-reconciler.ts
- src/runs/shared/nested-events.ts
- test/unit/ensure-accessible-dir.test.ts (new)

### DIFF
--- a/src/extension/doctor.ts
+++ b/src/extension/doctor.ts
@@ -5,9 +5,8 @@ import { isAsyncAvailable } from "../runs/background/async-execution.ts";
 import { diagnoseIntercomBridge, type IntercomBridgeDiagnostic } from "../intercom/intercom-bridge.ts";
 import { discoverAvailableSkills, type SkillSource } from "../agents/skills.ts";
 import {
-	ASYNC_DIR,
 	CHAIN_RUNS_DIR,
-	RESULTS_DIR,
+	DIRS,
 	TEMP_ROOT_DIR,
 	type ExtensionConfig,
 	type SubagentState,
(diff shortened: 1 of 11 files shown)


nicobailon:
Thanks for sending this. The Windows startup failure is a real thing to fix, but I would not merge this PR as-is.

It now conflicts with current main in the async execution and resume code, and main has newer async path consumers that this branch does not update, including foreground chain attachment. There is also a concrete bug in the patch: nested-events.ts calls DIRS.async/results without importing DIRS.

Could you rebase this on current main and keep the change focused on the directory fallback? The regression should exercise the production helper and an actual EPERM/EACCES fallback path rather than copying ensureAccessibleDir into the test.

nicobailon:
I'm closing this for now while we clean up the PR queue. The Windows startup resilience issue is real, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #237, opened 2026-05-31
fix: animate spinner in foreground and async subagent widgets

NOirBRight:
## Problem

The spinner (running glyph) in both foreground subagent results and async widget frequently freezes during long-running subagent tasks — especially when the child agent is waiting for LLM responses with no status file updates.

## Root Cause

1. **No animation driver**: The spinner glyph was computed from static state values (seed = sum of toolCount/durationMs/etc). When the subagent went idle waiting for LLM, none of these values changed, so the rendered glyph stayed frozen on the same frame.
2. **Static component tree**:  created a  of  nodes at render time. The text content was baked in and never refreshed unless  changed (which only happens on status file updates).
3. **Missing foreground timer**: The foreground  had no mechanism to trigger periodic re-renders while the subagent was running.

### 
- ****: Now mixes  into the seed, so the selected Braille frame rotates continuously (~8fps) even when underlying state is unchanged.
- ****: Returns a dynamic component whose  callback recalculates all lines on every TUI frame, instead of pre-computing a static .
- **Widget animation timer**: Added  / . When any async job is , a  calls  to keep the widget refreshing. Cleared automatically when all jobs finish.
- ****: Added to the widget component for pi-tui contract compliance.

### 
- ****: Extracted a typed interface for the render context, replacing scattered  assertions.
- ****: When the result still has running children, creates a  that calls . The timer is automatically cleaned up via a  hook on the returned component (called by pi-tui when the tool result is removed from the chat).

## Verification

- Async widget spinner rotates smoothly during 20–30 second LLM waits.
- Foreground result spinner rotates smoothly while the child agent streams output.
- Both timers are cleaned up after the subagent completes (no leaks observed in session logs).

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -179,6 +179,11 @@ function parseSubagentNotifyContent(content: string): SubagentNotifyDetails | un
 	};
 }
 
+interface ResultRenderContext {
+	state: { subagentResultAnimationTimer?: ReturnType<typeof setInterval> };
+	invalidate?: () => void;
+}
+
 class SubagentControlNoticeComponent implements Component {
 	constructor(
 		private readonly details: SubagentControlMessageDetails,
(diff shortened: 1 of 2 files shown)


NOirBRight:
Closing in favor of #244, which carries the spinner animation fix together with the session-scoped temp directory changes currently needed by the fork.

PR #244, opened 2026-06-02
fix: keep subagent spinners animating with session-scoped dirs

NOirBRight:
## Summary

- combines the session-scoped temp directory fixes currently used by the fork with the spinner animation fix from #237
- keeps running glyphs animated by mixing a time seed into the spinner frame selection
- renders async widgets dynamically per TUI frame and drives rerenders while jobs are running
- adds/updates regression coverage for unchanged progress seeds and running widget animation ticks

## Verification

- node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/render-widget.test.ts
- node --experimental-strip-types --test test/unit/session-scoped-dirs.test.ts
- node --experimental-strip-types --test test/unit/ensure-accessible-dir.test.ts

## Notes

This is intended as the combined upstreamable version of the spinner fix after the fork temporarily pinned session-scoped temp directory work. It supersedes the isolated spinner branch in #237 for users that need both fixes.

### DIFF
--- a/src/extension/doctor.ts
+++ b/src/extension/doctor.ts
@@ -5,9 +5,8 @@ import { isAsyncAvailable } from "../runs/background/async-execution.ts";
 import { diagnoseIntercomBridge, type IntercomBridgeDiagnostic } from "../intercom/intercom-bridge.ts";
 import { discoverAvailableSkills, type SkillSource } from "../agents/skills.ts";
 import {
-	ASYNC_DIR,
 	CHAIN_RUNS_DIR,
-	RESULTS_DIR,
+	DIRS,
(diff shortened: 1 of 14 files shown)


nicobailon:
I would not merge this combined PR now. It conflicts with current main, and the session-scoped directory change still leaves the result watcher and async job tracker pointed at the old results directory, so completions can be missed. The session suffix also comes from the first characters of the session file path, which is not unique across normal sessions.

The spinner part also brings back timer-driven redraws that current main later moved away from. I am closing this stale branch; the useful pieces are worth fresh smaller PRs, with the EPERM/DIRS work continuing separately in #232 if it is still wanted.

PR #248, opened 2026-06-03
fix: remove allOf/anyOf/const/if/then from tool schemas for provider compat

sergio-agosti:
## Problem

Providers with limited JSON Schema support (kimi/Moonshot, Antigravity proxy, and potentially others) reject the `subagent` tool registration because the parameter schema contains keywords not supported by their minimal JSON Schema parsers:

- **`AcceptanceOverride`**: Uses `anyOf` with 3 alternatives (string enum, `const: false`, nested object with nested `anyOf` in criteria.items and review)
- **`ChainItem`**: Uses `allOf` with `if`/`then`/`not` conditional constraints

kimi returns:
```
Error: 400 Error from provider: JSON Schema not supported: could not understand the instance {required: [criteria]}
```
then:
```
Error: 400 Error from provider: Conflict in schema definitions for key if. Previous: {required: [expand]}, New: {required: [collect]}
```

Antigravity proxy (Google Cloud Code Assist): same class of error with `anyOf`/`const` (previously noted in CHANGELOG).

## Fix

Two changes in `src/extension/schemas.ts`:

### 1. `AcceptanceOverride` — no `anyOf`/`allOf`/`const` keywords

| Before | After |
|---|---|
| Top-level `anyOf` with 3 alternatives (string, `const: false`, object) | `type: ["string", "boolean", "object"]` — same expressiveness, basic JSON Schema |
| `criteria.items.anyOf` with string/AcceptanceGateSchema | `type: ["string", "object"]` with inline properties (`id`, `must`, `evidence`, `severity`) and `required: ["id", "must"]` |
| `review.anyOf` with `const: false` / AcceptaceReviewGateSchema | `type: ["boolean", "object"]` with inline properties (`agent`, `focus`, `required`) |

All field-level type constraints are **preserved** — `verify` must be array, `stopRules` must be array, `criteria` items must be string or object with `id`+`must`, etc. A malformed input like `{ verify: "npm test" }` still fails schema validation because `verify.type` is `"array"`.

### 2. `ChainItem` — remove `allOf` with `if`/`then`/`not`

Replace the conditional schema (expand requires parallel+collect, collect requires expand+parallel, expand without parallel type=array) with `additionalProperties: true`. The expand/collect/parallel conditional logic is enforced by runtime execution code.

## Verification

- Tested with kimi (moonshot/kimi-k2.6) via OpenCode Go subscription: `subagent()` calls with acceptance blocks work without 400 errors
- Structural validation preserved: malformed acceptance objects (wrong types) still rejected by schema
- All runtime validation paths (`validateAcceptanceInput()`, `normalizeAcceptanceInput()`) unchanged
- No `allOf`/`anyOf`/`const`/`if`/`then`/`not` in AcceptanceOverride or ChainItem schemas

## Related

- CHANGELOG previously noted `anyOf`/`const` issues with Antigravity proxy
- This extends the fix to cover `allOf`/`if`/`then`/`not` keywords and applies to the acceptance schema

### DIFF
--- a/src/extension/schemas.ts
+++ b/src/extension/schemas.ts
@@ -77,37 +77,59 @@ const AcceptanceReviewGateSchema = Type.Object({
 	required: Type.Optional(Type.Boolean()),
 }, { additionalProperties: false });
 
+// Uses type unions (type:[...]) instead of anyOf/allOf/const to avoid
+// JSON Schema keywords that some provider APIs (kimi/Moonshot,
+// Antigravity proxy) do not support in tool function.parameters.
+// Validation for malformed inputs is handled at runtime by
+// validateAcceptanceInput() / normalizeAcceptanceInput().
 const AcceptanceOverride = Type.Unsafe({
-	anyOf: [
-		{ type: "string", enum: ["auto", "none", "attested", "checked", "verified", "reviewed"] },
-		{ const: false },
-		{
-			type: "object",
+	type: ["string", "boolean", "object"],
+	properties: {
+		level: { type: "string", enum: ["auto", "none", "attested", "checked", "verified", "reviewed"] },
+		criteria: {
+			type: "array",
+			items: {
+				type: ["string", "object"],
+				properties: {
+					id: { type: "string" },
+					must: { type: "string" },
+					evidence: { type: "array", items: { type: "string", enum: ["changed-files", "tests-added", "commands-run", "validation-output", "residual-risks", "no-staged-files", "diff-summary", "review-findings", "manual-notes"] } },
+					severity: { type: "string", enum: ["required", "recommended"] },
+				},
+				required: ["id", "must"],
+				additionalProperties: false,
+			},
+		},
+		evidence: { type: "array", items: { type: "string", enum: ["changed-files", "tests-added", "commands-run", "validation-output", "residual-risks", "no-staged-files", "diff-summary", "review-findings", "manual-notes"] } },
+		verify: {
+			type: "array",
+			items: {
+				type: "object",
+				properties: {
+					id: { type: "string" },
+					command: { type: "string" },
+					timeoutMs: { type: "integer", minimum: 1 },
+					cwd: { type: "string" },
+					env: { type: "object", additionalProperties: { type: "string" } },
+					allowFailure: { type: "boolean" },
+				},
+				required: ["id", "command"],
+				additionalProperties: false,
+			},
+		},
+		review: {
+			type: ["boolean", "object"],
 			properties: {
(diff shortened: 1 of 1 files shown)


sombraxxl:
Thanks for this fix — hitting the exact same errors on `kimi-k2.6` (provider: `moonshotai`, API: `openai-completions`).

**Confirmed reproduction**

```
Error: 400 Error from provider: Provider returned error
{"error":{"message":"Invalid request: tools.function.parameters is not a valid moonshot flavored json schema, details: <At path 'properties.task': type is not defined>","type":"invalid_request_error"}}
```

After retries:
```
Error: 400 Error from provider: Conflict in schema definitions for key 'if'. Previous: {'required': ['expand']}, New: {'required': ['collect']}
```

**Additional technical context**

Examined the schema generated by TypeBox 1.2.8 and confirmed the root cause:
- `ChainItem.allOf` with two independent `if` clauses → Moonshot collapses them into a single object and detects duplicate `if` keys (even though they are valid under standard JSON Schema as independent array entries).
- `Type.Unsafe({ anyOf: [...] })` in `SkillOverride`, `OutputOverride`, `AcceptanceOverride`, etc. generates properties without a direct `type` field → the "moonshot flavored" validator rejects them.

The direction of this PR (replacing `allOf`/`anyOf`/`const`/`if`/`then`/`not` with `type` unions and `additionalProperties: true`) is the right one. The structural validation lost by the schema relaxation can be recovered at runtime via `validateAcceptanceInput()`, as the P1 review comment suggests.

Happy to help with testing if needed. 👍

nicobailon:
Thanks for taking this on. The provider schema bug is real, and this PR is moving in a useful direction, but I would not merge it as-is.

The current branch relaxes validation beyond the old contract: `acceptance: true`, `review: true`, unknown chain-step fields, and dynamic `expand` steps with array `parallel` can now pass the tool schema. The focused schema regression catches that, and the branch also adds `type: [...]` unions that the current provider-compatibility check rejects.

The smallest useful revision would keep the keyword cleanup while preserving the existing validation shape: only `false` disables acceptance or review, dynamic `expand`/`collect` cannot mix with static parallel arrays, and unknown chain-step fields should stay rejected.

nicobailon:
One small item for the same revision: please add an Unreleased changelog note for the provider schema compatibility fix, with credit to Sergio Agosti (@sergio-agosti).

nicobailon:
I'm closing this for now while we clean up the PR queue. The schema compatibility problem is real, but this branch needs a tighter current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #251, opened 2026-06-04
feat: parallel groups, metadata, group options, completion via /chain

loss-and-quick:
Brings the inline `/chain` slash command to parity with the `ChainStep` model for
everything that fits on a single line: static parallel groups, per-step/per-task
metadata, group-level options, and group-aware tab completion. Dynamic fan-out
(`expand`/`collect`) is intentionally left out — it stays available through the
`subagent({ chain: [...] })` tool API and saved `.chain.json` files.

```text
/chain scout[as=ctx,phase=recon] "scan" -> (reviewer "review A" | reviewer[count=2] "review B")[concurrency=2,failFast] -> writer[outputSchema=schema.json,acceptance=verified] "fix"
```

### Parallel groups (base)
Wrap agents in `( ... )` separated by `|` to fan them out as a single
`{ parallel: [...] }` chain step; the next `->` step continues once they finish.
Linear `/chain` behavior is unchanged.

### Inline step/task metadata
Extended the `[key=value,...]` config with:

| Key | Description |
|-----|-------------|
| `as` | Name this step's output for later reference. |
| `label` | Human-readable step label. |
| `phase` | Group steps into a named phase. |
| `cwd` | Run the step in a subdirectory. |
| `count` | Fan a group task into N copies (only inside a `( ... )` group). |
| `outputSchema` | Path to a JSON Schema file, read relative to cwd and validated. |
| `acceptance` | Acceptance level (`none`/`attested`/`checked`/`verified`/`reviewed`). |

`outputSchema` is validated via `assertJsonSchemaObject`, `acceptance` via
`validateAcceptanceInput`; parse/load errors surface through `ctx.ui.notify`
instead of throwing. Inline values must not contain spaces or commas (unchanged
constraint).

### Group-level options
A trailing `[...]` suffix after the closing `)` sets step-level options on the group:

```text
/chain a "x" -> (b "y" | c "z")[concurrency=2,failFast,worktree]
```

`concurrency`, `failFast`, and `worktree` map onto the emitted parallel step.

### Group-aware tab completion
The multi-agent completion now finds the current step by the last top-level `->`,
`(`, or `|` (quote-aware), so agents are suggested after an opening paren, after a
`|` separator, and on each new arrow step. A bare `->`/`|` gets a separating space;
completion is suppressed inside an open quote or once a `--` task has begun.
Linear-chain completion is unchanged.

## Non-goals
- Dynamic fan-out (`expand`/`collect`) inline — data-driven fan-out relies on JSON
  pointer paths and schemas that don't fit a single slash line.
- Full `acceptance` config objects inline — only the level shorthand is supported;
  use a saved chain for the full structure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,7 @@ Skip this section until you want exact syntax.
 |---------|-------------|
 | `/run <agent> [task]` | Run one agent; omit the task for self-contained agents |
 | `/chain agent1 "task1" -> agent2 "task2"` | Run agents in sequence |
+| `/chain scout "scan" -> (reviewer "A" \| reviewer "B") -> writer "fix"` | Run a chain with a static parallel group inline |
 | `/parallel agent1 "task1" -> agent2 "task2"` | Run agents in parallel |
 | `/run-chain <chainName> -- <task>` | Launch a saved `.chain.md` or `.chain.json` workflow |
 | `/subagents-doctor` | Show read-only setup diagnostics |
@@ -267,6 +268,21 @@ Both double and single quotes work. You can also use `--` as a delimiter:
 ```
 
 Steps without a task inherit behavior from the execution mode. Chain steps get `{previous}`, the prior step’s output. Parallel steps use the first available task as a fallback.
+
+### Inline parallel groups in `/chain`
+
+Wrap a group of agents in parentheses and separate them with `|` to fan them out within a single chain step. The group runs all of its tasks concurrently, then the next `->` step continues once they finish:
+
+```text
+/chain scout "scan" -> (reviewer "review A" | reviewer "review B") -> writer "fix"
+```
+
+Notes:
+
+- Groups must contain at least two tasks separated by ` | `.
+- Group syntax is only valid between ` -> ` separators, and the group must appear as a complete step.
+- A group is treated as the prior step’s output for the next sequential step.
+- Dynamic fanout (`expand` / `collect`) and chain metadata fields are not yet available inline; use the `subagent({ chain: [...] })` tool API or a saved `.chain.json` for those.
 
 ```text
 /chain scout "analyze auth" -> planner -> worker
--- a/README.md
+++ b/README.md
@@ -279,10 +279,25 @@ Wrap a group of agents in parentheses and separate them with `|` to fan them out
 
 Notes:
 
-- Groups must contain at least two tasks separated by ` | `.
+- Groups must contain at least two tasks separated by ` | `, each with its own task.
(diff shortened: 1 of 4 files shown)


elecnix:
Reviewed this — here's what stood out to me that isn't in the PR description:

**Bare `|` doesn't trigger group parsing.** Only `(` does. This is a good backward-compat call since `agent -- do x | y` was already valid and would break otherwise. But worth knowing: if someone writes `a | b` without parens, it silently falls through to the legacy single-agent path with `|` as part of the task. No error, no hint.

**Groups require `->` arrows.** You can't just write `(a "x" | b "y")` alone — it gets rejected. This prevents confusion with `/parallel` but also means there's no shorthand for "just run these in parallel, nothing else."

**`outputSchema` resolves a file path at parse time.** It's not inline JSON — it reads a `.json` file from disk relative to cwd. Feels right for the no-spaces constraint, but it does mean the file has to exist when you type the command, which could surprise someone iterating on a schema.

**`count=N` on a sequential step is silently ignored.** No error, no warning. Makes sense semantically (it's a `ParallelTaskItem` field), but someone might not realize it's not taking effect outside a group.

**The old `parseAgentArgs` got decomposed into small testable functions** (`parseSingleTaskToken`, `parseGroupSegment`, `parseChainExpression`, `buildChainExpressionSteps`). The linear path still goes through `parseAgentArgs` though, so there's some duplication now between the old and new parsing paths.

**Tab completion now walks the prefix with a quote-aware state machine** instead of `lastIndexOf(" -> ")`. This actually changes behavior for linear chains too — the old regex didn't handle quotes at all, so agent names inside quoted tasks could get false completions before.

**New `SlashParseError` class** — nice, separates parse errors from runtime errors and surfaces them via the notification UX instead of crashing.

nicobailon:
I would not merge this as-is. The inline group direction is useful, but this currently changes existing /chain task parsing in a way that would surprise people. A normal shared-task command like /chain scout -- inspect auth (backend) is rejected because the parser treats any unquoted parenthesis as group syntax.

There is also a completion regression outside groups: after task text has started, a bare pipe can make agent completion resume, so scout -- do x | wr suggests writer even though that pipe is part of the task text.

A small revision that keeps the existing -- task behavior for unquoted parentheses and suppresses completion after --, with focused regressions for both cases, would make this much easier to approve.

nicobailon:
I'm closing this for now while we clean up the PR queue. The inline chain syntax work may still be useful, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #255, opened 2026-06-05
feat: pi permission system integration

loss-and-quick:
Add permission-system `ask` forwarding (like `@gotgenes/pi-permission-system`) from subagent child processes back to the parent Pi session UI, preserves per-agent `permission:` frontmatter blocks, and documents the integration.

## Changes

- Add `PI_SUBAGENT_PARENT_SESSION` support for child subprocesses.
- Explicitly propagate parent session identity through:
  - foreground single runs
  - foreground parallel runs
  - foreground chain runs
  - async runner step execution
- Keep inherited `process.env.PI_SUBAGENT_PARENT_SESSION` as a defensive fallback.
- Preserve nested `permission:` YAML frontmatter through agent discovery and serialization.
- Add `/subagents-doctor` diagnostics for permission-system integration:
  - parent session env status
  - subagent process detection
  - permission-system config presence
- Document the integration in `README.md`.
- Add regression coverage for:
  - explicit parent session env precedence
  - inherited env fallback
  - `permission:` frontmatter round-trip preservation
## Why

`pi-permission-system` supports subprocess subagent prompt forwarding through the `PI_SUBAGENT_PARENT_SESSION` environment variable.

Before this PR, subprocess children launched by `pi-subagents` did not reliably receive that parent session identity, so `ask` permissions in child processes could be auto-denied instead of being forwarded to the parent UI.

This PR implements the documented integration contract and makes `permission:` frontmatter safe to use in `pi-subagents` agent files.

## Example

```yaml
---
name: worker
tools: bash,read,write,edit
permission:
  "*": ask
  read: allow
  bash:
    "*": ask
    "git *": allow
    "npm test": allow
---

You are a worker agent.
```

`tools:` continues to control tool visibility via `pi-subagents`.

`permission:` is consumed independently by `@gotgenes/pi-permission-system` as the runtime policy layer.

## Validation

Targeted tests:

```bash
bun test test/unit/pi-args.test.ts test/unit/agent-frontmatter.test.ts test/unit/doctor.test.ts
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,77 @@ For normal use, you do not need to configure anything. Advanced users can tune t
 
 At this point, you know enough to use the plugin. The rest of this README is reference material for exact command syntax, custom agents, saved chains, worktrees, and configuration.
 
+## Optional pi-permission-system integration
+
+[`@gotgenes/pi-permission-system`](https://github.com/gotgenes/pi-packages/tree/main/packages/pi-permission-system)
+adds a second policy layer — `allow` / `ask` / `deny` — on top of
+pi-subagents' visibility-based tool restrictions.
+
+The two compose independently:
+
+| Layer | What it controls | Who provides it |
+|-------|-----------------|-----------------|
+| Visibility | Which tools are registered before the session starts | pi-subagents (`tools:` frontmatter key) |
+| Policy  | Runtime allow/ask/deny decisions on every tool call, bash command, MCP operation | pi-permission-system (`permission:` frontmatter key) |
+
+### Installing
+
+```bash
+pi install npm:@gotgenes/pi-permission-system
+```
+
+No configuration is required for the integration — it is automatic when both
+extensions are installed. pi-subagents passes the parent session identity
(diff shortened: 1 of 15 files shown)


nicobailon:
Thanks for putting this together. The env-var integration is a real gap, but I would not merge this as-is.

The branch is currently conflicting with main, and after running it locally the async dynamic chain path fails because `parentSessionId` is being added to runner step objects and then rejected as an unsupported dynamic fanout field. Async single also still builds runner steps without an explicit `parentSessionId`, so that path relies on the inherited fallback instead of the direct propagation this PR describes.

The smallest useful revision would be to rebase onto current main, carry the parent session through async single and dynamic async children without leaking it into user-facing dynamic validation, and keep the README/frontmatter docs scoped to behavior that the permission system can actually resolve. Then this is worth another look.

nicobailon:
I'm closing this for now while we clean up the PR queue. The permission-system integration is still a real area to revisit, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #258, opened 2026-06-06
fix: prevent skill files from being registered as subagents

### LINKED ISSUE #257: Skill files incorrectly registered as subagents, causing duplicate entries and namespace pollution
## Bug Summary

Pi runtime is incorrectly registering **skill files** (`SKILL.md`) as **executable subagents**, causing `subagent list` output to be bloated with ~100+ entries (including many duplicates) and making it impossible to distinguish real agents from skill stubs.

### 1. Concept conflation: Skill ≠ Agent

- **Skill** = prompt fragment injected into an agent's system prompt. Contains `name/description/metadata/version` frontmatter, no `systemPrompt/tools/model`.
- **Agent** = independent sub-process with its own model, tool allowlist, and context.

Current behavior: Pi scans skill files and registers them as subagents. This is semantically incorrect.

### 2. Duplicate registration

Skill packages install sub-skills that may be discovered through multiple scan paths, causing the same logical skill to be registered twice.

## Reproduction Steps

1. Install a skill package that contains sub-skills.
2. Run `subagent({ action: "list" })`
3. Observe that skill entries appear as `(user)` agents alongside real builtin/project agents.

## Expected Behavior

- `subagent list` should only contain **real agent definitions** (builtin, user/project `.md` files with agent frontmatter: `systemPrompt`, `tools`, `model`, etc.).
- Skill files (`SKILL.md` with `name/description/metadata/version` frontmatter) should **not** appear in subagent list.
- No duplicate entries for the same logical agent/skill.

## Actual Behavior

Skill files are listed as executable agents, inflating the registry and making it unreadable.

## Impact

- **Namespace pollution**: user agents discovered include many skill stubs, not real agents.
- **User confusion**: impossible to tell which entries are runnable agents vs. skill prompt fragments.
- **List bloat**: `subagent list` output becomes unreadable.

## Suggested Fixes

1. **Frontmatter validation**: only `.md` files containing agent-specific frontmatter (or lacking skill-specific keys like `metadata` / `version`) should be registered as subagents. Files with only skill frontmatter should be rejected.
2. **Deduplication**: if the same logical agent/skill exists in multiple scan paths, register it only once.

## Environment

- `pi-subagents` version: latest
- Pi version: 0.78.1
- Doctor output: `agents: total 104 (builtin 8, user 64, project 32)`, `skills: total 54`

chyax98:
## Problem

The agent discovery scanner recursively walks the user agent directory tree looking for agent definition files. Because skill packages install their sub-skills under the same tree, the scanner treats `SKILL.md` files as agent definitions and registers them as executable subagents.

This causes two problems:

1. **Concept conflation**: Skills (prompt fragments meant to be injected into an agent's system prompt) are registered as agents (independent sub-processes with their own model/tools/context).
2. **Registry bloat**: `subagent list` output inflates from ~10 real agents to 100+ entries, making it unreadable and polluting the agent namespace.

## Root cause

`loadAgentsFromDir` validates that a `.md` file has `name` and `description` frontmatter, then registers it as an agent. Skill files also contain `name` and `description`, so they pass validation and get registered.

## Fix

After validating `name` and `description`, check whether the frontmatter also contains a `metadata` or `version` key. These keys are standard in skill files but not used in agent definitions. If either is present, skip the file instead of registering it as an agent.

This is robust regardless of where skill files live in the directory tree — no hard-coded directory names needed.

### Changes

- `src/agents/agents.ts`:
  - `loadAgentsFromDir`: added frontmatter check for `metadata` or `version` keys to filter out skill files

## Verification

- All 463 existing unit tests pass.
- The fix is surgical: only the agent discovery path is affected; skill resolution (`skills.ts`) remains unchanged.

Fixes #257

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -576,6 +576,14 @@ function loadAgentsFromDir(dir: string, source: AgentSource): AgentConfig[] {
 			continue;
 		}
 
+		// Skip skill files that are masquerading as agent definitions.
+		// Skill files typically contain a `metadata` or `version` frontmatter
+		// key that agent definitions do not use. If either is present, treat
+		// the file as a skill and skip agent registration.
+		if ("metadata" in frontmatter || "version" in frontmatter) {
+			continue;
+		}
+
 		const localName = frontmatter.name;
 		const parsedPackage = parsePackageName(frontmatter.package, `Agent '${localName}' package`);
 		if (parsedPackage.error) continue;


nicobailon:
Thanks for sending this. I’m going to close it rather than merge it because the recent package-discovery change covered the overlapping part, and this metadata/version check would still miss plenty of SKILL.md files. The remaining skill-directory case needs a more direct fix.

PR #261, opened 2026-06-08
feat(skills): support nested directories and lightweight injection

### LINKED ISSUE #183: Skill injection inlines full skill bodies into system prompt — biases agents and bloats context
## Summary

`buildSkillInjection` (in `src/agents/skills.ts`) inlines the **full body** of every inherited skill into a subagent's system prompt. For agents with `inheritSkills: true` and a non-trivial skill set, this is thousands of tokens of unrelated documentation injected on every turn, and — worse — it biases the model toward whichever skill has the most assertive content, even when the system prompt explicitly tells it not to use that skill.

Claude Code's native skill mechanism, by contrast, only injects `name + description + location` and instructs the model to `Read` the file lazily when the task matches. This keeps the prompt small and lets explicit system-prompt instructions actually win.

## Current behavior

```ts
// src/agents/skills.ts
export function buildSkillInjection(skills: ResolvedSkill[]): string {
  if (skills.length === 0) return "";
  return skills
    .map((s) => `<skill name="${s.name}">\n${s.content}\n</skill>`)
    .join("\n\n");
}
```

Every inherited skill's full SKILL.md content is concatenated into the system prompt.

### 1. Strong bias toward inlined skill content

Consider an agent whose system prompt explicitly says:

> **IMPORTANT — issue tracking**: identifiers like `FOO-123` are tracked in tool **A**, NOT tool **B**. Tool B is never used here.

With `inheritSkills: true`, the full SKILL.md for tool B (e.g. several hundred lines of CLI reference) gets inlined into the system prompt. Despite the explicit instruction, the agent consistently reaches for tool B's CLI on its very first action when asked to look up `FOO-123`, because:

- The inlined tool-B reference uses the same `PROJECT-NUMBER` identifier format throughout
- Concrete tool examples beat abstract "don't do this" instructions
- This reproduces reliably with strong models (e.g. claude-sonnet-4-6) and is robust even after narrowing the offending skill's `description` field to be very restrictive — the model reads the description but still uses the inlined examples because the full reference is sitting in its system prompt

The general failure pattern: when two skills cover overlapping use cases (any two issue trackers, any two cloud CLIs, any two search providers), the one with the bulkier inlined body wins, regardless of system prompt instructions.

### 2. Token cost on every turn

A user with ~14 skills installed easily ships 5–10k+ tokens of skill content per subagent turn, the vast majority of it unread.

### 3. Context budget pressure

For long-running orchestrator agents, the inlined skills consume budget that should belong to actual delegation/coordination context.

## Proposed fix

Match Claude Code's pattern — emit references only, let the model `read` lazily:

```ts
export function buildSkillInjection(skills: ResolvedSkill[]): string {
  if (skills.length === 0) return "";
  const entries = skills.map((s) =>
    `  <skill>\n    <name>${s.name}</name>\n    <description>${s.description ?? ""}</description>\n    <location>${s.path}</location>\n  </skill>`
  ).join("\n");
  return `<available_skills>\nUse the read tool to load a skill's file when the task matches its description.\n${entries}\n</available_skills>`;
}
```

Agents already have a `read` tool available (or can be granted one) and can pull in the full content on demand. This:

- Removes the bias problem (no inlined examples to copy from)
- Cuts thousands of tokens per turn
- Lets explicit system-prompt instructions win over skill content
- Keeps the same set of skills "available" to the agent

### Optional: keep both as a flag

If lazy-loading would break existing behavior for some users, gate it behind frontmatter:

```yaml
skillInjection: lazy   # default future
skillInjection: inline # current behavior, opt-in
```

## Workaround today

Use the `skills:` frontmatter allowlist to limit which skills are inlined per agent:

```yaml
inheritSkills: true
skills:
  - skill-a
  - skill-b
```

This works but requires every agent author to know about and curate the allowlist, and the underlying token cost remains for whatever they do inherit.

## Repro

1. Create an agent with `inheritSkills: true`
2. Install two skills that handle overlapping use cases (e.g. any two issue trackers — Jira + Linear, two cloud CLIs, two search providers)
3. Write the agent's system prompt to explicitly mandate skill A and forbid skill B
4. Dispatch a task that's ambiguous between them (e.g. an issue-key pattern that could match either)
5. Observe the agent reaching for whichever skill has the bulkier inlined body, even when the system prompt forbids it

Happy to send a PR if you'd accept this direction.

### LINKED ISSUE #262: Skill discovery only descends one level — misses nested layouts that the host runtime resolves
## Summary

`pi-subagents`'s `collectFilesystemSkills` (in `src/agents/skills.ts`) only descends one level into a registered skill root. It finds `<root>/<name>/SKILL.md` but stops short of `<root>/<group>/<name>/SKILL.md`. The host runtime `pi-coding-agent` already supports the deeper layout via its own `loadSkillsFromDirInternal`, which means agents and subagents disagree about which skills exist when both look at the same skills directory.

## Current behavior

`collectFilesystemSkills` (excerpt):

```ts
for (const child of childEntries) {
  if (child.name.startsWith(".")) continue;
  const childPath = path.join(skillPath.path, child.name);
  if (child.isDirectory() || child.isSymbolicLink()) {
    // Only checks one level deep — does not recurse
    const nestedSkillPath = path.join(childPath, "SKILL.md");
    if (fs.existsSync(nestedSkillPath)) {
      pushEntry(child.name, nestedSkillPath, skillPath.source);
    }
    continue;
  }
  // ...
}
```

For comparison, `pi-coding-agent`'s `loadSkillsFromDirInternal` recurses into subdirectories until it finds a `SKILL.md`, treating that directory as the skill anchor. From its source:

> Discovery rules:
> - if a directory contains SKILL.md, treat it as a skill root and do not recurse further
> - otherwise, load direct .md children in the root
> - recurse into subdirectories to find SKILL.md

### Hierarchical skill organization is a natural pattern

When a user collects more than a handful of skills, grouping them by domain becomes desirable for readability and maintenance:

```
skills/
  shell/
    safe-bash/SKILL.md
    tmux/SKILL.md
  web/
    chrome-devtools/SKILL.md
  data/
    sqlite/SKILL.md
    parquet/SKILL.md
```

The host scanner finds all five skills. The subagent scanner finds none of them, because there's no `SKILL.md` directly under `shell/`, `web/`, or `data/`.

### Frontmatter `skills:` references silently fail in subagents

A user who configures an agent with `skills: safe-bash, tmux` will see those skills resolved when the host renders the available-skills listing, but `resolveSkillsWithFallback` returns them as missing when the same agent runs as a subagent. The failure is silent unless `pi-subagents` is the resolved skill (which has a special-cased error).

### The two scanners should agree

There's no reason for the host and subagent runtimes to produce different views of the same directory. Whatever a user can declare in `pi-coding-agent` they should be able to inherit or reference from `pi-subagents`.

## Proposed fix

Update `collectFilesystemSkills` to recurse into subdirectories the same way `pi-coding-agent`'s `loadSkillsFromDirInternal` does:

- If a directory contains `SKILL.md`, treat it as the skill anchor and stop (do not recurse further into it).
- Otherwise recurse into its subdirectories looking for `SKILL.md`.
- Skip hidden directories and `node_modules` (matches the host scanner).

Layouts that previously resolved must keep resolving:

- `<root>/SKILL.md`
- `<root>/<name>.md`
- `<root>/<name>/SKILL.md`

Layouts that should newly resolve:

- `<root>/<group>/<name>/SKILL.md`
- arbitrary depth as long as no ancestor has its own `SKILL.md`

## Repro

1. Create a project skill at `.pi/skills/group/my-skill/SKILL.md`
2. Run `discoverAvailableSkills(cwd)` from `pi-subagents` — `my-skill` is missing
3. The same skill is visible to the host scanner via `pi-coding-agent`'s `loadSkillsFromDir`

## Workaround today

Flatten the directory layout so every skill sits one level under the root. This works but discards organizational structure once the skill count grows.

## Note

I have a PR ready for this — see #261, which fixes both this and #183 in the same change set.

Weaxs:
Closes #183
Closes #262

## Why

Two gaps surface when subagents and the host `pi-coding-agent` share the same skill directory.

### 1. Nested skill discovery mismatch

`pi-coding-agent`'s `loadSkillsFromDirInternal` recursively walks subdirectories looking for `SKILL.md`, treating any directory that contains one as a skill root. This lets users group skills hierarchically:

```
skills/
  shell/
    safe-bash/SKILL.md
    tmux/SKILL.md
  web/
    chrome-devtools/SKILL.md
```

`pi-subagents`'s `collectFilesystemSkills` only descends one level — it finds `<root>/<name>/SKILL.md` but not `<root>/<group>/<name>/SKILL.md`. The result is that a subagent's frontmatter `skills: safe-bash` resolves in the host but fails in the subagent runtime, even when both agents are pointed at the same skills tree.

### 2. Heavyweight injection is the only option

When an agent declares `skills: a, b, c` in frontmatter, `buildSkillInjection` always wraps each skill's full `SKILL.md` body into the system prompt. For agents that bundle many skills, this can add 100KB+ to every cold start, even though most invocations only touch one or two skills.

`pi-coding-agent`'s `formatSkillsForPrompt` solves the same problem on the host side by emitting a compact `<available_skills>` listing of name + description + path, and pointing the model at the read tool. Subagents have no equivalent escape hatch — they either inject everything or know nothing about declared skills.

### 1. Recursive skill discovery (matches `pi-coding-agent`)

`collectFilesystemSkills` now recurses into subdirectories of each registered skill root, stopping at any directory that contains a `SKILL.md` (treating it as the skill anchor for that subtree). Hidden directories and `node_modules` are skipped to mirror the host scanner.

Layouts that previously resolved are unchanged:
- `<root>/SKILL.md` (single skill at the root)
- `<root>/<name>.md` and `<root>/<name>/SKILL.md`

New layouts now resolved:
- `<root>/<group>/<name>/SKILL.md`
- arbitrary depth as long as no ancestor has its own `SKILL.md`

### 2. Optional lightweight skill injection

A new agent frontmatter field `skillInjection: full | light` (default `full`, fully backwards-compatible) controls how `buildSkillInjection` shapes the system prompt.

Given an agent with `skills: safe-bash, tmux`, here is the actual text appended to the subagent's system prompt in each mode.

**`full`** (default, current behavior) — the entire SKILL.md body is inlined per skill:

~~~text
<skill name="safe-bash">
# safe-bash

Use this skill when the task requires running shell commands.

## Guidelines
- ...
- ...
</skill>

<skill name="tmux">
# tmux

When the task involves long-running processes, use tmux sessions.

## Commands
- `tmux new-session -d -s <name>`
- ...
</skill>
~~~

**`light`** — only name, description, and absolute path are emitted, plus instructions to load each file on demand via the read tool:

~~~text
The following skills are available for this agent. Use the read tool to load a skill's file when the task matches its description.

<available_skills>
  <skill>
    <name>safe-bash</name>
    <description>Run shell commands safely with audit trail.</description>
    <location>/abs/path/to/skills/safe-bash/SKILL.md</location>
  </skill>
  <skill>
    <name>tmux</name>
    <description>Manage tmux sessions for long-running tasks.</description>
    <location>/abs/path/to/skills/tmux/SKILL.md</location>
  </skill>
</available_skills>
~~~

The `light` shape mirrors `pi-coding-agent`'s `formatSkillsForPrompt` output, so a model that knows how to use the host's `<available_skills>` listing already knows how to use this one.

> Note on naming: #183 sketches the same direction with `skillInjection: lazy | inline`. This PR uses `full | light` to make the contrast read more naturally as a heaviness scale (`full` SKILL.md vs `light` listing) and to keep the default explicitly the existing behavior. Happy to rename to match the issue if maintainers prefer.

The new `buildLightSkillInjection` helper is exported alongside the existing `buildSkillInjection`. The field is plumbed through `BuiltinAgentOverrideBase`, `BuiltinAgentOverrideConfig`, `AgentConfig`, `KNOWN_FIELDS`, the agent serializer, agent management input parsing, and the `subagent` extension tool description.

### Example agent file

A user/project agent declaring many skills but opting into the lightweight listing — only the four "core" skills are injected in full; the rest stay as references the agent loads on demand:

~~~yaml
---
name: shell-ops
description: Long-running shell and tmux operations specialist
tools: read, grep, find, ls, bash
systemPromptMode: replace
inheritProjectContext: true
inheritSkills: false
skillInjection: light
skills: safe-bash, tmux, chrome-devtools
---

You are a shell operations specialist.

Use the read tool to load any skill listed in your `<available_skills>` block when the task matches its description. Pull only what you need.
~~~

Without `skillInjection: light`, the same `skills:` declaration would inline three full SKILL.md bodies into the system prompt at every cold start. With `light`, only the listing block (name + description + location) is appended, and the agent loads each SKILL.md via the read tool when actually needed.

## Tests

`test/unit/skills-extensions.test.ts` adds 9 cases covering:

- Two-level nested discovery (`<root>/<group>/<name>/SKILL.md`)
- Recursion stops at the first `SKILL.md` (no nested SKILL.md leakage)
- `node_modules` and dotfile directories are excluded
- Deeply-nested skills resolve via `resolveSkills`
- `buildLightSkillInjection` emits name/description/location, never the full body
- Missing description gracefully omitted in light mode
- Empty skill list returns empty string
- XML-special characters escaped in name/description/location
- `buildSkillInjection` (default) still emits full body — regression guard

Full test suite: **512 / 512 pass** (`npm run test:unit`).

## Backwards compatibility

- Existing layouts (`<root>/<name>/SKILL.md` etc.) keep working unchanged.
- `skillInjection` defaults to `full`, so agents without the field behave exactly as before.
- Serializer only writes `skillInjection` when value is non-default (`light`), keeping existing agent files byte-identical on round-trip.


nicobailon:
Thanks for the PR. I would not merge this as-is, but I think it can get there with a smaller revision.

The scanner still walks inside a configured skill path after adding that directory's own `SKILL.md`. That means a settings or package skill path that points directly at a skill root can leak nested helper skills, which conflicts with the “stop at the first `SKILL.md`” rule this is trying to match.

For `skillInjection: light`, please also handle agents with an explicit tool allowlist that omits `read`. Right now the prompt tells the child to load the skill file, but a `tools: bash` style agent would not have the tool needed to do that. A validation error, warning, or automatic `read` inclusion would make that behavior clear.

Please add focused coverage for those cases, and update the user docs for the new frontmatter/override field.

nicobailon:
One more small thing for when you revise this: please add an Unreleased changelog entry for the user-visible skill discovery/light injection change, with credit to Weaxs (@Weaxs).

nicobailon:
I'm closing this for now while we clean up the PR queue. The nested-skill work still looks worth revisiting, but this branch needs a current-main revision before it can merge. Please feel free to open a fresh revised PR later if you want to pick it back up.

PR #270, opened 2026-06-12
fix: preserve nested fanout subagent tool history

jamesjwood:
## Summary
- avoid registering the main subagent extension in child mode; fanout children get the child-safe tool from the explicit fanout-child extension
- preserve live nested subagent tool calls/results inside authorized fanout children so they can observe list/delegate outcomes
- add regression coverage for duplicate registration and fanout history preservation

## Tests
- npm test
- npm run test:integration


nicobailon:
Thanks for the PR. I verified the nested fanout history issue is real, but this branch also pulls in the broader 0.27/0.28 acceptance-finalization work plus unrelated nested activity timestamp UI, so it is too wide to merge as this bugfix.

I split the narrow history preservation fix into #291 and left the separate fanout child registration conflict to #281.

PR #271, opened 2026-06-13
Fix legacy skill files being discovered as subagents

tacoman667:
## Summary

Fixes an issue where skill files under legacy `.agents/skills/**` and `~/.agents/skills/**` could be misrepresented as subagents.

Legacy agent discovery recursively scans `.agents/**/*.md` and `~/.agents/**/*.md`. Because legacy skill files also live under those roots, a `SKILL.md` file with normal skill frontmatter (`name` and `description`) could be parsed as an agent. Canonical paths do not have this overlap because `.pi/agents/**` and `.pi/skills/**` are separate directories.

## Solution

Skip the top-level `skills/` directory when scanning legacy `.agents` roots for agents. This keeps normal legacy agent discovery working while matching canonical behavior where skills are not discoverable as subagents.

Added unit coverage for both:
- project legacy skills under `.agents/skills/**`
- user legacy skills under `~/.agents/skills/**`

## Validation

- `node --experimental-strip-types --test test/unit/agent-frontmatter.test.ts` passes
- `npm run test:unit` was run; unrelated environment-sensitive tests failed due local Node flag support and local `pi-intercom` availability expectations

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -552,7 +552,11 @@ export function removeBuiltinAgentOverride(cwd: string, name: string, scope: "us
 	return filePath;
 }
 
-function listFilesRecursive(dir: string, predicate: (fileName: string) => boolean): string[] {
+function listFilesRecursive(
+	dir: string,
+	predicate: (fileName: string) => boolean,
+	options: { skipDirs?: Set<string> } = {},
+): string[] {
 	const files: string[] = [];
 	if (!fs.existsSync(dir)) return files;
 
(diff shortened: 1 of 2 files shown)


tacoman667:
I am in a couple of projects that have multiple coding agents contributing to it. 

I put all of our project skills and subagents into the `.agents/` folder in the project root for maximum compatibility. Because of this, `pi-subagents` keeps inconsistently pulling in skills from `.agents/skills/**/*` as subagents and messes up vital operations like code reviews, where we use a `code-review` skill for policy and guidelines, and a `code-reviewer` agent as a thin wrapper around it.

This PR is a proposal to correct what looks to be a potential bug in `pi-subagents` legacy discovery process.

Thanks for any feedback you may have!

nicobailon:
This fixes the legacy discovery bug. On main, a plain .agents/skills/.../SKILL.md can be picked up as a subagent; this branch keeps that skill file out while still loading an ordinary legacy agent from .agents.

I would take this over the older overlapping fixes: #204 is stale against main, and #258 would miss simple skill files that only declare name and description.

nicobailon:
Thanks for putting this together. The same legacy skill-discovery bug is now fixed on main by #303, with coverage for project/user legacy skills and package roots that expose broad agent directories.

I'm closing this as superseded rather than asking you to rework the conflicts against current main.

PR #272, opened 2026-06-13
fix: async parallel crash on Windows when concurrency > 1

1756141021:
## Summary

Async parallel runs crash reliably when `concurrency > 1` on Windows. The detached runner process exits silently before writing a result file, and stale-run reconciliation marks all children as failed.

- Reproduces with trivial delegate tasks ("Say A" / "Say B")
- Working combinations: async single, foreground parallel, async parallel `concurrency: 1`
- Broken combination: async parallel `concurrency > 1`

## Root Cause

Two unprotected call paths where `writeStatusPayload()` throws inside event callbacks become uncaught exceptions that kill the runner process.

**Path 1 — setInterval activity timer:**

```
setInterval callback → updateRunnerActivityState() → writeStatusPayload()
  → writeAtomicJson() → fs.renameSync() THROWS → uncaught exception → crash
```

**Path 2 — child stdout data handler:**

```
child.stdout.on("data") → processStdoutLine() → onChildEvent()
  → updateStepFromChildEvent() → writeStatusPayload()
  → writeAtomicJson() → fs.renameSync() THROWS → uncaught exception → crash
```

`writeAtomicJson` does `fs.renameSync(tempPath, statusPath)`. On Windows, `MoveFileExW` throws `EPERM`/`SHARING_VIOLATION` if the target file is held by another process (parent Pi polling `status.json`, Windows Defender, file indexer). With two concurrent child streams, `writeStatusPayload()` fires in rapid bursts — higher write frequency = higher collision probability.

Foreground parallel works because it uses push-based `onUpdate` callbacks, not file writes.

The runner is spawned with `stdio: "ignore"` and the entry point `.catch()` calls `process.exit(1)` without writing a result file — so all crash diagnostics are invisible.

### 1. `src/shared/atomic-json.ts` — retry transient Windows rename errors

Added `renameWithRetry()` that retries `fs.renameSync` up to 3 times on `EPERM`/`EBUSY`/`EACCES` with exponential backoff (50ms/100ms/200ms spin-wait). Non-retryable errors and exhausted retries re-throw.

### 2. `src/runs/background/subagent-runner.ts` — try/catch on unprotected callers

Wrapped the two unprotected `writeStatusPayload()` call sites (setInterval callback and `updateStepFromChildEvent`) with `try/catch`. These writes are purely for status display — swallowing the error is strictly better than crashing the runner.

### 3. `src/runs/background/subagent-runner.ts` — crash safety handlers

Added `uncaughtException` and `unhandledRejection` handlers that write a failure result file before exit. Future crashes produce a diagnosable result instead of relying on stale-run reconciliation.

### 4. `src/runs/background/subagent-runner.ts` — parallel interrupt fix

Changed `activeChildInterrupt` from a single shared variable to a `Map<number, () => void>` keyed by flat index. Previously, in parallel mode, the last task to register overwrote all others' interrupt handlers. Now each task registers independently and interrupt signals reach all running children.

## Verification

Tested on Windows 10 with pi-subagents v0.28.0. All 4 scenarios passed:

| Scenario | Result |
|---|---|
| Async parallel, `concurrency: 2` (previously broken) | **Pass** |
| Async single delegate | **Pass** |
| Foreground parallel, `concurrency: 2` | **Pass** |
| Async parallel, `concurrency: 1` | **Pass** |

The previously broken case — async parallel with `concurrency > 1` — completed successfully and returned both child outputs. The runner did not crash, did not fall into stale-run reconciliation, and produced normal async completion output.

## Test plan

- [x] Async parallel with `concurrency: 2` completes with correct outputs
- [x] Async single delegate still works
- [x] Foreground parallel still works
- [x] Async parallel with `concurrency: 1` still works
- [ ] Verify on Linux/macOS (rename collision is Windows-specific, but retry is safe no-op on other platforms)

### DIFF
--- a/src/runs/background/subagent-runner.ts
+++ b/src/runs/background/subagent-runner.ts
@@ -965,7 +965,7 @@ async function runSubagent(config: SubagentRunConfig): Promise<void> {
 	const eventsPath = path.join(asyncDir, "events.jsonl");
 	const logPath = path.join(asyncDir, `subagent-log-${id}.md`);
 	const controlConfig = config.controlConfig ?? DEFAULT_CONTROL_CONFIG;
-	let activeChildInterrupt: (() => void) | undefined;
+	const activeChildInterrupts = new Map<number, () => void>();
 	let interrupted = false;
 	let currentActivityState: ActivityState | undefined;
 	let activityTimer: NodeJS.Timeout | undefined;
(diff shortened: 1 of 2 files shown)


BingCoke:
This problem has troubled me for a long time, and network fluctuations can easily lead to the failure of subagents.

tssc67:
Thanks! This fix my issue.

nicobailon:
Thanks for digging into this and confirming the Windows async-parallel failure. I opened #292 as a narrower maintainer fix that keeps the change focused on the shared atomic JSON rename path and includes portable regression coverage.\n\nI’m closing this one so we have one approval path for the fix, but the reproduction details here were useful.

PR #273, opened 2026-06-14
fix: launch async runner with node

tetsuya-dev-jp:
## Summary

Fix async subagent runner startup when the parent Pi process is not running from a Node executable.

In some Pi runtimes, `process.execPath` points to the Pi executable rather than `node`. The async launcher used `process.execPath` to start the detached TypeScript runner via jiti, which could start `pi jiti ...` instead of `node jiti ...`. That process exits before writing an async result, and stale-run reconciliation marks the run failed.

This change uses `process.execPath` only when it looks like a Node executable. Otherwise it falls back to `node` / `node.exe` from PATH.

## Tests

- Added integration coverage for non-Node `process.execPath`.
- `npm run test:integration -- --test-name-pattern='process.execPath|cannot spawn' test/integration/async-execution.test.ts` passed.


nicobailon:
This is a real async-runner startup fix. The runner needs to be launched with Node even when the parent Pi runtime exposes a non-Node process path; otherwise jiti can be handed to the wrong executable.

I do not see this overlapping the Windows async crash reports. Those fail later while writing runner state; this fixes the process used to start the runner in the first place.

Issue #242, opened 2026-06-01
Add timeout for foreground subagent runs


ThePhoenixCoding:
# Feature request: timeout for foreground subagent runs

Please add a `timeoutMs` / `maxRuntimeMs` option for foreground `subagent` runs.

Use case: some workflows must stay foreground/blocking so normal Pi UX still works: `Esc` can abort the active tool call, user prompts are queued, and the parent receives the result in the same turn. Using `async: true` to implement timeouts changes that UX.

Expected behavior:

- Works for foreground single, parallel, and chain runs.
- When the timeout expires, soft-interrupt still-running child runs.
- Return partial results for completed children.
- Mark timed-out children clearly, e.g. `timedOut: true` plus a stable timeout message.
- Preserve existing manual `Esc` interrupt behavior.

Example:

```ts
subagent({
  tasks: [
    { agent: "reviewer", task: "Review the current diff." },
    { agent: "reviewer", task: "Review tests and edge cases." }
  ],
  concurrency: 2,
  context: "fresh",
  timeoutMs: 600000
})
```

nicobailon:
This shipped in v0.28.0. Foreground single, parallel, and chain runs support timeoutMs and maxRuntimeMs now, with partial results preserved and timed-out children marked clearly. Closing it out.

ThePhoenixCoding:
`timeoutMs` and `maxRuntimeMs` are nowhere to be found in the source code, only in the changelog 🤔

PR #280, opened 2026-06-16
[Feature Request] Configurable minimum timeout

orbitingflea:
## Summary

This PR adds `minForegroundTimeoutMs` to the subagent extension config.

For foreground runs, `timeoutMs` and `maxRuntimeMs` now have an effective lower bound. The lower bound defaults to `300000` ms (5 min). Users can change it in `~/.pi/agent/extensions/subagent/config.json`.

If a foreground call asks for a shorter timeout, `pi-subagents` raises it to the configured minimum. If it asks for a longer timeout, `pi-subagents` uses that longer value. Async/background runs are unchanged: they still reject `timeoutMs` and `maxRuntimeMs`.

## Motivation

The code that launches subagents can set one foreground timeout for many child runs. That caller often does not know which model each child will use, or how slow that model will be. A timeout like two minutes can look safe to the caller, but it is often too short for state-of-the-art reasoning models.

With a slow model, this causes predictable failures. Review, validation, and acceptance-finalization runs can time out before they finish. The child is not stuck; the caller chose a timeout without enough model-latency context.

A config-level minimum lets the installation owner set the shortest allowed foreground timeout. Per-call timeouts still work above that floor, but callers cannot set values that are too short for the models used in that installation.

## Changes

- Add `minForegroundTimeoutMs?: number` to extension config.
- Add shared helpers:
  - `resolveMinForegroundTimeoutMs(config)`
  - `normalizeForegroundTimeoutMs(timeoutMs, minTimeoutMs)`
- Use the configured floor when resolving foreground timeouts.
- Raise short timeout arguments before tool schema validation, while preserving the existing error for unequal `timeoutMs` / `maxRuntimeMs` aliases.
- Keep the public schema minimum at `1`, so runtime config controls the timeout floor.
- Show the effective minimum in `subagent({ action: "doctor" })`.
- Document the config option in the README, skill guidance, tool description, and changelog.
- Add unit coverage for default, custom, invalid, and schema behavior.

## Validation

```bash
node --experimental-strip-types --test test/unit/foreground-timeout-config.test.ts
bun build src/extension/index.ts --target=node --outdir /tmp/pi-subagents-pr-check \
  --external @earendil-works/pi-coding-agent \
  --external @earendil-works/pi-agent-core \
  --external @earendil-works/pi-ai
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Changed
+- Added `minForegroundTimeoutMs` config for the foreground `timeoutMs`/`maxRuntimeMs` floor, defaulting to 300000 ms (5 min), and raise shorter caller-provided values instead of letting foreground subagent runs time out too early.
+
 ## [0.28.0] - 2026-06-03
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Use `~/.pi/agent/settings.json` for a user override or `.pi/settings.json` for a
 
 ## Where running subagents show up
 
-Foreground runs stream progress in the conversation while they run. Use `timeoutMs` or its alias `maxRuntimeMs` when a foreground run must return within a wall-clock budget. When the timeout expires, running children are soft-interrupted, completed children stay in the result, and timed-out children return `timedOut: true` with a stable timeout message.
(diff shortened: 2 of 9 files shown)


nicobailon:
The useful part here is foreground timeout support, which still tracks #242. I would split that out as the mergeable change.

I do not think this PR should merge as-is because it also changes omitted foreground timeouts from unbounded to 5 minutes. That is a behavior change, not just a safety floor. If we want a timeout floor, it should apply only when a caller provides `timeoutMs` or `maxRuntimeMs`; omitted timeouts should keep the existing behavior.

nicobailon:
Closing this for now because the useful piece should come back as a narrower #242 fix. Foreground timeout support makes sense, but this branch also changes omitted foreground timeouts from unbounded to 5 minutes, and I do not want to take that behavior change as part of the same PR.

PR #205, opened 2026-05-22
fix: prevent duplicate 'subagent' tool registration in child processes

celsolom:
## Problema

Quando um subagente child tem `tools: subagent` no frontmatter (ex: agentes orquestradores que fazem fan-out), a extensão `pi-subagents` é carregada de duas formas:

1. **Via `package.json` manifest** → `./src/extension/index.ts`
2. **Via `--extension` flag** → `./src/extension/fanout-child.ts`

Ambos os caminhos chamavam `registerFanoutChildSubagentExtension(pi)`, que registrava uma tool chamada `"subagent"` nas duas APIs de extensão. O `detectExtensionConflicts()` do resource-loader do pi detectava duas extensões com paths diferentes registrando a mesma tool e rejeitava o carregamento com:

```
Error: Tool "subagent" conflicts with .../fanout-child.ts
```

## Causa raiz

Em `src/runs/shared/pi-args.ts`, quando `fanoutAuthorized = true` (porque `declaredBuiltinTools.includes("subagent")`):
- `FANOUT_CHILD_EXTENSION_PATH` é adicionado ao array `runtimeExtensions`
- O child recebe `--extension /path/to/fanout-child.ts` na linha de comando
- O child também carrega `index.ts` via descoberta normal de pacotes (`package.json`)

Em `index.ts`:
```ts
if (process.env[SUBAGENT_CHILD_ENV] === "1") {
    if (process.env[SUBAGENT_FANOUT_CHILD_ENV] === "1") registerFanoutChildSubagentExtension(pi);
    return;
}
```

Quando `SUBAGENT_FANOUT_CHILD_ENV=1`, ele chamava `registerFanoutChildSubagentExtension(pi)`, que registrava `"subagent"` no API do `index.ts`. Simultaneamente, o `fanout-child.ts` carregado como extensão independente também registrava `"subagent"` no seu próprio API. Conflito.

## Correção

No `index.ts`, quando `SUBAGENT_CHILD_ENV=1`, retornar cedo sem chamar `registerFanoutChildSubagentExtension`. O `fanout-child.ts` já é carregado como extensão independente via `--extension` e faz o registro sozinho.

## Testes

Testado com `masterpi.orchestrator` (agente com `tools: subagent`). Antes da correção: falha com conflito. Depois: executou 9 turns com sucesso, listando ~63 agentes sem erros.

PR #281, opened 2026-06-16
fix: use process-global boolean for fanout-child dedup instead of per-pi WeakSet

### LINKED ISSUE #279: Fanout-authorized child fails to start: Tool "subagent" conflicts with fanout-child.ts (double registerTool)
## Summary

In `pi-subagents@0.28.0`, any **fanout-authorized child** (i.e. a subagent whose `tools:` list includes `subagent`) fails to start with a tool-registration conflict:

```
Error: Failed to load extension ".../pi-subagents/src/extension/index.ts":
Tool "subagent" conflicts with .../pi-subagents/src/extension/fanout-child.ts
```

The root cause is a double call to `pi.registerTool({ name: "subagent", ... })` inside the child process. `registerFanoutChildSubagentExtension(pi)` is invoked **twice** — once from the package's `index.ts` (via its child-mode early-return) and once from the standalone `fanout-child.ts` entry — and the dedup guard inside that function keys on the `pi` instance via a `WeakSet`, but each extension receives a **distinct** `ExtensionAPI` instance, so the guard never matches.

## Environment

- `pi-subagents`: 0.28.0 (latest at time of writing)
- `@earendil-works/pi-coding-agent`: 0.79.4
- Node: v24.16.0
- macOS 26.5.1

## Reproduction

Minimal repro — launch `pi` directly in fanout-child mode with the standalone extension (no model call needed; the conflict is at extension-load time):

```bash
PI_SUBAGENT_CHILD=1 PI_SUBAGENT_FANOUT_CHILD=1 \
  pi --no-session --no-skills --print \
  --extension node_modules/pi-subagents/src/extension/fanout-child.ts \
  "reply with the single word ok"
```

Output:

```
Error: Failed to load extension ".../pi-subagents/src/extension/index.ts": Tool "subagent" conflicts with .../pi-subagents/src/extension/fanout-child.ts
```

### Real-world trigger

Any agent definition with `subagent` in its `tools:` list and no explicit `extensions:` field, e.g.:

```yaml
---
name: discovery-orchestrator
tools: read, write, edit, bash, subagent
context: fresh
maxSubagentDepth: 2
---
```

The orchestrator then launches a child via `subagent({ agent: "discovery-worker", ... })`, and that child process dies on startup with the conflict above.

## Root cause

1. Because the child's declared builtin tools include `subagent`, `buildPiArgs` sets `fanoutAuthorized = true` (`src/runs/shared/pi-args.ts:101`), which both **adds the `fanout-child.ts` extension** to the child's `--extension` list (`pi-args.ts:118-119`) **and** sets `PI_SUBAGENT_CHILD=1` / `PI_SUBAGENT_FANOUT_CHILD=1` (`pi-args.ts:158`).
2. When the child has **no explicit `extensions:` field**, `input.extensions === undefined`, so `buildPiArgs` does **not** emit `--no-extensions` (`pi-args.ts:121`). The child therefore **also inherits the package's `index.ts` extension** (declared under `pi.extensions` in `package.json`).
3. In the child, `index.ts` hits its child-mode early-return at `src/extension/index.ts:211-213`, which calls `registerFanoutChildSubagentExtension(pi)`. Separately, the standalone `fanout-child.ts` entry calls the **same** function (`src/extension/fanout-child.ts:128`).
4. The dedup inside `registerFanoutChildSubagentExtension` keys on the `pi` instance:

   ```ts
   const registeredApis = globalStore[registeredKey] instanceof WeakSet
       ? globalStore[registeredKey] as WeakSet<ExtensionAPI>
       : new WeakSet<ExtensionAPI>();
   globalStore[registeredKey] = registeredApis;
   if (registeredApis.has(pi)) return;   // <-- never true
   registeredApis.add(pi);
   ```

   But each extension receives a **distinct** `ExtensionAPI` instance, so `registeredApis.has(pi)` is always `false` → both calls proceed → `pi.registerTool("subagent")` runs twice (`src/extension/fanout-child.ts:172`) → pi rejects the duplicate with the "Tool conflicts" error.

The same function is defined once and called from two places, so the bug is in the dedup key, not the call sites:

```
src/extension/index.ts:37   import registerFanoutChildSubagentExtension from "./fanout-child.ts";
src/extension/index.ts:212      if (process.env[SUBAGENT_FANOUT_CHILD_ENV] === "1") registerFanoutChildSubagentExtension(pi);
src/extension/fanout-child.ts:128  export default function registerFanoutChildSubagentExtension(pi: ExtensionAPI)
```

## Suggested fix

Key the dedup on a process-global boolean instead of a per-`pi` `WeakSet`. Whichever extension loads first wins; the other no-ops (the executor/state the second call would build is identical to what the first already set up).

```diff
 export default function registerFanoutChildSubagentExtension(pi: ExtensionAPI): void {
 	if (process.env[SUBAGENT_CHILD_ENV] !== "1" || process.env[SUBAGENT_FANOUT_CHILD_ENV] !== "1") return;

-	const globalStore = globalThis as Record<string, unknown>;
-	const registeredKey = "__piSubagentFanoutChildRegisteredApis";
-	const registeredApis = globalStore[registeredKey] instanceof WeakSet
-		? globalStore[registeredKey] as WeakSet<ExtensionAPI>
-		: new WeakSet<ExtensionAPI>();
-	globalStore[registeredKey] = registeredApis;
-	if (registeredApis.has(pi)) return;
-	registeredApis.add(pi);
+	const globalStore = globalThis as Record<string, unknown>;
+	const registeredKey = "__piSubagentFanoutChildRegistered";
+	if (globalStore[registeredKey] === true) return;
+	globalStore[registeredKey] = true;
```

## Verification

- Reproduced the exact error with the original code using the minimal repro above.
- Applied the fix → the fanout child loads and responds cleanly (`ok`).
- `node --experimental-strip-types --check src/extension/fanout-child.ts` passes.

Happy to open a PR if this analysis looks right.

elecnix:
### Problem
Any **fanout-authorized child** (subagent with `subagent` in its `tools:` list) fails to start with a tool-registration conflict:

```
Error: Failed to load extension ".../pi-subagents/src/extension/index.ts":
Tool "subagent" conflicts with .../pi-subagents/src/extension/fanout-child.ts
```

### Root Cause
`registerFanoutChildSubagentExtension(pi)` is called twice:
1. From `index.ts` (child-mode early-return)
2. From standalone `fanout-child.ts`

The dedup guard used a `WeakSet<ExtensionAPI>` keyed on the `pi` instance, but each extension receives a **distinct** `ExtensionAPI` instance, so the guard never matches → `pi.registerTool("subagent")` runs twice → pi rejects the duplicate.

### Fix
Key the dedup on a **process-global boolean** instead of a per-`pi` `WeakSet`. Whichever extension loads first wins; the other no-ops (the executor/state the second call would build is identical to what the first already set up).

### Verification
- `node --experimental-strip-types --check src/extension/fanout-child.ts` ✓
- The minimal repro from the issue no longer produces the conflict

### DIFF
--- a/src/extension/fanout-child.ts
+++ b/src/extension/fanout-child.ts
@@ -129,13 +129,9 @@ export default function registerFanoutChildSubagentExtension(pi: ExtensionAPI): 
 	if (process.env[SUBAGENT_CHILD_ENV] !== "1" || process.env[SUBAGENT_FANOUT_CHILD_ENV] !== "1") return;
 
 	const globalStore = globalThis as Record<string, unknown>;
-	const registeredKey = "__piSubagentFanoutChildRegisteredApis";
-	const registeredApis = globalStore[registeredKey] instanceof WeakSet
-		? globalStore[registeredKey] as WeakSet<ExtensionAPI>
-		: new WeakSet<ExtensionAPI>();
-	globalStore[registeredKey] = registeredApis;
-	if (registeredApis.has(pi)) return;
-	registeredApis.add(pi);
+	const registeredKey = "__piSubagentFanoutChildRegistered";
+	if (globalStore[registeredKey] === true) return;
+	globalStore[registeredKey] = true;
 
 	const config = loadConfig();
 	const state = createChildSafeState();
--- a/src/extension/fanout-child.ts
+++ b/src/extension/fanout-child.ts
@@ -131,7 +131,6 @@ export default function registerFanoutChildSubagentExtension(pi: ExtensionAPI): 
 	const globalStore = globalThis as Record<string, unknown>;
 	const registeredKey = "__piSubagentFanoutChildRegistered";
 	if (globalStore[registeredKey] === true) return;
-	globalStore[registeredKey] = true;
 
 	const config = loadConfig();
 	const state = createChildSafeState();
@@ -162,5 +161,6 @@ export default function registerFanoutChildSubagentExtension(pi: ExtensionAPI): 
 	};
 
 	pi.registerTool(tool);
+	globalStore[registeredKey] = true;
 	startNestedControlInboxListener(pi, state);
 }
--- a/test/unit/index-child-registration.test.ts
+++ b/test/unit/index-child-registration.test.ts
@@ -185,6 +185,47 @@ describe("subagent extension child mode", () => {
 		);
 	});
 
+	it("deduplicates the child-safe subagent tool when both child extension entries load", () => {
+		const script = String.raw`
+			import registerSubagentExtension from "./src/extension/index.ts";
+			import registerFanoutChildSubagentExtension from "./src/extension/fanout-child.ts";
+			import { SUBAGENT_CHILD_ENV, SUBAGENT_FANOUT_CHILD_ENV } from "./src/runs/shared/pi-args.ts";
+			process.env[SUBAGENT_CHILD_ENV] = "1";
+			process.env[SUBAGENT_FANOUT_CHILD_ENV] = "1";
+			const registeredNames = new Set();
+			const calls = [];
+			function makePi(source) {
+				return {
+					events: { on() { return () => {}; }, emit() {} },
+					registerTool(tool) {
+						if (registeredNames.has(tool.name)) throw new Error("Tool " + JSON.stringify(tool.name) + " conflicts with " + source);
+						registeredNames.add(tool.name);
+						calls.push(source + ":" + tool.name);
+					},
+					getSessionName() { return undefined; },
+				};
+			}
+			registerSubagentExtension(makePi("index.ts"));
+			registerFanoutChildSubagentExtension(makePi("fanout-child.ts"));
+			if (calls.length !== 1 || calls[0] !== "index.ts:subagent") {
+				throw new Error("expected exactly one child-safe subagent registration, got " + JSON.stringify(calls));
+			}
+		`;
+
+		execFileSync(
+			process.execPath,
+			[
+				"--experimental-transform-types",
+				"--import",
+				"./test/support/register-loader.mjs",
+				"--input-type=module",
+				"--eval",
+				script,
+			],
+			{ cwd: projectRoot, stdio: "pipe" },
+		);
+	});
+
 	it("lets fanout children call read-only list but blocks mutating management actions", () => {
 		const script = String.raw`
 			import registerFanoutChildSubagentExtension from "./src/extension/fanout-child.ts";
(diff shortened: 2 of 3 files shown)


elecnix:
## Test results

Installed from `git:github.com/elecnix/pi-subagents@fix/fanout-tool-conflict` and verified the fix against the exact reproduction scenario from the issue:

### Tests run

| # | Test | Subagent type | Model | Result |
|---|------|--------------|-------|--------|
| 1 | Single worker | `worker` | `deepseek/deepseek-v4-flash` | ✅ launched without conflict |
| 2 | Chain (researcher → planner → worker) | chain / fanout | `deepseek/deepseek-v4-flash` | ✅ steps 1-2 succeeded; step 3 rejected non-implementation task (expected) |
| 3 | Parallel fan-out (2 tasks) | `researcher` + `scout` | `deepseek/deepseek-v4-flash` | ✅ both launched and produced results concurrently |
| 4 | Parallel delegates (2 tasks) | `delegate` × 2 | `deepseek/deepseek-v4-flash` | ✅ both launched and produced results concurrently |

### What this proves

Every test spawns exactly the fanout-authorized child scenario from the issue — a subagent that inherits the `subagent` tool → child process gets `PI_SUBAGENT_FANOUT_CHILD=1` + the `fanout-child.ts` extension. Prior to the fix, this double-load would crash with:

```
Error: Tool "subagent" conflicts with .../pi-subagents/src/extension/fanout-child.ts
```

All 4 tests passed without any tool-registration error. The process-global boolean dedup correctly prevents the duplicate `pi.registerTool("subagent")` call.

nicobailon:
I reproduced the duplicate child registration path with two distinct ExtensionAPI shims sharing one tool registry, then pushed a regression test for that case plus a small guard tweak so the process-global marker is only set after registerTool succeeds.

Verification: `node --experimental-strip-types --test test/unit/index-child-registration.test.ts`, `node --experimental-strip-types --test test/unit/nested-control.test.ts`, `npm run test:unit`, and `node --experimental-strip-types --check src/extension/fanout-child.ts`.

nicobailon:
Thanks for putting this together. I checked the branch against the current tests, and the underlying bug is real, but this approach is not green: `node --experimental-strip-types --test test/unit/nested-control.test.ts` fails because the process-global flag prevents later fresh fanout-child registrations/listeners in the same process; `npm run test:unit` also fails with the same two nested-control failures.\n\nThis also overlaps the older fix in #205 for the same duplicate tool registration problem, so I’m closing this one in favor of continuing there.

PR #283, opened 2026-06-17
fix: subagents inherit parent session model instead of global settings.json fallback (#266)

### LINKED ISSUE #266: [CRITICAL] Cross-session model contamination: subagents inherit wrong session's model via global settings.json fallback
## Summary

When multiple PI sessions are open with different models (e.g. Session A: `claude-sonnet`, Session B: `gpt-5.5`), subagents spawned by Session B incorrectly use Session A's model instead of Session B's. This silently corrupts model selection across all subagent workflows.

## Root Cause (Verified via Source Code Analysis)

The bug has two components working together:

### 1. `model: "inherit"` is not handled — `resolveModelCandidate` returns the string `"inherit"` unchanged

In `src/runs/shared/model-fallback.ts`:
```typescript
export function resolveModelCandidate(model, availableModels, preferredProvider) {
    if (!model) return undefined;
    if (model.includes("/")) return model;
    const matches = availableModels.filter((e) => e.id === model); // "inherit" → empty
    if (matches.length !== 1) return model; // ← returns "inherit" string as-is!
}
```

### 2. The child process receives `--model inherit` (invalid), falls back to global `settings.json`

In `src/runs/shared/pi-args.ts`, `buildPiArgs` emits `--model inherit` to the child process.

In `@earendil-works/pi-coding-agent/dist/core/model-resolver.js`, `findInitialModel` has:
```javascript
if (cliProvider && cliModel) { ... } // Requires BOTH — --provider is never set → SKIPPED entirely
// Falls through to:
if (defaultProvider && defaultModelId) {
    return { model: modelRegistry.find(defaultProvider, defaultModelId) };
    // Reads from ~/.pi/agent/settings.json — a SHARED GLOBAL FILE
}
```

### 3. `settings.json` is written by any session that changes model in the TUI

In `agent-session.js`, model cycling and selection (Ctrl+N/Ctrl+P, model picker) calls:
```javascript
this.settingsManager.setDefaultModelAndProvider(model.provider, model.id);
```

This writes `defaultProvider` and `defaultModel` to the **global** `~/.pi/agent/settings.json`.

### 4. The `delegate` agent has the same bug

`delegate` has `"model": false` in its config. When spawned, no `--model` arg is passed to the child at all, which also falls back to the global `settings.json` default. The description "inherits the parent model" is currently incorrect.

## Reproduction

```
1. Open Session A with model claude-sonnet
2. In Session A, switch models in the TUI (Ctrl+N, then back) — writes claude-sonnet to settings.json
3. Open Session B with model gpt-5.5 via --model flag (does NOT update settings.json)
4. In Session B, spawn:
     subagent({ agent: "delegate", task: "..." })
   OR:
     subagent({ agent: "worker", model: "inherit", task: "..." })
5. The subagent runs with claude-sonnet instead of gpt-5.5  ← BUG
```

## Impact

- **Every subagent workflow using `model: "inherit"` or agents with `model: false` (delegate, worker without explicit model, etc.) is broken** when multiple PI sessions are open
- Silent failure — no error shown, wrong model used
- Race condition: whichever session last changed model in TUI "poisons" all future child processes from other sessions
- The `delegate` agent, which documents "inherits the parent model", never actually inherits correctly

## Fix

In `src/runs/foreground/subagent-executor.ts`, both paths need to use `ctx.model` (the parent session's actual in-memory model) when no explicit model is provided or when `"inherit"` is requested:

### Parallel path:
```typescript
// BEFORE:
const modelOverrides = params.tasks.map((task, index) =>
    resolveModelCandidate(task.model ?? agentConfigs[index]?.model, availableModels, currentProvider),
);

// AFTER:
const modelOverrides = params.tasks.map((task, index) => {
    const rawModel = (task.model ?? agentConfigs[index]?.model) || undefined;
    // "inherit" or no model specified → use parent session's actual model to avoid
    // falling back to the global settings.json default (cross-session contamination).
    if ((rawModel === "inherit" || rawModel === undefined) && ctx.model) {
        return `${ctx.model.provider}/${ctx.model.id}`;
    }
    return resolveModelCandidate(rawModel, availableModels, currentProvider);
});
```

### Single-agent path:
```typescript
// BEFORE:
const modelOverride = resolveModelCandidate(
    (params.model as string | undefined) ?? a.model,
    availableModels,
    currentProvider,
);

// AFTER:
const rawModelSingle = ((params.model as string | undefined) ?? (a.model || undefined));
// "inherit" or no model specified → use parent session's actual model to avoid
// falling back to the global settings.json default (cross-session contamination).
const modelOverride = (rawModelSingle === "inherit" || rawModelSingle === undefined) && ctx.model
    ? `${ctx.model.provider}/${ctx.model.id}`
    : resolveModelCandidate(rawModelSingle, availableModels, currentProvider);
```

### Why this works

- `ctx.model` is the parent's **in-memory** runtime model (not read from settings.json)
- Passing `provider/id` (with slash) causes `resolveModelCandidate` to return it directly (`if (model.includes("/")) return model`)
- The child receives an explicit `--model provider/id` arg and resolves it correctly
- Sessions are now isolated: each session's children inherit only that session's model

### Note on chain path

The chain execution path may have a similar issue where per-step model resolution does not inherit from `ctx.model`. Worth a follow-up investigation in `executeAsyncChain`.

## Environment

- `pi-subagents` version: `0.28.0`
- Runtime: `@earendil-works/pi-coding-agent`

elecnix:
## Summary

Fixes #266. Subagents spawned with no explicit model, with `model: false` (e.g. the `delegate` agent), or with `model: "inherit"` were resolving to `undefined` / the literal `"inherit"` string. The child `pi` process then received no usable `--model` arg and fell back to the **shared global** `~/.pi/agent/settings.json` default — which any open session can overwrite by changing its model in the TUI. With multiple PI sessions open, a subagent silently ran on a *different session's* model.

## Root cause

`resolveModelCandidate("inherit", …)` finds no registry match and returns `"inherit"` verbatim; `undefined`/`false` model configs return `undefined`. Neither yields the parent session's actual model, so the child resolves its own model from the global settings file.

## Fix

New helper `resolveSubagentModelOverride` (`src/runs/shared/model-fallback.ts`): when no model is explicitly requested (`undefined`, `false`, empty, or the `"inherit"` sentinel), it pins the parent session's **in-memory** `provider/id` (`ctx.model`). An explicit model string still flows through `resolveModelCandidate`. Passing an explicit `provider/id` (which `resolveModelCandidate` returns unchanged) keeps each session's children isolated to that session's model.

Applied at all four subagent dispatch sites in `subagent-executor.ts`:
- async parallel (`runAsyncPath`)
- async single (`runAsyncPath`)
- sync parallel (`runParallelPath`)
- sync single (`runSinglePath`)

This also fixes the `delegate` agent, whose documented "inherits the parent model" behavior was previously incorrect.

### Follow-up (out of scope)

The chain execution path (`chain-execution.ts`) resolves per-step models separately and the async chain ctx only carries `provider` (not `id`); inheriting there needs the full parent model threaded through. Left as a documented follow-up, consistent with the issue's note.

## Test plan

**Unit** — `test/unit/model-fallback.test.ts`, 9 new cases covering: inherit on `undefined`/`"inherit"`/`false`/empty/whitespace, explicit `provider/id` and bare-id passthrough, `undefined` when no parent model is known, and a regression guard that the literal `"inherit"` is never emitted.

- `npm run test:all` → 512 unit + 365 integration, all passing.

**Live reproduction** — ran the real coding agent (`pi -e`) in tmux with parent model `deepseek/deepseek-v4-flash` while the global `settings.json` default was `deepseek-v4-pro`, then spawned the `delegate` subagent (a "hello world" task) and inspected the subagent session JSONL:

| Code | Parent session model | Subagent session model | Result |
|------|----------------------|------------------------|--------|
| `origin/main` (before) | `deepseek-v4-flash` | `deepseek-v4-pro` | ❌ contaminated by global default |
| this branch (after) | `deepseek-v4-flash` | `deepseek-v4-flash` | ✅ inherits parent session model |

### DIFF
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -13,7 +13,7 @@ import { handleManagementAction } from "../../agents/agent-management.ts";
 import { buildDoctorReport } from "../../extension/doctor.ts";
 import { clearPendingForegroundControlNotices } from "../../extension/control-notices.ts";
 import { runSync } from "./execution.ts";
-import { resolveModelCandidate } from "../shared/model-fallback.ts";
+import { resolveModelCandidate, resolveSubagentModelOverride } from "../shared/model-fallback.ts";
 import { aggregateParallelOutputs } from "../shared/parallel-utils.ts";
 import { recordRun } from "../shared/run-history.ts";
 import {
@@ -1092,7 +1092,7 @@ function runAsyncPath(data: ExecutionContextData, deps: ExecutorDeps): AgentTool
 	if (hasTasks && params.tasks) {
 		const agentConfigs = params.tasks.map((task) => agents.find((agent) => agent.name === task.agent));
 		const modelOverrides = params.tasks.map((task, index) =>
-			resolveModelCandidate(task.model ?? agentConfigs[index]?.model, availableModels, currentProvider),
+			resolveSubagentModelOverride(task.model ?? agentConfigs[index]?.model, ctx.model, availableModels, currentProvider),
 		);
 		const skillOverrides = params.tasks.map((task) => normalizeSkillInput(task.skill));
 		const parallelTasks = params.tasks.map((task, index) => ({
@@ -1179,7 +1179,7 @@ function runAsyncPath(data: ExecutionContextData, deps: ExecutorDeps): AgentTool
 		const normalizedSkills = normalizeSkillInput(params.skill);
 		const skills = normalizedSkills === false ? [] : normalizedSkills;
 		const maxSubagentDepth = resolveChildMaxSubagentDepth(currentMaxSubagentDepth, a.maxSubagentDepth);
-		const modelOverride = resolveModelCandidate((params.model as string | undefined) ?? a.model, availableModels, currentProvider);
+		const modelOverride = resolveSubagentModelOverride((params.model as string | undefined) ?? a.model, ctx.model, availableModels, currentProvider);
 		return executeAsyncSingle(id, {
 			agent: params.agent!,
 			task: params.context === "fork" ? wrapForkTask(params.task ?? "") : (params.task ?? ""),
@@ -1635,7 +1635,7 @@ async function runParallelPath(data: ExecutionContextData, deps: ExecutorDeps): 
 		...(task.model ? { model: task.model } : {}),
 	}));
 	const modelOverrides: (string | undefined)[] = tasks.map((_, i) =>
-		resolveModelCandidate(behaviorOverrides[i]?.model ?? agentConfigs[i]?.model, availableModels, currentProvider),
+		resolveSubagentModelOverride(behaviorOverrides[i]?.model ?? agentConfigs[i]?.model, ctx.model, availableModels, currentProvider),
 	);
 
 	if (params.clarify === true && ctx.hasUI) {
@@ -1918,8 +1918,9 @@ async function runSinglePath(data: ExecutionContextData, deps: ExecutorDeps): Pr
 	const currentProvider = ctx.model?.provider;
 	const availableModels: ModelInfo[] = ctx.modelRegistry.getAvailable().map(toModelInfo);
 	let task = params.task ?? "";
-	let modelOverride: string | undefined = resolveModelCandidate(
+	let modelOverride: string | undefined = resolveSubagentModelOverride(
 		(params.model as string | undefined) ?? agentConfig.model,
+		ctx.model,
 		availableModels,
 		currentProvider,
 	);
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for the PR. I verified the core issue and the single/parallel parts of this fix are good, but chain execution still had the same fallback path: chain steps with no model or `model: inherit` could still fall back to shared settings.json.

I opened #293 carrying this change forward with the foreground and async chain gap covered, plus regression tests. Keeping review focused there since I cannot safely push to this branch.

nicobailon:
Thanks for the solid starting point here. I opened #293 with this fix carried forward plus the missing foreground/async chain coverage, and that replacement is green on Ubuntu and Windows, so I’m closing this one as superseded.

PR #288, opened 2026-06-18
feat: env overrides to load bundled agents/intercom from read-only dirs

dbarrosop:
Add PI_SUBAGENT_EXTRA_AGENT_DIRS (PATH-style, split on path.delimiter) so a launcher can expose bundled agent definitions from read-only locations (e.g. a Nix store) without copying or symlinking them into the writable agent dir. Loaded as a "user" source, below the user's own agent dir on name collisions.

Add PI_INTERCOM_EXTENSION_DIR to override the pi-intercom bridge package directory, so the same hermetic install can point the bridge at a read-only path instead of seeding it under <agentDir>/extensions.

Both are no-ops when unset. Adds unit tests for each.

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -768,6 +768,22 @@ function resolveNearestProjectChainDirs(cwd: string): { readDirs: string[]; pref
 }
 const BUILTIN_AGENTS_DIR = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..", "..", "agents");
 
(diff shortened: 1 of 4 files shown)


nicobailon:
I pushed a small test isolation fix so the new extra-agent-dir tests do not read a developer's real ~/.agents directory. With that in place, the two new narrow unit files pass and the full unit suite passes locally (510 tests). This looks ready for maintainer review; I did not merge it.

PR #289, opened 2026-06-18
fix: disable completion guard for delegate

turret-at-jackandjill:
## Summary

- Marks the packaged `delegate` agent with `completionGuard: false`.
- Keeps mutation completion-guard behavior available on implementation-oriented agents like `worker` while making generic delegated read-only/output-artifact tasks safe by default.
- Adds coverage that packaged `delegate` loads with the guard disabled.

## Validation

- `node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/unit/agent-frontmatter.test.ts test/unit/completion-guard.test.ts`

## Context

Jo hit a false red on an async read-only `delegate` task with an output artifact: the child completed successfully, but the background runner reported `Subagent completed without making edits for an implementation task.` A Jo-side override works as a stopgap, but this is a better package-level default for the generic `delegate` agent.

### DIFF
--- a/agents/delegate.md
+++ b/agents/delegate.md
@@ -4,6 +4,7 @@ description: Lightweight subagent that inherits the parent model with no default
 systemPromptMode: append
 inheritProjectContext: true
 tools: read, grep, find, ls, bash, edit, write, contact_supervisor
+completionGuard: false
 inheritSkills: false
 ---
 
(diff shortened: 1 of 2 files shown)


turret-at-jackandjill:
Closing this after design feedback. `delegate` should remain generic/action-capable; globally disabling the mutation guard there weakens the wrong tool. The read-only path should use `scout` or a narrower completion-policy/read-only classification fix instead.

PR #311, opened 2026-06-20
feat: pi permission system integration

loss-and-quick:
Add permission-system `ask` forwarding (for `@gotgenes/pi-permission-system`) from
subagent child processes back to the parent Pi session UI, preserve per-agent
`permission:` frontmatter blocks, and document the integration.

This is a fresh, current-main revision of the previously closed PR. See
"Changes since the previous review" below for what changed.

## Why

`pi-permission-system` supports subprocess subagent prompt forwarding through the
`PI_SUBAGENT_PARENT_SESSION` environment variable.

Before this PR, subprocess children launched by `pi-subagents` did not reliably
receive that parent session identity, so `ask` permissions in child processes
could be auto-denied instead of being forwarded to the parent UI.

This PR implements the documented integration contract and makes `permission:`
frontmatter safe to use in `pi-subagents` agent files.

## Changes

- Set `PI_SUBAGENT_PARENT_SESSION` at session start in the interactive (root)
  session; child subprocesses inherit it through the process environment.
- Explicitly propagate the launching session identity through:
  - foreground single runs
  - foreground parallel runs
  - foreground chain runs
  - async single runs
  - async dynamic-fanout children
- Keep inherited `process.env.PI_SUBAGENT_PARENT_SESSION` as a defensive fallback.
- Preserve nested `permission:` YAML frontmatter through agent discovery and
  serialization.
- Add a `/subagents-doctor` "Permission system" section reporting the signals
  pi-subagents actually owns: parent-session env status and subagent-process
  detection.
- Document the integration in `README.md`, scoped to behavior the permission
  system can actually resolve.
- Regression coverage for: explicit parent-session env precedence, inherited env
  fallback, `permission:` frontmatter round-trip, and the dynamic-fanout
  runner-vs-user validation split.

## Changes since the previous review

- **Rebased onto current main** (was conflicting). The `fanout-child` registration
  moved out of `index.ts` into a standalone extension upstream; the parent-session
  env wiring was re-applied on top of that structure.
- **Async dynamic fanout fixed.** `parentSessionId` is now allowed on the
  runner-internal dynamic parallel template only, so dynamic async children
  receive the parent session without the field leaking into user-facing dynamic
  validation (a `.chain.json` that sets `parentSessionId` is still rejected).
  Previously the injected field was rejected as an unsupported dynamic fanout
  field and the async dynamic path failed.
- **Async single now propagates explicitly** instead of relying only on the
  inherited env fallback.
- **Docs/diagnostics scoped down.** Dropped the inaccurate "nested forwarding goes
  directly to the root UI session" / agent-name-metadata claims (a nested child's
  direct parent is a headless subagent with no UI), and removed the doctor's
  guessed pi-permission-system config-path check that produced false
  "not installed" diagnostics. `pi list` is the install check.

## Example

```yaml
---
name: worker
tools: bash,read,write,edit
permission:
  "*": ask
  read: allow
  bash:
    "*": ask
    "git *": allow
    "npm test": allow
---

You are a worker agent.
```

`tools:` continues to control tool visibility via `pi-subagents`. `permission:` is
consumed independently by `@gotgenes/pi-permission-system` as the runtime policy
layer. The two layers compose without collision.

`ask` forwarding resolves an interactive prompt for direct children of the
interactive session; place `ask` policies on agents that run as direct children.

## Validation

```bash
node --experimental-strip-types --test \
  test/unit/pi-args.test.ts \
  test/unit/agent-frontmatter.test.ts \
  test/unit/dynamic-fanout.test.ts \
  test/unit/doctor.test.ts
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,77 @@ For normal use, you do not need to configure anything. Advanced users can tune t
 
 At this point, you know enough to use the plugin. The rest of this README is reference material for exact command syntax, custom agents, saved chains, worktrees, and configuration.
 
+## Optional pi-permission-system integration
+
+[`@gotgenes/pi-permission-system`](https://github.com/gotgenes/pi-packages/tree/main/packages/pi-permission-system)
+adds a second policy layer — `allow` / `ask` / `deny` — on top of
+pi-subagents' visibility-based tool restrictions.
+
+The two compose independently:
+
+| Layer | What it controls | Who provides it |
+|-------|-----------------|-----------------|
+| Visibility | Which tools are registered before the session starts | pi-subagents (`tools:` frontmatter key) |
+| Policy  | Runtime allow/ask/deny decisions on every tool call, bash command, MCP operation | pi-permission-system (`permission:` frontmatter key) |
+
+### Installing
+
+```bash
+pi install npm:@gotgenes/pi-permission-system
+```
+
+No configuration is required for the integration — it is automatic when both
+extensions are installed. pi-subagents passes the parent session identity
+to child processes via the `PI_SUBAGENT_PARENT_SESSION` environment variable,
+which the permission system uses to forward `ask` prompts from headless
+subagent processes back to the parent session's UI.
+
+### Per-agent permission frontmatter
+
+Agent files can include a `permission:` block alongside the standard `tools:`
+key. The permission system reads it independently:
+
+```yaml
+---
+name: worker
+tools: bash,read,write,edit
+permission:
+  "*": ask
+  read: allow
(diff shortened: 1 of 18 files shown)


nicobailon:
This is now merge-ready from my review. I pushed a small follow-up so async launches pass the actual parent session id into the permission-forwarding env instead of reusing the async status identity, which can be a session file path.

That keeps forwarded requests aligned with the session id the permission system polls in the parent UI session. The merged tree passes the focused regression and the broader unit/integration suites.

PR #312, opened 2026-06-20
feat: parallel groups, metadata, group options, completion via /chain

loss-and-quick:
Brings the inline `/chain` slash command to parity with the `ChainStep` model for
everything that fits on a single line: static parallel groups, per-step/per-task
metadata, group-level options, and group-aware tab completion. Dynamic fan-out
(`expand`/`collect`) is intentionally left out — it stays available through the
`subagent({ chain: [...] })` tool API and saved `.chain.json` files.

This is a fresh, current-main revision of the previously closed PR. See
"Changes since the previous review" below for what changed.

```text
/chain scout[as=ctx,phase=recon] "scan" -> (reviewer "review A" | reviewer[count=2] "review B")[concurrency=2,failFast] -> writer[outputSchema=schema.json,acceptance=verified] "fix"
```

### Parallel groups (base)
Wrap agents in `( ... )` separated by `|` to fan them out as a single
`{ parallel: [...] }` chain step; the next `->` step continues once they finish.
A group must *open* a step — only a step that begins with `(` is parsed as a group,
so parentheses inside a shared `--` task (e.g. `/chain scout -- inspect auth (backend)`)
stay literal and keep the legacy single-agent behavior. Linear `/chain` behavior is
unchanged.

### Inline step/task metadata
Extended the `[key=value,...]` config with:

| Key            | Description                                                           |
| -------------- | --------------------------------------------------------------------- |
| `as`           | Name this step's output for later reference.                          |
| `label`        | Human-readable step label.                                            |
| `phase`        | Group steps into a named phase.                                       |
| `cwd`          | Run the step in a subdirectory.                                       |
| `count`        | Fan a group task into N copies (only inside a `( ... )` group).       |
| `outputSchema` | Path to a JSON Schema file, read relative to cwd and validated.       |
| `acceptance`   | Acceptance level (`none`/`attested`/`checked`/`verified`/`reviewed`). |

`outputSchema` is validated via `assertJsonSchemaObject`, `acceptance` via
`validateAcceptanceInput`; parse/load errors surface through `ctx.ui.notify`
instead of throwing. Inline values must not contain spaces or commas (unchanged
constraint).

### Group-level options
A trailing `[...]` suffix after the closing `)` sets step-level options on the group:

```text
/chain a "x" -> (b "y" | c "z")[concurrency=2,failFast,worktree]
```

`concurrency`, `failFast`, and `worktree` map onto the emitted parallel step.

### Group-aware tab completion
The multi-agent completion finds the current step after the last top-level `->` or
`(`, or after a `|` *inside* a group — tracking quotes and paren depth so separators
inside a task are ignored. Agents are suggested after an opening paren, after a `|`
inside a group, and on each new arrow step. A bare `->`/`|` gets a separating space.
Completion is suppressed inside an open quote and once a top-level `--` task has
begun. Linear-chain completion is unchanged.

## Changes since the previous review

- **Rebased onto current main** (was conflicting). Applied cleanly with no behavioral
  changes to the group work.
- **Unquoted parens in a `--` task no longer rejected.** `hasGroupSyntax` now treats
  `(` as group syntax only when it opens a top-level step, so a normal shared-task
  command like `/chain scout -- inspect auth (backend)` parses as plain task text
  instead of failing as a malformed group.
- **Completion no longer resumes after a `--` task.** A `|` only restarts agent
  suggestions inside a `( ... )` group; outside a group it is task text, so
  `scout -- do x | wr` no longer wrongly suggests `writer` past the `--`.
- Added focused regressions for both cases (parser + completion) and documented the
  "a group opens with `(`" rule in `README.md`.

## Non-goals
- Dynamic fan-out (`expand`/`collect`) inline — data-driven fan-out relies on JSON
  pointer paths and schemas that don't fit a single slash line.
- Full `acceptance` config objects inline — only the level shorthand is supported;
  use a saved chain for the full structure.

## Validation

```bash
node --experimental-strip-types --test test/unit/slash-chain-groups.test.ts
node --experimental-transform-types --import ./test/support/register-loader.mjs \
  --test test/integration/slash-commands.test.ts
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,6 +329,7 @@ Skip this section until you want exact syntax.
 |---------|-------------|
 | `/run <agent> [task]` | Run one agent; omit the task for self-contained agents |
 | `/chain agent1 "task1" -> agent2 "task2"` | Run agents in sequence |
+| `/chain scout "scan" -> (reviewer "A" \| reviewer "B") -> writer "fix"` | Run a chain with a static parallel group inline |
 | `/parallel agent1 "task1" -> agent2 "task2"` | Run agents in parallel |
 | `/run-chain <chainName> -- <task>` | Launch a saved `.chain.md` or `.chain.json` workflow |
 | `/subagents-doctor` | Show read-only setup diagnostics |
@@ -352,6 +353,21 @@ Both double and single quotes work. You can also use `--` as a delimiter:
 ```
 
 Steps without a task inherit behavior from the execution mode. Chain steps get `{previous}`, the prior step’s output. Parallel steps use the first available task as a fallback.
+
+### Inline parallel groups in `/chain`
+
+Wrap a group of agents in parentheses and separate them with `|` to fan them out within a single chain step. The group runs all of its tasks concurrently, then the next `->` step continues once they finish:
+
+```text
+/chain scout "scan" -> (reviewer "review A" | reviewer "review B") -> writer "fix"
+```
+
+Notes:
+
+- Groups must contain at least two tasks separated by ` | `.
+- Group syntax is only valid between ` -> ` separators, and the group must appear as a complete step.
+- A group is treated as the prior step’s output for the next sequential step.
+- Dynamic fanout (`expand` / `collect`) and chain metadata fields are not yet available inline; use the `subagent({ chain: [...] })` tool API or a saved `.chain.json` for those.
 
 ```text
 /chain scout "analyze auth" -> planner -> worker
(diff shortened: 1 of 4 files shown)


nicobailon (changes requested):
I would not merge this as-is. The new group parser still rejects ordinary task text with balanced parentheses once a real group appears, so `/chain scout "scan" -> (reviewer "A" | reviewer "B") -> writer -- fix (backend)` fails with `Unmatched parentheses in chain segment`, even though the same `writer -- fix (backend)` step works in a normal linear chain.

The PR says only a step that opens with `(` is a group, so balanced parentheses in a non-group `--` task should stay literal. The smallest fix I verified locally is to let those non-group segments pass through and add a regression in `test/unit/slash-chain-groups.test.ts`.

Issue #334, opened 2026-06-24
Async runner fails silently (0.31.0): src/shared/utils.ts value-imports uninstalled peer @earendil-works/pi-coding-agent


ashlineldridge:
Since 0.31.0, every async/parallel subagent run fails immediately with:

```
Async runner process <pid> exited or disappeared before writing a result.
Marked run failed by stale-run reconciliation.
```

There's no other output because the detached runner is spawned with `stdio: "ignore"`, so it looks transient and survives reloads/restarts.

## Root cause

The detached async runner loads `src/shared/utils.ts`, which value-imports `@earendil-works/pi-coding-agent`:

```
src/shared/utils.ts:8   import * as piCodingAgent from "@earendil-works/pi-coding-agent";
```

`pi-coding-agent` is a `peerDependencies: "*"`, so it isn't installed in pi-subagents' own tree. The runner is a separate `node` process (not the host pi), so it resolves from the filesystem and can't find the peer, crashing at startup. The in-process extension is fine because the host pi provides the module; only the detached runner hits this.

Load chain: `subagent-runner.ts` -> `artifacts.ts` -> `utils.ts` -> `@earendil-works/pi-coding-agent`.

## Reproduction

```bash
mkdir /tmp/pisa && cd /tmp/pisa && npm init -y >/dev/null

npm i pi-subagents@0.31.0 >/dev/null
./node_modules/.bin/jiti node_modules/pi-subagents/src/shared/utils.ts
# Error: Cannot find module '@earendil-works/pi-coding-agent'
#   Require stack: .../pi-subagents/src/shared/utils.ts        (exit 1)

npm i pi-subagents@0.30.0 >/dev/null
./node_modules/.bin/jiti node_modules/pi-subagents/src/shared/utils.ts
# loads clean                                                  (exit 0)
```

## When it changed

New in **0.31.0**:

- 0.30.0: `pi-coding-agent` is value-imported only from modules the runner doesn't load (`tui/render.ts`, `shared/fork-context.ts`); `utils.ts` does not import it.
- 0.31.0: adds the `utils.ts:8` import above, and `utils.ts` is in the runner's load chain.

`pi-coding-agent` / `pi-ai` / `pi-agent-core` have always been peer deps, so the gap was latent until this import landed in a runner-loaded module. (Same dependency class as #182, which fixed `pi-tui` by moving it into `dependencies`.)

## Potential fix

[...]

PR #340, opened 2026-06-26
Fix async runner peer import diagnostics

salemsayed:
## Summary

Fixes the async runner startup crash reported in #334 by removing the runner-loaded value import from `@earendil-works/pi-coding-agent` while preserving custom config-dir support.

This PR also makes future async runner startup failures diagnosable by capturing detached runner stdout/stderr and surfacing the stderr tail during stale-run reconciliation.

## What changed

- Avoid value-importing `@earendil-works/pi-coding-agent` from `src/shared/utils.ts`.
- Resolve the config directory name from, in order:
  1. explicit `codingAgentModule` argument,
  2. `PI_SUBAGENTS_CONFIG_DIR_NAME`,
  3. nearest entry-point `package.json` `piConfig.configDir`,
  4. default `.pi`.
- Pass `PI_SUBAGENTS_CONFIG_DIR_NAME` into detached async runners when the parent can derive it from the Pi package root.
- Capture detached runner stdout/stderr to `runner-stdout.log` and `runner-stderr.log` under the async run directory.
- Include a bounded stderr tail in stale-run failures so startup crashes are visible instead of only reporting that the process disappeared.
- Add regression coverage for the no-peer-import path, env propagation, stdio logs, and stale-run diagnostics.

## Validation

- `npm run test:all`
  - unit: 666 passed
  - integration: 420 passed

## Notes

This intentionally does not change completion-guard policy/inference behavior; it is scoped to the async runner peer-import crash and diagnostics path.

### DIFF
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -215,6 +215,74 @@ function resolveAsyncRunnerNodeCommand(): string {
 	return process.platform === "win32" ? "node.exe" : "node";
 }
 
+interface RunnerStdioFiles {
+	stdio: "ignore" | ["ignore", number, number];
+	closeParentFds: () => void;
+	stderrPath?: string;
+}
+
+const CONFIG_DIR_NAME_ENV = "PI_SUBAGENTS_CONFIG_DIR_NAME";
+
+function cleanString(value: unknown): string | undefined {
+	return typeof value === "string" && value.trim() ? value : undefined;
+}
+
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks for this. Most of the peer-import pieces were superseded by the async runner fixes that landed on main, and I adapted the useful stderr diagnostic piece there too. Closing this PR now.

salemsayed:
noted, thank you for your help, appreciated

PR #341, opened 2026-06-26
Add PI_SUBAGENT_PI_BINARY override for child Pi launches

dbarrosop:
## Summary

- Add `PI_SUBAGENT_PI_BINARY` support to `getPiSpawnCommand`
- Use the override to launch child Pi processes through an explicit wrapper/binary
- Document the new env var and add changelog entry
- Add regression coverage for override and blank-value fallback behavior

## Why

Packaged wrappers like need child subagents to launch through the same wrapper so they inherit bundled resources, flags, and environment setup without relying on fragile `PATH` ordering.

## Testing

- `node --experimental-strip-types --test test/unit/pi-spawn.test.ts`
- `env -u PI_SUBAGENT_PI_BINARY -u PI_CODING_AGENT_DIR -u PI_INTERCOM_EXTENSION_DIR npm run test:unit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 - Added `/chain` inline parallel groups with per-step metadata, group options, and tab completion.
 - Added subagent profile commands and provider model catalog generation for quota and quality model profiles.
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks for this. I landed the PI_SUBAGENT_PI_BINARY override on main with the changelog rebase fix, so I’m closing this PR now.

PR #342, opened 2026-06-26
fix: lazy-load pi-coding-agent to prevent async runner crash

huangkun3251:
The static import of @earendil-works/pi-coding-agent fails in async subprocesses because the package is a peer dependency installed at a different path than the subagent runner's module resolution expects. This caused every async subagent to crash immediately with exit code 1.

Replace the module-level import with a lazy require() wrapped in try/catch. When pi-coding-agent is unavailable, resolveConfigDirName falls back to the default '.pi' directory name. Foreground mode is unaffected since it runs under the pi binary where resolution works.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -5,7 +5,6 @@
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
-import * as piCodingAgent from "@earendil-works/pi-coding-agent";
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks for sending this. The async peer-import issue is now covered on main through the reviewed peer-import fixes, so this PR is superseded and I’m closing it.

PR #344, opened 2026-06-27
fix: cold-start empty output treated as success + fork-context parallel serialization

### LINKED ISSUE #161: Fork-context parallel chain steps serialize instead of running concurrently
## Description

Parallel chain steps using fork-context agents (worker, planner, oracle) execute serially instead of concurrently, despite `mapConcurrent` using `MAX_CONCURRENCY = 4`. Observed wall time equals the sum of individual task times, not the longest task time.

## Steps to Reproduce

A chain with a parallel block of 5 `worker` agents:

```json
{
  "chain": [
    {
      "parallel": [
        { "agent": "worker", "task": "Task A" },
        { "agent": "worker", "task": "Task B" },
        { "agent": "worker", "task": "Task C" },
        { "agent": "worker", "task": "Task D" },
        { "agent": "worker", "task": "Task E" }
      ]
    }
  ]
}
```

## Expected Behavior

4 workers start concurrently (MAX_CONCURRENCY = 4), the 5th queues. Wall time approximately equals longest worker time.

## Actual Behavior

Workers run one at a time: worker A runs while B through E show "pending." Wall time equals the sum of all five worker times. Observed on a real run: 5 explorers took ~6 minutes serially (~1 min each, cumulatively summed) instead of ~1.5 minutes in parallel.

## Analysis

The bottleneck appears to be in fork session creation (`fork-context.ts`). Each worker calls `sessionFileForIndex(N)` inside the `mapConcurrent` worker body, which triggers `sessionManager.constructor.open(parentSessionFile)` on the **same parent session file**:

https://github.com/nicobailon/pi-subagents/blob/main/src/shared/fork-context.ts

Multiple concurrent workers all attempt to open/branch from the same parent session file. This serializes before the child process even spawns.

`mapConcurrent` itself is correctly parallel (uses `Promise.all` with a worker pool in `parallel-utils.ts`). The serialization happens before its parallelism can take effect.

## Affected Agents

Only agents with `defaultContext: fork` are affected: `worker`, `planner`, `oracle`. Unaffected: `scout`, `reviewer`, `researcher`, `context-builder`, `delegate` (all default to fresh).

## Suggested Fix Direction

Pre-create fork session files before dispatching parallel workers, so each worker receives an already-created session file:

```typescript
// Before mapConcurrent, pre-warm all fork sessions:
if (context === "fork") {
  for (let i = 0; i < tasks.length; i++) {
    sessionFileForIndex(i);
  }
}
// Then dispatch mapConcurrent with no fork contention
```

Alternative: make `sessionFileForIndex` promise-safe so concurrent callers don't race the same index, or emit a warning when fork-context parallelism is expected to degrade.

## Environment

- pi-subagents: v0.24.0
- Pi: @earendil-works/pi-coding-agent v.0.74.0
- Node.js: v24.14.0
- OS: Ubuntu 24.04

### LINKED ISSUE #178: Bug: Empty output from local models (cold-start) treated as success — no retry attempted
## Problem

Local models (observed with gemma4 on Ollama 0.23.1) produce empty or incomplete responses on first 2-3 inferences after loading into VRAM ("cold start"). The subagent executor treats these as success (exitCode 0, no error), producing 0-token empty results to the parent. No retry is attempted because `attemptSucceeded = exitCode === 0 && !result.error` evaluates to `true`.

## Reproduction

1. Use a local Ollama model (gemma4-tool:latest) for subagents
2. Ensure model is unloaded (`ollama stop gemma4-tool:latest`)
3. Run a subagent (scout, worker, test-runner)
4. **Cold start**: First run produces 0 tokens, empty output, but exitCode 0
5. **After warmup** (3-5 inferences): model works perfectly

### Session evidence (scout, cold start)

```jsonl
{"role":"assistant","content":[{"type":"thinking","thinking":"..."}],
 "stopReason":"stop","usage":{"input":0,"output":0,"totalTokens":0}}
```
Single turn, no tool calls. `getFinalOutput` → `""`.

### Session evidence (test-runner, cold start)

```jsonl
{"role":"assistant","content":[{"type":"toolCall","name":"bash",...}],
 "stopReason":"toolUse"}
{"role":"toolResult","content":"hello from gemma4\n"}
{"role":"assistant","content":[],  ← EMPTY content
 "stopReason":"stop","usage":{"totalTokens":0}}
```
Two turns, tool executed, but final assistant response has empty content.

## Root cause

In `src/runs/foreground/execution.ts`, `runSingleAttempt`:

```typescript
if (result.exitCode === 0 && !result.error) {
    const errInfo = detectSubagentError(result.messages);
    // detectSubagentError only checks tool errors and bash exit codes.
    // Empty model output is never flagged as an error.
}
```

Then in `runSync`:
```typescript
const attemptSucceeded = result.exitCode === 0 && !result.error;
if (attemptSucceeded) break;  // ← exits retry loop immediately
```

`detectSubagentError` checks tool failures and bash exit codes but never detects the case where the model produces zero meaningful output. Since exitCode is 0 and no error is set, the retry loop breaks.

## Proposed fix

Add empty-output detection after `detectSubagentError`:

```typescript
if (!result.error) {
    const finalText = getFinalOutput(result.messages);
    if (!finalText.trim()) {
        result.error = "Subagent produced no output (possible model cold-start or empty response).";
        result.exitCode = 1;
    }
}
```

And add matching retryable patterns in `model-fallback.ts`:
```typescript
/cold.start/i,
/empty response/i,
/no output/i,
/model.*(?:load|fail|error)/i,
```

## Applied locally

Both changes applied to local node_modules. Jiti cache verified. Ready for PR.

## Environment

- pi 0.74.0
- pi-subagents 0.24.0 (npm)
- Ollama 0.23.1
- gemma4-tool:latest (9.6GB, CUDA)
- NixOS, NVIDIA RTX

gr3enarr0w:
## Fixes

Closes #178
Closes #161

### fix(execution): detect empty model output as retryable error (#178)

Local models (Ollama, gemma4, etc.) can return 0-token empty responses on cold start. The previous `detectSubagentError` check only caught tool failures and bash exit codes — empty model output fell through as `exitCode === 0 && !result.error`, breaking out of the retry loop immediately.

Added a blank-output check after `detectSubagentError`:
```typescript
if (result.exitCode === 0 && !result.error) {
    const finalText = getFinalOutput(result.messages);
    if (!finalText?.trim()) {
        result.exitCode = 1;
        result.error = "Subagent produced no output (possible model cold-start or empty response).";
    }
}
```

### fix(model-fallback): add cold-start retry patterns (#178)

Added `cold-start`, `empty response`, `no output`, and `model load/fail/error` to `RETRYABLE_MODEL_FAILURE_PATTERNS` so the fallback model loop fires on these new error strings.

### fix(subagent-executor): pre-warm fork sessions before mapConcurrent (#161)

Fork-context parallel workers (`worker`, `planner`, `oracle`) were serializing because each called `sessionFileForIndex(i)` concurrently inside `mapConcurrent`, racing to branch the same parent session file. `sessionFileForIndex` has an index-keyed cache but the first call per index triggers `SessionManager.open(parentSessionFile)` — this I/O on a shared file path serializes under concurrent access.

Fix: iterate all indices sequentially before `mapConcurrent` so every session file is created and cached before parallel dispatch begins. Cache hits inside the worker bodies return instantly.

```typescript
// Pre-warm fork session files sequentially before concurrent dispatch
for (let i = 0; i < input.tasks.length; i++) {
    input.sessionFileForIndex(i);
}
return mapConcurrent(input.tasks, input.concurrencyLimit, async (task, index) => { ... });
```

Observed speedup: 5 fork-context workers that took ~6 min serially now complete in ~1.5 min in parallel.

### DIFF
--- a/src/runs/foreground/execution.ts
+++ b/src/runs/foreground/execution.ts
@@ -772,6 +772,13 @@ async function runSingleAttempt(
 				: `${errInfo.errorType} failed with exit code ${errInfo.exitCode}`;
 		}
 	}
+	if (result.exitCode === 0 && !result.error) {
+		const finalText = getFinalOutput(result.messages);
+		if (!finalText?.trim()) {
+			result.exitCode = 1;
+			result.error = "Subagent produced no output (possible model cold-start or empty response).";
+		}
+	}
 	if (options.structuredOutput && result.exitCode === 0 && !result.error) {
 		const structured = readStructuredOutput({
 			schema: options.structuredOutput.schema,
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -1938,6 +1938,12 @@ function findDuplicateParallelOutputPath(input: {
 }
 
 async function runForegroundParallelTasks(input: ForegroundParallelRunInput): Promise<SingleResult[]> {
+	// Pre-warm fork session files sequentially before concurrent dispatch to avoid
+	// races where multiple workers simultaneously try to branch the same parent session.
+	// sessionFileForIndex caches results, so these calls return instantly inside mapConcurrent.
+	for (let i = 0; i < input.tasks.length; i++) {
+		input.sessionFileForIndex(i);
+	}
 	return mapConcurrent(input.tasks, input.concurrencyLimit, async (task, index) => {
 		const behavior = input.behaviors[index];
 		const effectiveSkills = behavior?.skills;
--- a/src/runs/shared/model-fallback.ts
+++ b/src/runs/shared/model-fallback.ts
@@ -126,6 +126,10 @@ const RETRYABLE_MODEL_FAILURE_PATTERNS = [
 	/\b502\b/,
 	/\b503\b/,
 	/\b504\b/,
+	/cold.?start/i,
+	/empty response/i,
+	/no output/i,
+	/model.*(?:load|fail|error)/i,
 ];
 
 export function isRetryableModelFailure(error: string | undefined): boolean {


nicobailon:
Thanks for this. I landed the cold-start empty-output and fork-session work on main with the reviewed async and structured-output fixes, so I’m closing this PR now.

PR #347, opened 2026-06-27
fix(async): fall back from stale node exec path

0xRichardH:
## Summary

- Reuse `process.execPath` for the async runner only when it is node-like and accessible.
- Fall back to `node` / `node.exe` from PATH when the current Node executable is stale or deleted.
- Add regression coverage for a node-like `process.execPath` that no longer exists.

## Why

Long-running Pi sessions can survive a mise/Node upgrade where the original Node binary is deleted. In that state, `process.execPath` still looks like Node by basename, but spawning it fails before the async runner gets a PID.

This happened locally and was fixed by falling back to PATH `node` when the node-like `process.execPath` is no longer usable.

## Validation

- Tested locally in the failing Pi environment; async subagent startup works after this patch.
- `npm run test:integration -- --test-name-pattern='process.execPath|stale|cannot spawn' test/integration/async-execution.test.ts`
- `npm run test:unit`
- Reviewer subagent also ran `npm run test:integration -- test/integration/async-execution.test.ts` successfully: 420 passed, 0 failed.

## Notes

This intentionally preserves the existing behavior from #273:
- Node-like and accessible `process.execPath` is reused.
- Non-Node or stale/deleted `process.execPath` falls back to PATH `node` / `node.exe`.

### DIFF
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -207,9 +207,22 @@ export function isAsyncAvailable(): boolean {
 	return jitiCliPath !== undefined;
 }
 
+function isNodeExecutableName(execPath: string): boolean {
+	const basename = path.basename(execPath).toLowerCase();
+	return basename === "node" || basename === "node.exe" || basename === "nodejs" || basename === "nodejs.exe";
+}
+
+function canUseCurrentNodeExecutable(execPath: string): boolean {
+	try {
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for this. I landed the stale Node executable fallback on main and kept the changelog credit, so I’m closing this PR now.

PR #348, opened 2026-06-27
feat: surface totalCost rollup on run results (#345)

### LINKED ISSUE #345: feat: cost aggregation across parallel chains
## Problem

When running parallel subagent chains, there is no way to see the total token cost for an entire run. Costs are tracked per-agent but never rolled up to the chain or workflow level.

## Proposed Solution

Add a `CostAccumulator` that collects `{ inputTokens, outputTokens, costUsd }` from each agent result and surfaces a rollup on the `SingleResult[]` return value from `runForegroundParallelTasks` and `runForegroundChainTasks`. The final `RunResult` should include a `totalCost` field.

## Acceptance Criteria
- `RunResult` includes `totalCost: { inputTokens: number, outputTokens: number, costUsd: number }`
- Cost is summed across all agents in parallel and chain runs
- Nested subagent costs propagate up to parent run
- Displayed in the TUI progress view alongside agent count

gr3enarr0w:
## Summary

Adds cost aggregation across all run types so callers can see total token spend for an entire parallel or chain run.

- `src/shared/utils.ts` — `sumResultsCost(results)` helper that sums `usage.{input,output,cost}` across all `SingleResult[]`; returns `undefined` when all values are zero
- `src/shared/types.ts` — `totalCost?: { inputTokens, outputTokens, costUsd }` added to `Details`
- `src/runs/foreground/chain-execution.ts` — wired into `buildChainExecutionDetails` (covers sequential, parallel-step, and dynamic fanout)
- `src/runs/foreground/subagent-executor.ts` — wired into `runParallelPath` and `runSinglePath`

Closes #345

## Test plan
- [ ] Parallel run with multiple agents shows non-zero `totalCost` on result
- [ ] Chain run accumulates cost across all steps
- [ ] Single-agent run has `totalCost` matching its own usage
- [ ] Run with no token data returns `totalCost: undefined`

### DIFF
--- a/src/runs/foreground/chain-execution.ts
+++ b/src/runs/foreground/chain-execution.ts
@@ -34,7 +34,7 @@ import { discoverAvailableSkills, normalizeSkillInput } from "../../agents/skill
 import { INTERCOM_BRIDGE_MARKER } from "../../intercom/intercom-bridge.ts";
 import { runSync } from "./execution.ts";
 import { buildChainSummary } from "../../shared/formatters.ts";
-import { compactForegroundDetails, getSingleResultOutput, mapConcurrent, resolveChildCwd } from "../../shared/utils.ts";
+import { compactForegroundDetails, getSingleResultOutput, mapConcurrent, resolveChildCwd, sumResultsCost } from "../../shared/utils.ts";
 import { recordRun } from "../shared/run-history.ts";
 import {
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks for this. I landed the total-cost rollup work on main with the reviewed scope and test fixes, so I’m closing this PR now.

PR #352, opened 2026-06-29
fix: avoid runtime peer import in async runner

### LINKED ISSUE #353: Async subagents can fail before writing results due runtime Pi peer import
## Bug

Async/background subagent runs can fail before the detached runner writes `status.json` or a result file when `@earendil-works/pi-coding-agent` is not resolvable from the `pi-subagents` package install.

## What happens

The parent reports a generic stale-run failure, for example:

```text
Async runner process <pid> exited or disappeared before writing a result. Marked run failed by stale-run reconciliation.
```

Running the detached runner in the foreground exposed the hidden startup error:

```text
Cannot find module '@earendil-works/pi-coding-agent'
Require stack:
- .../pi-subagents/src/shared/utils.ts
```

## Why

`src/shared/utils.ts` imports `@earendil-works/pi-coding-agent` at runtime only to read `CONFIG_DIR_NAME`. The detached async runner is launched as a standalone `node <jiti-cli> src/runs/background/subagent-runner.ts <cfg>` process, outside Pi's extension-loader module context. In package installs where the Pi peer package is not resolvable from `pi-subagents/node_modules`, the top-level import crashes before the runner can write diagnostic status/result files.

## Expected behavior

Detached async runs should not require a runtime value import of the Pi peer package just to resolve the config directory name. They should start successfully and preserve configurable config-dir behavior where possible.

## Proposed fix

PR: https://github.com/nicobailon/pi-subagents/pull/352

The PR removes the runtime peer import, resolves `piConfig.configDir` from Pi package metadata, passes the resolved Pi package root to detached async runner processes, and adds unit coverage.

aurbina83:
## Summary

- Remove the runtime value import of `@earendil-works/pi-coding-agent` from shared utilities used by the detached async runner.
- Preserve configurable Pi config directory resolution by reading `piConfig.configDir` from Pi package metadata instead of importing the peer package.
- Pass the resolved Pi package root into detached async runner processes via env so hoisted `jiti` launches can still find package metadata.
- Add unit coverage for fallback, explicit module shape, package metadata, explicit package root, and no runtime peer import.

## Why

Detached async runner processes start outside Pi's extension-loader module context. If `pi-subagents` does not have the Pi package peer resolvable from its own install, a top-level runtime import of `@earendil-works/pi-coding-agent` can crash the runner before it writes status/result files. The parent then reports only a stale-run failure.

## Test plan

- [x] `node --experimental-strip-types --test test/unit/config-dir-runtime.test.ts test/unit/pi-args.test.ts test/unit/pi-coding-agent-dir.test.ts test/unit/package-manifest.test.ts`
- [x] Ran full `npm run test:unit`; unrelated baseline failures observed:
  - `test/unit/agent-frontmatter.test.ts` invalid package normalization expectation
  - two `test/unit/worktree.test.ts` failures due local 1Password git signing (`failed to write commit object`)

Fixes #353

### DIFF
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -16,7 +16,7 @@ import { buildChainInstructions, isDynamicParallelStep, isParallelStep, resolveS
 import type { RunnerStep } from "../shared/parallel-utils.ts";
 import { resolvePiPackageRoot } from "../shared/pi-spawn.ts";
 import { buildSkillInjection, normalizeSkillInput, resolveSkillsWithFallback } from "../../agents/skills.ts";
-import { resolveChildCwd } from "../../shared/utils.ts";
+import { PI_CODING_AGENT_PACKAGE_ROOT_ENV, resolveChildCwd } from "../../shared/utils.ts";
 import { buildModelCandidates, resolveModelCandidate, resolveSubagentModelOverride, type AvailableModelInfo, type ParentModel } from "../shared/model-fallback.ts";
 import { resolveEffectiveThinking } from "../../shared/model-info.ts";
 import { resolveExpectedWorktreeAgentCwd } from "../shared/worktree.ts";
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for this. I landed the async runner peer-import fix on main with the reviewed follow-up fixes, so I’m closing this PR now.

PR #358, opened 2026-06-30
fix: capture async runner stdio logs

### LINKED ISSUE #357: Capture detached async runner stderr/stdout in run directory
## Problem

When a detached async subagent runner crashes during startup, the parent process currently reports only the stale-run reconciliation fallback, for example:

```text
Async runner process <pid> exited or disappeared before writing a result. Marked run failed by stale-run reconciliation.
```

The underlying stderr/stdout is lost because the detached runner is spawned with `stdio: "ignore"`.

This made #334/#353 harder to diagnose: the actionable startup error was a hidden `Cannot find module '@earendil-works/pi-coding-agent'`, but the async run directory had only `status.json` / `events.jsonl` and no child session/result file.

## Expected behavior

Async run directories should retain detached runner stdout/stderr, especially for startup failures before a result file exists. A failed run should leave enough local evidence to distinguish:

- runner bootstrap/import failures,
- config parsing failures,
- child Pi process failures,
- stale-run reconciliation fallback.

## Proposed fix

Capture detached runner stdio into files under the async run directory, e.g.:

- `runner.stdout.log`
- `runner.stderr.log`

This is complementary to the runtime peer-import fix in #352 and would make future startup crashes self-diagnosing rather than opaque.

danim47c:
## Summary

- Capture detached async runner stdout/stderr into each async run directory.
- Add `runner.stdout.log` and `runner.stderr.log` path resolution coverage.
- Preserve detached/unref behavior while closing parent-side log fds after spawn.

## Why

When the async runner crashes before writing a result, stale-run reconciliation currently reports only that the process disappeared. Capturing runner stdio makes startup failures diagnosable from the run directory instead of losing the underlying error.

This complements #352 / #334 / #353: even after the peer-import root cause is fixed, future runner bootstrap failures should leave logs.

Fixes #357.

## Test plan

- [x] `node --experimental-strip-types --test test/unit/async-execution.test.ts test/unit/stale-run-reconciler.test.ts`
- [x] `git diff --check`
- [x] Attempted full unit suite with `node --experimental-strip-types --test test/unit/*.test.ts`; it still has unrelated baseline failures in `test/unit/slash-chain-groups.test.ts` reporting `Unknown agent: scout`.

### DIFF
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -215,6 +215,26 @@ function resolveAsyncRunnerNodeCommand(): string {
 	return process.platform === "win32" ? "node.exe" : "node";
 }
 
+export function resolveAsyncRunnerLogPaths(cfg: object): { stdoutPath: string; stderrPath: string } | undefined {
+	const asyncDir = typeof (cfg as { asyncDir?: unknown }).asyncDir === "string"
+		? (cfg as { asyncDir: string }).asyncDir
+		: undefined;
+	if (!asyncDir) return undefined;
+	return {
+		stdoutPath: path.join(asyncDir, "runner.stdout.log"),
+		stderrPath: path.join(asyncDir, "runner.stderr.log"),
+	};
+}
+
+function closeFd(fd: number | undefined): void {
+	if (fd === undefined) return;
+	try {
+		fs.closeSync(fd);
+	} catch {
+		// Best-effort cleanup; child process already owns its duplicated stdio fd.
+	}
+}
+
 /**
  * Spawn the async runner process
  */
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for this. I landed the reviewed async runner stdio logging work on main, with the batch follow-up fixes included, so I’m closing this PR now.

danim47c:
Thanks Nico, really appreciate the super quick turnaround 🙌

Also just wanted to say: great repo/tooling.

I’m using pi-subagents a lot lately and it’s been genuinely useful. Happy to keep dogfooding and sending small fixes when I run into things.

PR #360, opened 2026-07-01
Recommend companion packages

nicobailon:
This adds a gentle companion-package hint when pi-subagents can work better with pi-intercom or pi-prompt-template-model. It only suggests install commands, keeps the note visible to the agent, and lets people hide it per workspace or globally.

I also wired the status into doctor and covered the detection and dismissal paths in tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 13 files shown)


nicobailon:
Landed the reviewed version on main with the follow-up config edge-case fixes, so I’m closing this PR to keep the queue tidy.

PR #364, opened 2026-07-01
Fix live-detail keybinding hints

kylegl:
## Problem
When a chain starts, pi-subagents shows a live-detail hint. In my setup, that hint said to use `Ctrl+O`, but I had changed Pi's `app.tools.expand` keybinding in my Pi config. Pressing `Ctrl+O` did not expand live detail, which made the hint confusing.

## Summary
- derive live-detail and full-notification hints from Pi's configured `app.tools.expand` keybinding
- update the render-widget test shim to prove the hint follows a configured key instead of a hardcoded shortcut

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -16,7 +16,7 @@ import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
 import type { AgentToolResult } from "@earendil-works/pi-agent-core";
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks for this, @kylegl. I reimplemented the fix on main in 05019cda, including the missing render-fork-badge assertion update and changelog credit. Closing this PR because the fix is now on main.

PR #365, opened 2026-07-01
Add wait tool so background-launching skills can finish in one turn

robobryce:
## Problem

Background (async) subagent runs are detached. Today the guidance tells the parent to **end its turn** and rely on Pi to deliver the completion notification later. That works interactively, but breaks in two common cases:

1. **Run-to-completion skills.** A skill whose whole job is to keep N async workers in flight until all rounds finish is told to "end your turn" to wait — a direct contradiction with "run to completion / don't stop early".
2. **Non-interactive runs (`pi -p "..."`).** The entire task is a single turn. Once it ends there is no subsequent turn to receive the completion, so the still-running children are simply abandoned and their results never land.

## What this does

Adds a **`wait` tool** that blocks the current turn until the session's async runs reach a terminal state (`complete` / `failed` / `paused`), delivering each completion into the conversation as it arrives, then returns.

- `wait()` — wait for every active async run in this session
- `wait({ id })` — wait for a single run (id or prefix)
- `wait({ timeoutMs })` — cap the wait (the runs keep going regardless; default 30 min)

It builds on the existing `listAsyncRuns` reconciliation, so a crashed runner is marked failed rather than hanging the wait forever, and it resolves early if the turn is aborted. It's registered **parent-side only** (after the child-mode early return), so child subagents don't receive it.

Async-start guidance in the tool output, `SKILL.md`, and `README.md` now points at `wait()` instead of "end your turn", and explains that ending the turn to wait is only safe in an interactive session that will prompt the agent again.

## Testing

- New `test/unit/wait.test.ts`: immediate return when nothing is pending, resolve once all runs go terminal, session scoping, `id` targeting, timeout, and abort — all passing.
- Scoped the `index-child-registration` test's tool capture to the `subagent` tool by name so it's robust to additional registered tools.
- Full unit suite shows no new regressions (the pre-existing `worktree` / `agent-disabled` sandbox failures are unchanged from `main`).
- Smoke-tested the real extension via the integration loader: both `subagent` and `wait` register, and `wait()` executes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Added a `wait` tool that blocks until background (async) subagent runs in the session finish, delivering each completion into the conversation before it returns. This lets a subagent-launching skill run to completion in a single turn — including non-interactive `pi -p` runs, where ending the turn to wait for a completion notification would abandon the still-running children. Accepts `{ id }` to wait on one run and `{ timeoutMs }` to cap the wait, reconciles crashed runners so it cannot hang indefinitely, and returns early if the turn is aborted. Async-start guidance in the tool output, `SKILL.md`, and `README.md` now points to `wait` instead of "end your turn".
 - Added `subagents.defaultModel` so subagents can have a global default model separate from the parent session model. Thanks to Artem Timofeev (@atimofeev) for #339.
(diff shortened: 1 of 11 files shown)


nicobailon:
Thanks for this, @robobryce. I reimplemented the wait tool on main in 05019cda, resolved the changelog conflict, fixed the needs_attention-at-start case, tightened ambiguous id-prefix handling, and updated the docs/changelog credit. Closing this PR because the feature is now on main.

PR #386, opened 2026-07-03
fix: scope active-run restore to the starting session

### LINKED ISSUE #383: Since 0.32.0 my async'd subagents have been starting to show up in all active pi sessions and periodically affect the context in sessions the subagent didn't originate
Before I updated to 0.32.0 i used async subagents on most of my workflows. The subagent TUI was contained to that specific session.

However, since updating to 0.32.0 the async'd subagents have been showing up in other pi sessions in other directory cwds.

There was also a case where I had a new session open but hadn't prompted anything yet. The non-related subagent finished and the new session picked up the context. The originating session got the no activity check, saw that it was done, and kept going but it threw the blank session into a confused state because it had no original prompt or agent spinup to know what to do with the context.

diurnalist:
`listAsyncRuns` already had a `sessionId` option, and restoreActiveJobs
was passing state.currentSessionId through it, but the option was a
single ad hoc field bolted onto the loop. Generalize it into a small
composable filters mechanism (`AsyncRunFilter`, `filterBySessionId`,
`sessionFilters`) so restoreActiveJobs, wait, run-status, and the fleet
view all scope to the owning session the same way, and so a future
scope (e.g. by cwd, per #383's original config proposal) is a new
filter function instead of another option field threaded everywhere.

Verified with a script that seeds the shared async dir with runs from
several unrelated sessions and confirms restoreActiveJobs only revives
the one matching the resuming session, matching the reported behavior
in #383.

Fixes: #383

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Fixed
 - Scope async subagent completion notifications to the exact owning Pi session so another session in the same repo no longer receives result notices.
+- Restore only the async runs spawned by the session that is starting, instead of every active run in the shared async directory, so unrelated Pi sessions (and their footers) no longer pick up someone else's background subagents on launch. `listAsyncRuns` now takes a composable `filters` list instead of an ad hoc `sessionId` option, so future restore-scoping filters (e.g. by cwd) can be added without changing its call sites. Thanks to @EtienneChollet, @TwistedTabby, and @vicary for confirming the regression. Fixes #383.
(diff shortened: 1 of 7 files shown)


diurnalist:
This change is part of the following stack:

- #386 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

diurnalist:
Closing — turns out the underlying fix already landed in main via #369 before I started; this PR was just a refactor of already-working logic, not an actual fix for #383. Sorry for the noise!

PR #403, opened 2026-07-07
fix: preserve async resume model

nishuzumi:
## Summary
- Carry the original async child model/thinking metadata through resume target resolution.
- Use the original child model when reviving an async resume unless the caller supplies a new model.
- Add regression coverage for live/revived resume targets and spawned revive args.

## Tests
- node --experimental-strip-types --test test/unit/async-resume.test.ts
- node --experimental-transform-types --import ./test/support/register-loader.mjs --test test/integration/intercom-result-delivery.test.ts
- npm test

### DIFF
--- a/src/runs/background/async-resume.ts
+++ b/src/runs/background/async-resume.ts
@@ -35,6 +35,8 @@ export type AsyncResumeTarget = {
 	intercomTarget: string;
 	cwd?: string;
 	sessionFile?: string;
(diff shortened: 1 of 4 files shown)


nicobailon:
Closing this as superseded by #405, which merged your original async resume model preservation fix plus the follow-up metadata repairs. Thanks again for the contribution.

Issue #407, opened 2026-07-07
Agent/user should be able to stop subagents.


seans-leadsonline:
Was using a local LLM server and really bogged it down by accidentally spinning up too many subagents. I'd like there to be a way to stop agents manually for cases like this. I'd think the implementation would be a slash command available for force ending the agents, without having to ask the (also bogged down) main agent to interrupt/stop the subagents.

PR #408, opened 2026-07-07
feat: add non-resumable 'stop' action to force-kill async subagent runs

seans-leadsonline:
Proposed implementation to my [issue](https://github.com/nicobailon/pi-subagents/issues/407)


nicobailon:
Thanks again for filing #407 and putting together this implementation, Sean. I ended up landing the feature via #411 so we could make the no-arg flow selector-first and keep manual stop as its own `stopped` lifecycle instead of reusing timeout semantics. Your issue and PR were the reference point for the replacement implementation.

Issue #326, opened 2026-06-23
feat: Save generated markdown files and others to project-local .pi-subagents directory


carolitascl:
Please save all generated Markdown files and any other auxiliary artifacts under a dedicated .pi-subagents/ directory (or a similar project-scoped folder) within the repository. Keeping generated files isolated from source code helps maintain a clean project structure, reduces workspace clutter, and makes it easier to manage, review, and exclude generated artifacts when necessary.

nicobailon:
Fixed on main in e5734fe. Generated markdown outputs, progress files, chain run directories, and saved artifacts now default under the project-local `.pi-subagents/` tree, while explicit absolute output paths remain respected. Added docs, gitignore, and unit/integration coverage.

PR #421, opened 2026-07-10
refactor: nest project-local subagent artifacts under .pi/subagents

ohlulu:
Small follow-up to #326 / the `.pi-subagents/` default.

`.pi-subagents/` currently sits as its own top-level dot-folder in the project root. Every coding agent already claims its own project dot-folder these days — `.claude/`, `.cursor/`, `.codex/`, `.gemini/` — so repo roots fill up fast, and one consolidated folder per tool feels cleaner than two Pi-related ones. Pi's slot is `.pi/`, and this codebase already uses it for agent memory (`getProjectConfigDir()` → `<project>/.pi/`, used in `agent-memory.ts`). This PR reuses that same convention for artifacts.

## Changes

- `getProjectSubagentsDir()` now builds off `getProjectConfigDir()` instead of a hardcoded `.pi-subagents` constant. New default: `<project>/.pi/subagents/`. Since the config dir name is resolvable (custom Pi builds can override `.pi`), the artifact path follows it automatically.
- The watchdog ignore rules resolve the artifact directory per repo root instead of hardcoding the path, so custom config dir names stay ignored too. Added unit tests covering the default dir, a custom config dir, and the static `tmp/` / `node_modules/` rules.
- Updated README mentions, `.gitignore` (old `.pi-subagents/` entry kept for existing worktrees), and the tests that asserted the old path.
- Existing `.pi-subagents/` directories from prior runs are untouched (not migrated) — noted in the CHANGELOG, safe to delete manually.

If you'd rather keep the old path as default and gate this behind a config option / env var instead, I'm happy to go that route — wanted to start with the direct version since it's a pure rename with no behavior change, but no strong attachment to the approach if you see it differently.

## Testing

- `npm run test:unit` — 1085 passing
- Touched integration suites (`parallel-execution`, `async-execution`, `single-execution`) — 168 passing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 .pi/todos/
+.pi/subagents/
+# Legacy artifacts location; keep ignored until old worktrees are cleaned up.
 .pi-subagents/
 .DS_Store
 AGENTS.md
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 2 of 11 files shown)


nicobailon:
Thanks for the thoughtful cleanup and tests. I’m keeping `.pi-subagents` as the supported project artifact location for now because related extensions currently depend on it. I may reconsider consolidating under `.pi/` later, but I don’t want to introduce a partial ecosystem path change today, so I’m closing this PR.

PR #428, opened 2026-07-11
feat: support agent-local skill paths

kylegl:
## Problem

Agent definitions can select skills with `skills:`, but those names currently have to be discoverable through Pi's global/project skill catalog. This couples two separate concerns:

- agent files define the execution environment (model, tools, output, context behavior)
- skills define reusable behavioral instructions

A package that ships role-specific skills for subagents must therefore expose those skills to the parent/default Pi agent as well. Specialized planner, reviewer, or worker instructions then add unrelated entries to the parent's global skill catalog even though only the corresponding subagent needs them.

This lets agents load dedicated skill packs intended only for them, while keeping those skills portable as standard `SKILL.md` directories and out of the global skill catalog.

## Desired behavior

An agent can declare private skill discovery roots while continuing to select skills by name:

```yaml
skills: safe-bash, review-checklist
skillPath: ./skills, ./private/review-checklist/SKILL.md
```

`skillPath` follows the existing comma-separated frontmatter convention used by other list-valued agent fields.

- relative paths resolve from the agent definition file
- paths may point to a `SKILL.md` file or a discoverable skill directory
- `skillPath` discovers candidates; `skills` still controls selection
- local matches take precedence
- missing or unreadable local matches fall back to normal global discovery
- local candidates are scoped to that agent invocation and never enter Pi's parent/global catalog

## Solution

This PR adds the `skillPath` agent field, preserves it through serialization, passes the defining agent's directory through foreground and background execution, and extends skill resolution with an invocation-local candidate map.

Resolution remains local-first and only performs normal global discovery when no readable local candidate exists. Agents without `skillPath` continue through the existing resolution path.

## Validation

Added focused coverage for:

- comma-separated frontmatter parsing and serialization
- direct `SKILL.md` files and discovery directories
- agent-file-relative resolution
- local precedence and global fallback
- malformed global settings with a valid local skill
- unreadable local candidates falling back globally
- cross-agent invocation isolation
- foreground child prompt injection
- background single child prompt injection
- background chain/parallel child isolation

Results:

- focused frontmatter/resolver tests: 72 passing
- full integration suite: 465/466 on the latest run; the single failure was an unrelated existing timing assertion in dynamic fanout cancellation (`elapsed 3314ms`)
- an earlier full integration run passed before the test-only reductions

## Compatibility

Existing `skills:` configuration remains valid. `inheritSkills: false` can be combined with `skills` and `skillPath` so a child receives only explicitly configured private skills without inheriting the parent's discovered skill catalog.


nicobailon:
Thanks, Kyle. The agent-local `skillPath` direction is useful, and I’ll preserve your credit, but this branch now also contains unrelated telemetry, widget, architecture, watchdog, and repository-guidance changes. That makes it unsafe to review or merge as one feature. I’m closing this version and will replace it with a clean `skillPath`-only implementation.

kylegl:
Oh shoot my apologies was experimenting thought I had that on a separate branch.

PR #464, opened 2026-07-15
fix: support read-only output persistence

### LINKED ISSUE #426: Read-only agents block on output paths despite runner-side artifact persistence
## Summary

When a task configures `output` for a read-only agent, `pi-subagents` tells the child to “Write your findings to exactly this path” even though the child may have no write-capable tool. The runner can already persist the child's final response to that path, but the child is not told this. In observed runs, read-only analyzers treated the mismatch as a blocker, called blocking `contact_supervisor(reason="need_decision")`, stalled, hit turn budgets, and were interrupted/revived; revival then created additional failed attempts.

## Use case

Run parallel, read-only codebase analyzers with distinct reports:

```ts
subagent({
  tasks: [{
    agent: "codebase-analyzer", // tools: read, grep, find, ls
    task: "Analyze the runtime. Do not modify project files.",
    output: "/tmp/runtime.md",
    outputMode: "file-only"
  }],
  async: true,
  context: "fresh"
})
```

This is a supported and useful safety boundary: analysis remains read-only while the extension owns artifact persistence.

## Current behavior

`src/runs/shared/single-output.ts` injects the same direct-write instruction into the task and system prompt:

```text
Write your findings to exactly this path: /tmp/runtime.md
This path is authoritative for this run.
```

A child without `write`, `edit`, or `bash` cannot obey literally. The default intercom guidance says to contact the supervisor when blocked, so children can reasonably issue a blocking request such as:

```ts
contact_supervisor({
  reason: "need_decision",
  message: "The required artifact must be written, but no write tool is available."
})
```

The runner already has the desired fallback: `resolveSingleOutput()` calls `persistSingleOutput(outputPath, fallbackOutput)` when the child did not modify the file. Because that capability is invisible to the child, orchestration can degenerate into attention notices, repeated steering, turn-budget aborts, interrupts, and revivals instead of one successful report.

Observed with `pi-subagents` 0.34.0. Five read-only analyzers in one retry all blocked on this mismatch; failed/paused analyzer work consumed about 934k non-cache tokens before recovery was abandoned.

## Expected behavior

A read-only child should return the complete artifact normally; the runner should persist it to `output` without supervisor coordination. Missing direct-write capability should not be presented as a blocker or trigger retries.

## Recommended solution

Make output instructions capability-aware and explicit about runner persistence:

- If the effective child tools can write the output, retain the direct-write instruction.
- Otherwise inject: “Return the complete artifact in your final response. The runtime will persist it to `<path>`. Do not call `contact_supervisor` merely because no write tool is available.”
- Add the same rule to child-facing intercom guidance, not only the parent-only skill.
- Add tests for a read-only agent with `output` and `outputMode: "file-only"`: no write tool required, no supervisor request, final text persisted, compact reference returned.
- Consider a preflight diagnostic when task requirements conflict with effective tools.

## Evidence

Target `main` at `c940fe2`; relevant surfaces:

- `src/runs/shared/single-output.ts`: direct-write prompt injection and runner-side fallback persistence
- `src/intercom/intercom-bridge.ts`: blocking supervisor guidance
- `skills/pi-subagents/SKILL.md`: parent-only rule against supervisor use for output-artifact confusion

No upstream duplicate was found with searches for `output artifact write tool contact_supervisor`, `read-only output path`, or `resume model override`.

nicobailon:
Makes configured output instructions match the child’s effective tool capability. Explicit read-only builtin allowlists now return the complete artifact in the final response for runtime persistence, while mutation-capable children keep direct-write instructions.

The behavior is wired through foreground and async single, parallel, sequential chain, chain parallel, and dynamic fanout paths. Child supervisor guidance now makes clear that a missing write tool is not itself a blocker.

Validation: `npm run test:all` (1,126 unit passed, 1 skipped; 525 integration passed; 1 E2E passed). Fresh read-only review found no issues.

Fixes #426


Avg8888:
Thank you @nicobailon, love your work! 🙏

Issue #482, opened 2026-07-16
Defer turn-budget termination while a child tool call is active


juicetin:
## Incident

Run `672e31ef` launched a mutation-capable worker with `turnBudget: { maxTurns: 20, graceTurns: 3 }` and a 20-minute elapsed timeout. The process was killed after 4 minutes 28 seconds at 23 assistant turns and 42 tool calls while mid-mutation. Its resumed writer needed more than 38 turns.

Turn count measured model/tool cycles, not elapsed time, slice scope, buildability, or delivery progress.

## Runtime evidence

On current `main`:

1. `shouldAbortForTurnBudget` aborts at the exact hard limit when `terminalAssistantStop` is false.
2. `updateStepTurnBudget` runs on assistant `message_end`.
3. A message with `stopReason: "toolUse"` is non-terminal at the exact limit.
4. `activeChildTurnBudgetAborts` immediately sends `SIGINT`, then `SIGTERM` after 1 second and `SIGKILL` after 4 seconds.

The hard turn limit can therefore terminate the child immediately after it emits or starts a tool call.

## Runtime slice

Keep this separate from the guidance correction.

Investigate deferring hard turn-budget termination until no tool is active and the child reaches a safe assistant boundary. Add focused process-mode regressions for:

- exact hard limit on `stopReason: "toolUse"` does not signal the child while the tool is active;
- termination can proceed at a later safe assistant boundary;
- timeout and explicit user stop retain their documented behavior;
- status and partial-output fields remain truthful while termination is deferred.

The runtime cannot infer repository buildability or delivery safety. Do not make a generic runtime claim that a worktree is safe. Writer/fix-worker orchestration guidance must continue to prohibit hard turn/tool caps even if this runtime race is fixed.

PR #483, opened 2026-07-16
docs: prohibit hard budgets for writer agents

juicetin:
## Proposal

Correct pi-subagents orchestration guidance after live incident `672e31ef`. Hard assistant-turn and tool-call caps are suitable for bounded read-only work, not mutation-capable writers.

### Exact change

- Prohibit `turnBudget` and hard `toolBudget` on implementation workers, fix workers, reviewers with edit authority, and other mutation-capable children.
- Direct writer recipes to use a narrow delivery slice, an outer elapsed `timeoutMs`/`maxRuntimeMs`, and a mutation-safe checkpoint requested before that limit after the current tool returns. State explicitly that elapsed timeout is not a mutation-safe boundary and may still signal during a tool call.
- Define checkpoint contents: changed files, build/test state, remaining work, and commit or PR state.
- Keep bounded turn/tool caps available for explicitly read-only scouts, reviewers, and validators.
- Record incident `672e31ef`: 4m28s, 23 turns, 42 tools, 20-minute timeout, resumed writer requiring 38+ turns, and the `toolUse` SIGINT/SIGTERM/SIGKILL path.
- Add focused guidance regression tests.

### Ownership boundary

This PR changes README, the packaged pi-subagents skill, the review-loop recipe, and guidance tests only. It adds no guardian code and no runtime termination change.

The separate runtime race is tracked in #482 with focused process-mode regression requirements. Runtime cannot infer whether a worktree is buildable, so fixing that race would not make hard count caps appropriate for writers.

### Verification

- Focused writer-budget guidance tests: 3/3 passed.
- Full unit run: 1,220 passed, 1 skipped, 13 unrelated environment/baseline failures. The new suite passed within that run. Failures were in existing agent settings/profile/skill discovery and one steering timing test; no changed runtime source participates.
- `git diff --check`: passed.

## Review prompt

> Verify this PR changes orchestration guidance and recipes only; does not claim the guardian owns worker lifecycle; forbids hard turn/tool caps for every mutation-capable role; preserves bounded caps for explicitly read-only roles; requires checkpointing only after the current tool returns; records incident 672e31ef accurately; and leaves runtime signal behavior to #482.


nicobailon:
Replaced by #505, which keeps the useful writer-budget guidance and regression coverage without the unrelated Beads, hook, and agent-config files, and adds the runtime safe-boundary fix for #482. Thanks JT for the incident report and initial implementation; the changelog and replacement PR retain your credit.

PR #492, opened 2026-07-16
feat: persist resolved subagent display names

SSS135:
## Summary
- add optional `label` support to single and top-level parallel calls
- resolve deterministic, sanitized child display names with explicit-label precedence and stable fallback deduplication
- persist resolved labels across foreground, async, dynamic fanout, append, nested status, and recovery paths
- expose `workflowGraph` in async summaries and single-run startup details

## Compatibility
- existing calls do not need to add labels; task-derived fallback names are automatic
- legacy status/recovery artifacts without `label` or `workflowGraph` remain readable
- model-facing output and full persisted child output are unchanged

## Validation
- 95 focused unit tests passed
- focused async single integration test passed
- full Windows unit suite: 1245 passed, 11 failed, 13 skipped; failures are existing shared-fixture/agent-management collisions and Windows symlink `EPERM`, unrelated to this diff
- two independent review passes found no remaining blockers


SSS135:
Superseded by #494, which consolidates resolved names and workflow persistence into hierarchical async observability.

PR #493, opened 2026-07-16
feat: expand async widget parallel groups

SSS135:
## Summary
- keep all async chain steps and workflow hierarchy data in the live tracker, including while a parallel group is active
- render named static group parents plus member rows in pending, running, and terminal states
- show dynamic groups as `awaiting targets` until children materialize
- preserve label/role, model, thinking, activity, usage, elapsed, nested, and expanded live details
- bound collapsed groups with an explicit `+N more agents` row

## Dependency
Stacked on draft PR #492 (`feat: persist resolved subagent display names`). The second commit contains the widget-specific diff; this PR can be rebased onto `main` after #492 lands.

## Validation
- widget/render focused unit tests: 24/24 passed
- async job tracker integration: 24/24 passed
- blocker-only independent review: no blockers


SSS135:
Superseded by #494, which consolidates expanded widget groups into hierarchical async observability.

PR #494, opened 2026-07-16
feat: add hierarchical async subagent observability

SSS135:
## Summary

- Adds configurable async widget placement, persisted deterministic display names, and named static and dynamic parallel-group rows.
- Shows the run, logical step, parallel group, agent, and nested-run hierarchy in widget and fleet views.
- Clears or gates transient terminal liveness fields while preserving final metrics, history, and paused-state wording.

## Compatibility

- Rebased onto `origin/main` at `dd24e35`, including the deterministic steering acknowledgment and Windows async hard-kill synchronization fixes from #509 while preserving this PR's six-commit feature scope.
- Keeps `aboveEditor` as the compatible widget-placement default and continues to read legacy artifacts without workflow graphs or resolved labels.
- Retains headless and interactive async guidance, turn-budget deferral, nested routing, and existing foreground and background control semantics.

## Validation

- All #494-specific placement, naming, widget, fleet, nested, hierarchy, dynamic, terminal-state, and schema assertions passed in the focused changed-file suites.
- `test/unit/steering-action.test.ts`: 10/10 passed.
- Targeted Windows async hard-kill regression: 1/1 passed.
- Targeted foreground chain-timeout regression: 1/1 passed; no timing-bound change was needed.
- Broad changed-file runs completed with all feature-specific coverage passing. Remaining local Windows-only failures were an unprivileged symlink setup (`EPERM`) and two async process-start waits; the UTF-8 protocol wait passed on isolated rerun, while the unchanged deferred-turn-budget startup wait remained locally timing-sensitive. No unrelated production or test behavior was added to mask them.
- `git diff --check origin/main...HEAD` passed and the worktree/index were clean before push.

Supersedes drafts #491, #492, #493, and #502.

PR #502, opened 2026-07-17
fix: gate terminal subagent activity state

SSS135:
## Summary
- clear transient activity/tool/path fields when foreground and background children become terminal
- gate live activity rendering by lifecycle state so legacy stale artifacts cannot show inactivity on completed rows
- retain final turns, tools, tokens, duration, output, artifacts, and nested/parallel metrics
- keep top-level activity synchronized with still-running parallel siblings

## Validation
- targeted foreground current-path test: 1/1 passed
- targeted parallel synchronization test: 1/1 passed
- targeted terminal failure cleanup test: 1/1 passed
- focused widget/tracker tests passed during implementation
- `git diff --check` passed

## Stack note
This branch is stacked on draft #493. The terminal-activity changes are the final two commits.


SSS135:
Superseded by #494, which consolidates terminal activity correctness into hierarchical async observability.

PR #503, opened 2026-07-17
fix: diagnose extension bootstrap failures without retrying

SSS135:
## Summary

- Classifies Pi extension-loader startup failures narrowly as `extension-bootstrap-suspected`.
- Preserves bounded stderr head and tail evidence across foreground, background, imported-root, stale-repair, and persisted result paths.
- Marks the classification as non-retryable and prevents fallback-model retries.
- Documents `extensions: []` isolation while retaining the pi-subagents runtime extensions that are required.

## Safety

This does not add an automatic retry. Pi does not expose a positive pre-task startup boundary, so the implementation fails closed to avoid duplicate task execution. Diagnostic persistence remains bounded and retains both head and tail evidence.

## Validation

- Rebased onto `origin/main` at `dd24e35` (including #509 deterministic steering fixes). `git range-diff` reports all three commits unchanged.
- Focused unit coverage for the classifier, child protocol, render helpers, imported chain roots, and stale reconciliation: 31/31 passed.
- Full steering-action unit file: 10/10 passed.
- Foreground bootstrap fail-closed and no-second-call integration test: 1/1 passed.
- Background bootstrap fail-closed and no-retry integration test: 1/1 passed.
- `git diff --check origin/main...HEAD`: passed.
- Fresh GitHub Actions run 29584132969: Ubuntu and Windows passed.


SSS135:
Pausing this for now. The branch, commits, and PR description are being kept so the work can be published again later.

Issue #531, opened 2026-07-18
v0.35.x: default-off watchdog can block startup by recursively hashing modified submodules


4ier:
> I am experiencing this directly. All paths, repository names, submodule names, filenames, session identifiers, and user identifiers have been removed or replaced with generic terms.

## What happened

After upgrading from `pi-subagents` 0.34.x to 0.35.x, Pi startup from a dirty monorepo became extremely slow. The terminal remains unavailable while startup performs sustained synchronous filesystem work.

Sanitized measurements from the same machine and Pi configuration:

| Scenario | Time to ready |
|---|---:|
| `pi-subagents` 0.34.0 in the monorepo | ~1.5 s |
| `pi-subagents` 0.35.0/0.35.1 in the monorepo | 30–95+ s |
| 0.35.1 with extensions disabled | ~0.9 s |
| 0.35.1 from a clean non-repository directory | ~0.8 s |
| 0.35.1 from one submodule directory | ~0.6 s |

The affected superproject had five submodules reported as modified. Their worktrees contained approximately 245,000 files / 4.2 GB in total, mostly ignored dependency and build content. Those names and paths are intentionally omitted.

This is high impact because it blocks every interactive startup in a common monorepo workflow. Users can reasonably interpret the long blank/frozen startup as Pi being broken.

## What I expected

- A watchdog that is disabled by default should add negligible startup work.
- A modified Git submodule should be represented by gitlink/submodule metadata, not by recursively reading and hashing its entire worktree.
- Ignored dependency/build directories should never be traversed for a startup change signature.

## Steps to reproduce

1. Install Pi 0.80.10 and `pi-subagents` 0.35.1 with no watchdog configuration, so the watchdog remains at its default off state.
2. Open a Git superproject containing at least one submodule that is reported as modified.
3. Ensure that submodule contains a large ignored dependency directory such as `node_modules`.
4. Start Pi from the superproject root with `PI_OFFLINE=1 PI_TIMING=1 pi`.
5. Observe that extension initialization blocks startup for tens of seconds.
6. Start Pi from a clean directory, from inside the submodule, or with extensions disabled; startup returns to roughly one second.

The regression can also be isolated around public commit `96d0bd8`: the parent commit starts in about two seconds in the same reproduction, while `96d0bd8` exceeds a 12-second cutoff.

## Additional context

[...]

PR #548, opened 2026-07-20
fix(watchdog): skip git-status baseline on startup when watchdog is disabled

BigSharkLx:
## Description

`resetRepoChangeBaseline()` calls `currentRepoChangeSignature()` unconditionally, even when the watchdog is disabled (which is the default for all users). `currentRepoChangeSignature()` runs:

```
spawnSync("git", ["status", "--porcelain=v1", "-z", "--untracked-files=all"])
```

This synchronous call happens on every `session_start` event — i.e., on every pi startup — via the following call chain:

```
registerMainWatchdog(pi)
  → pi.on("session_start", ...) → runtime.bindSession(ctx)
      → reset("session_start", { resetChangeSignature: true })
          → resetRepoChangeBaseline({ reviewed: true })   // call 1
      → resetRepoChangeBaseline()                         // call 2
          → currentRepoChangeSignature()
              → spawnSync("git status ...")               // ← blocks here
```

On small repositories the cost is negligible (~17 ms). On large repositories — monorepos, NX/Turborepo/Lerna workspaces, or any repo with a large `.git/index` — `git status` takes **1.5–4 seconds**, freezing the pi TUI on every startup even though the watchdog result is never used when the feature is disabled.

The fix is a single guard at the top of `resetRepoChangeBaseline`:

```ts
if (!this.isEnabled()) return;
```

`isEnabled()` returns `configResult.ok && configResult.config.main.enabled`. The watchdog defaults to `enabled: false`, so the expensive git scan is skipped entirely for all users who have not explicitly turned the watchdog on. When the watchdog is enabled the behaviour is completely unchanged.

### Reproduce

1. Install `pi-subagents` in a large git repository (e.g. an NX or Turborepo monorepo where `git status` takes > 1 s).
2. Leave the watchdog at its default disabled state (no `~/.pi/agent/subagents-watchdog.json`).
3. Run `PI_TIMING=1 pi` and observe:
   - `pi-subagents … factory: ~5 ms` — misleadingly fast because PI_TIMING ends before `session_start` fires.
   - TUI ready time: **~7–8 seconds** despite extensions loading in ~1.2 s total.
4. Temporarily replace the factory with a no-op stub — TUI ready drops to **~1.7 s**, confirming the remaining ~6 s comes from inside `registerMainWatchdog`.

### Validate

1. Apply the patch.
2. Restart pi in the same large repository.
3. Observe TUI ready time drops to **~1.9 s** — on par with small repositories.
4. Enable the watchdog (`/subagents-watchdog on`) and confirm it still functions correctly.

## PR Checklist

- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All relevant new and existing tests passed.
- [x] My code follows the code style of this project.
- [x] I ran the full lint checks with no new errors or warnings.
- [x] I checked for other open pull requests for the same change.

## Does This Introduce a Breaking Change?

- No

When the watchdog is disabled (default), `turnStartChangeSignature` stays `undefined` after `bindSession` — which is already the value assigned in the constructor. The watchdog only reads `turnStartChangeSignature` inside paths gated on `isEnabled()`, so no observable behaviour changes for disabled-watchdog users.

## Testing

- **OS:** macOS 15
- **Node.js:** 22
- **Repository under test:** large TypeScript monorepo (~3.4 GB `.git`, ~7 GB `node_modules`, 1 000+ packages)
- **Measurement method:** `expect` script recording time from process spawn to first TUI input prompt

| Scenario | Before | After |
|---|---|---|
| Large monorepo, watchdog disabled (default) | ~7 500 ms | ~1 900 ms |
| Small repo, watchdog disabled (default) | ~2 000 ms | ~2 000 ms (unchanged) |
| Large monorepo, watchdog **enabled** | — | watchdog functions correctly |

`PI_TIMING=1` extension timings (after patch):

```
pi-subagents module import:  217 ms
pi-subagents factory:          5 ms
TOTAL extensions:           1 201 ms
```

## Any Relevant Logs or Outputs

**Before patch** — `PI_TIMING=1` output in large monorepo:

```
--- Startup Timings: extensions ---
  pi-subagents/dist/index.js module import: 217ms
  pi-subagents/dist/index.js factory:         5ms   ← deceptive: session_start fires later
  TOTAL: 1201ms
---
TUI ready: ~7 500 ms   ← 6 300 ms unaccounted by PI_TIMING
```

**After patch** — same repository:

```
--- Startup Timings: extensions ---
  pi-subagents/dist/index.js module import: 217ms
  pi-subagents/dist/index.js factory:         5ms
  TOTAL: 1201ms
---
TUI ready: ~1 900 ms   ✓
```

**Bisection confirming root cause** — replacing `registerMainWatchdog` body with a no-op:

```
TUI ready: ~1 700 ms   ← proves the entire regression lives inside registerMainWatchdog
```

## Other Information or Known Dependencies

- No dependency changes.
- The `constructor` already assigns `this.turnStartChangeSignature = undefined` (a separate earlier fix in my local copy); this PR only adds the `isEnabled()` guard in `resetRepoChangeBaseline`, which is the actual hot path on startup.
- The issue affects all users working in repositories where `git status` is slow, regardless of OS or Node version.

### DIFF
--- a/src/watchdog/runtime.ts
+++ b/src/watchdog/runtime.ts
@@ -536,6 +536,13 @@ export class MainWatchdogRuntime {
 	}
 
 	private resetRepoChangeBaseline(options: { cwd?: string; reviewed?: boolean } = {}): void {
+		// Skip the blocking git-status call when the watchdog is disabled (the default).
+		// computeWatchdogRepoChangeSignature runs `git status --porcelain=v1 -z
+		// --untracked-files=all` synchronously via spawnSync. On large repositories
+		// (monorepos, NX/Turborepo workspaces, etc.) this can take 1.5–4 seconds and
+		// freezes the pi TUI on every startup — even though the result is never used
+		// until the watchdog is explicitly enabled.
+		if (!this.isEnabled()) return;
 		this.turnStartChangeSignature = this.currentRepoChangeSignature(options.cwd ?? this.cwd);
 		if (options.reviewed) this.lastReviewedChangeSignature = this.turnStartChangeSignature?.key;
 		else this.lastReviewedChangeSignature ??= this.turnStartChangeSignature?.key;


nicobailon:
Closing this as superseded by #532, which landed the disabled-watchdog startup scan guard plus nested Git worktree handling for #531. I added changelog credit for this PR’s measurements and validation in #550. Thanks for the detailed repro and timing data.

PR #568, opened 2026-07-21
feat: add searchable selector to subagents-admin

asp345:
## Summary

The `/subagents` admin flow presented its model, thinking, and agent choices through the flat `ctx.ui.select` list, which renders every option without a scroll window or search. With many models or agents the list grows past the viewport and the current selection scrolls off the top, so you cannot see what is selected.

This replaces those pickers with a searchable, bounded-scroll selector that mirrors Pi's built-in `/model` picker: a fixed header, a live search field, and a scroll window that keeps the highlighted row on screen with an `(n/total)` indicator and a `✓` on the current value.

## Changes

- Add `src/slash/selector.ts`: a reusable `SelectorComponent` built from Pi's own TUI primitives (`Container`/`Input`/`Text`/`Spacer`/`DynamicBorder` + `fuzzyFilter`), so it inherits the look and search behavior of the built-in selector.
- Route `chooseModel`, `chooseThinking`, and `selectAgent` (both the "Select subagent" list and the ambiguous-name disambiguation) through a `selectFromList` helper.
- The selector docks in place of the editor (`overlay: false`), matching the native `/model` picker rather than floating a centered overlay.
- Fall back to `ctx.ui.select` when a custom overlay is unavailable (RPC/print modes without a TUI).

## Notes

- Colors come from the `theme` passed to the `ctx.ui.custom` factory, since the module-level `theme` singleton is undefined under the extension's jiti module cache.
- Test suite result is unchanged from `main` (same 1310 pass / 10 pre-existing fail on this environment); no new failures are introduced, and no test imports the changed selector paths beyond the existing slash-command integration coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Replaced the flat `/subagents` admin model, thinking, and agent pickers with a searchable, bounded-scroll selector docked in place of the editor, matching Pi's built-in `/model` picker so the current selection no longer scrolls off screen when the option list is long.
 - Added `advisor` as an `oracle`-compatible bundled agent alias for users switching between Claude Code and Pi naming. Thanks to Serhii Chernenko (@serhii-chernenko) for #552.
 - Show each subagent child’s resolved `[fresh]` or `[fork]` launch context in foreground results, async status, fleet, and widget surfaces, with `[mixed]` on aggregate headers when a run uses both modes.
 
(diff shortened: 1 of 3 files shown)


nicobailon:
Closing this in favor of #603, which kept the searchable `/subagents` admin selector work, rebased it onto current `main`, added the missing changelog credit, and landed as `1913ca5c`.

Validation evidence:
- #603 exact-head Tests run `30071179742` passed on Ubuntu (`89412961168`) and Windows (`89412960607`).
- Post-merge Tests run `30071781848` passed on Ubuntu (`89413982618`) and Windows (`89413982599`).

Thanks again to Chanyeong Lim (@asp345) for the original picker improvement.

PR #569, opened 2026-07-21
fix: chain outputs now leave the building

### LINKED ISSUE #529: Chain output files written to chainDir, never persisted to CWD-relative output paths
## Summary

When a chain step specifies `output: "path/to/file.md"` with `outputMode: "file-only"`, the agent writes to `{chainDir}/path/to/file.md` (a temp directory). The chain framework does not copy the file back to the CWD-relative path. This means:

1. Files at the CWD path are stale from prior runs — never overwritten
2. Fresh output is only available in the chain temp directory
3. Consumers reading from the CWD path get outdated data

## Reproduction

Chain config:
```json
{
  "chain": [
    {
      "parallel": [
        {
          "agent": "reporter",
          "output": "reports/report.md",
          "outputMode": "file-only"
        }
      ]
    }
  ]
}
```

After the chain completes, `reports/report.md` does not exist (or is stale from a previous run). The actual output is at `{chainDir}/reports/report.md`.

## Expected behavior

Either:
1. Copy output files from `chainDir` to CWD-relative paths after chain completion, or
2. Document that `output` paths are chainDir-relative and provide a mechanism to persist them

## Environment

- pi-subagents: v0.34.0
- Pi: 0.80.6

## Related

This is distinct from #420 (async parallel overwrite) — that covers collision within a single run. This issue is about the framework never persisting chain output to the configured path.

dougEfresh:
# Fixes #529: chain outputs now actually leave the building

Turns out writing a file to `{chainDir}/reports/report.md` and calling it "done" doesn't mean the poor thing ever makes it to the CWD path you actually asked for. It just sits in a temp dir like a package left on the porch, while the destination file rots stale from three runs ago. Nobody copied it. Nobody was ever going to.

So now, once a chain finishes clean, we walk every relative `output` path collected along the way (sequential steps, parallel groups, dynamic fanout — didn't play favorites) and copy each one from `chainDir` back to CWD, making directories as needed. Absolute paths are left alone, because they already know where they live. If a source file never got written for whatever reason, we shrug and move on — a missing copy shouldn't take down a chain that otherwise did its job.

Added three tests to keep this honest: the happy path persist, the "don't touch absolute paths" guard, and the "a file went missing, don't panic" case. All green, plus the full existing suite.

### DIFF
--- a/src/runs/foreground/chain-execution.ts
+++ b/src/runs/foreground/chain-execution.ts
@@ -666,6 +666,7 @@ export async function executeChain(params: ChainExecutionParams): Promise<ChainE
 	let prev = "";
 	let globalTaskIndex = 0;
 	let progressCreated = false;
+	const collectedOutputPaths: string[] = [];
 
 	for (let stepIndex = 0; stepIndex < chainSteps.length; stepIndex++) {
 		const step = chainSteps[stepIndex]!;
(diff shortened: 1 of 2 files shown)


nicobailon:
Closing this because the proposed behavior conflicts with the accepted chain-output contract from #529/#564. Relative chain outputs are documented to live under `{chain_dir}`; callers that need stable external paths should use a persistent `chainDir` or an absolute `output`.

The patch copies foreground chain outputs into the caller CWD only, leaves async chains on the existing artifact-base behavior, and silently ignores copy failures. That would make foreground/background chain output semantics diverge and weaken error reporting.

Thanks again for raising the persistence issue in #529; that contribution is already credited in the changelog via #564.

PR #587, opened 2026-07-23
feat: add persistent subagent FleetView

Zeppelinpp:
## Summary

- Add a persistent below-editor FleetView that shows `main` and every active child with its task, elapsed time, and token usage.
- Track foreground parallel children independently, preserving stable rows, lifecycle state, metrics, and transcript selection for concurrent reviewers.
- Upgrade the inspector with structured Markdown, code, and tool rendering, bounded compact/expanded results, and `Shift+J` / `Shift+K` line scrolling.
- Keep editor input safe: navigation only activates when the focused editor is empty, and the inspector remains read-only.
- Suspend the below-editor FleetView while the inspector is open so live status redraws cannot leak repeated frames into terminal scrollback.
- Replace the legacy async widget by default while retaining explicit `fleetView` and `asyncWidget` configuration.

## Value

Fleet activity stays visible without interrupting the editing flow. Independent child state prevents parallel runs from collapsing into one misleading row, while structured transcripts make progress and failures faster to inspect.

## Preview

Subagent FleetView inspector

## Validation

- Unit: 1,362 passed, 1 skipped
- Integration: 634 passed
- E2E: 2 passed
- `npm pack --dry-run`
- `git diff --check`


nicobailon:
Superseded by #606, which carries Rui Pu’s FleetView work onto current `main` while preserving the original authored commits and changelog credit. The replacement merged as `4b1df555c6466e1e319604febe732f595f814de7`; its reviewed tree matched the merge, and exact-merge-SHA CI passed on Ubuntu and Windows. Thanks again to @Zeppelinpp for #587.

PR #595, opened 2026-07-23
fix: map parallel live progress by child index

white-hat:
## Problem 
UI shows duplicate output for parallel `/chain` agents

duplicate output

## Fix
- match indexed parallel progress only to its matching child placeholder
- preserve agent-name matching for unindexed updates
- add regression coverage for sparse parallel progress updates
fixed output

## Test plan
- `node --experimental-strip-types --test test/integration/slash-live-state.test.ts`


nicobailon:
Thanks for the fix, Eli. I carried this forward in #602 so it could run on the maintainer branch with the required changelog credit. #602 was squash-merged as ab237ecc784f383c93286a13bc1938854cb8995e after exact-head Tests passed on Ubuntu and Windows, and the post-merge Tests run 30070512447 also passed on both platforms.

PR #610, opened 2026-07-24
feat: add owned delegation protocol v2

neumie:
## Summary
- add a public delegation v2 contract for workflow supervisors over the existing prompt-template event family
- preserve delegation v1 and the model-facing one-foreground-call guard while routing owned leaves through the concurrent delegated executor
- make attempt identity, cancellation, result mode, accounting, and boundary failures explicit and fail-closed

## Changes
- exports protocol-v2 request, started, update, response, cancellation, usage, thinking, and typed-result contracts from `pi-subagents/delegation`
- keys attempts by `(requestId, ownerRunId, nodeId)` and active logical nodes by `(ownerRunId, nodeId)`
- supports exact cancellation and cancel-before-start, `duplicate_node`, literal text vs schema-validated structured output, effective model/thinking, and detailed usage
- applies strict field allowlists, safe plain-JSON cloning, UTF-8 payload bounds, a safe timer ceiling, and fail-closed bounded cancellation/settled identity history
- adds direct bridge tests plus registered-extension v1/v2 integration coverage and documents authority/security limits

## Test plan
- [x] `node --experimental-strip-types --test test/unit/delegation-api.test.ts test/unit/package-manifest.test.ts test/unit/prompt-template-bridge.test.ts` (40/40)
- [x] registered extension v1 and v2 delegation integration tests (2/2)
- [x] `npm run test:e2e` (2/2)
- [x] `npm pack --dry-run --ignore-scripts`
- [x] edited-file TypeScript diagnostics and `git diff --check`

## Baseline note
The broader upstream suite has existing timing-sensitive timeout/watchdog failures. The same chain/parallel timeout failures reproduced from an exact upstream-base worktree; the delegation-focused, registered-extension, package, and E2E gates above are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Added delegation v2 for extension-owned concurrent foreground leaves, with logical run/node ownership, exact per-attempt cancellation, explicit duplicate-node outcomes, literal or structured values, effective model/thinking metadata, and detailed usage while preserving delegation v1 and the model-facing single-dispatch guard.
 - Added a persistent below-editor FleetView with safe empty-editor navigation and a structured inspector for Markdown, code, tool calls, and compact or expanded tool results. Thanks to Rui Pu (@Zeppelinpp) for #587.
 - Added `artifactDir` config to store subagent artifacts in the project, Pi session, or temp artifact directory while keeping project-local artifacts as the default. Thanks to WeZZard (@WeZZard) for #582.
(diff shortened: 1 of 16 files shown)


neumie:
The delegation-focused gates and the full Ubuntu/Windows unit, integration, and E2E matrix are green. This public seam is the provider prerequisite for pi-subagents-workflows; its final consumer dependency will wait for an upstream RC/release rather than using a Git or sibling dependency. Happy to adjust the contract or split the change if helpful.

nicobailon:
Superseded by replacement PR #614, which preserved Jakub Neumann's delegation v2 work, scoped the zero-tool budget marker to v2 delegated children, passed Ubuntu/Windows CI, and merged as c4080f28fe31cce73589ad76993651d006a341ab. Thanks again to Jakub Neumann (@neumie) for #610.

Issue #122, opened 2026-04-28
`This extension ctx is stale after session replacement or reload` crash in 0.70 after using /new


Thinkscape:
## Steps

1. Have pi with pi-subagents enabled.
2. Work in a session, ask to spawn subagents
3. [After work is complete]
4. Type `/new` to start new session
 
## Expected

New session starts

## Actual

pi crashes back to shell with:
```
Error: This extension ctx is stale after session replacement or reload. Do not use a captured pi or command ctx after ctx.newSession(), ctx.fork(), ctx.switchSession(), or ctx.reload(). For newSession, fork, and switchSession, move post-replacement work into withSession and use the ctx passed to withSession. For reload, do not use the old ctx after await ctx.reload().
    at ExtensionRunner.assertActive (file:///Users/thinkscape/node_modules/@mariozechner/pi-coding-agent/dist/core/extensions/runner.js:296:19)
    at get ui (file:///Users/thinkscape/node_modules/@mariozechner/pi-coding-agent/dist/core/extensions/runner.js:381:24)
    at Object.invalidate (/opt/homebrew/lib/node_modules/pi-subagents/index.ts:302:89)
    at Timeout._onTimeout (/opt/homebrew/lib/node_modules/pi-subagents/render.ts:121:43)
    at listOnTimeout (node:internal/timers:605:17)
    at process.processTimers (node:internal/timers:541:7)
```

nicobailon:
Fixed in v0.21.0. Animation timers now stop when Pi reports a stale extension context after /new or reload, instead of reusing the old ctx.ui.

PR #657, opened 2026-07-27
fix: stop FleetView refresh on stale context

kylegl:
## Problem

FleetView keeps a refresh interval while subagents are active. After a session replacement or reload, Pi invalidates the previous extension context, but an already queued refresh can still access `ctx.hasUI` and crash Pi with an uncaught stale-context error.

This is the same lifecycle failure previously addressed in #122 for legacy animation timers, reintroduced by the persistent FleetView.

## Fix

Stop FleetView activity when Pi reports that its captured extension context is stale:

- retain the UI reference separately so cleanup does not dereference the stale context
- stop the refresh timer and release terminal-input registration
- discard stale context and widget references
- continue throwing unrelated context errors, aggregating them when multiple cleanup operations fail

Regression coverage verifies the current Pi stale-context error from `SubagentFleetStatus.refresh`, confirms stale registrations are released, and ensures unrelated cleanup errors remain visible.

### DIFF
--- a/src/tui/fleet-status.ts
+++ b/src/tui/fleet-status.ts
@@ -48,6 +48,13 @@ function isActiveState(value: string): boolean {
 	return value === "running" || value === "queued" || value === "pending";
 }
 
+function isStaleExtensionContextError(error: unknown): boolean {
+	// Pi currently exposes stale contexts as plain Errors without a stable code or subtype.
+	return error instanceof Error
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the report and patch. I landed a maintainer-side version on main as 5ccefc140, including the stale-context cleanup and changelog credit for #657.

PR #661, opened 2026-07-27
fix(fleet): invalidate live overlay before refresh

shaneconner:
## Summary
- invalidate the live Fleet inspector component before each timer-driven render
- keep roster/transcript refresh behavior unchanged
- cover timer invalidation in the existing live-refresh unit test

## Why
Pi TUI components must invalidate cached frames before `requestRender()` after state changes. Fleet refreshed its snapshot directly, so a live overlay could leave repeated stale structured-header frames (task/stats lines) while a single child continued normally. Persisted child execution/transcripts were unaffected; this is a package UI fix.

## Validation
- `node --experimental-strip-types --test test/unit/fleet.test.ts`
- 11/11 passed

### DIFF
--- a/src/tui/fleet.ts
+++ b/src/tui/fleet.ts
@@ -463,7 +463,7 @@ export class SubagentFleetComponent implements Component {
 		this.refresh();
 		this.timer = setInterval(() => {
 			if (this.disposed) return;
-			this.refresh();
+			this.invalidate();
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the fix. I landed a maintainer-side version on main as 5d89b86f1, including the timer invalidation change and changelog credit for #661.

PR #670, opened 2026-07-28
fix: tolerate steering inbox scan failures

hughcars:
## Summary

`consumeSteerRequestsFromDir()` did not guard its initial `readdirSync()`. A scan error such as `ENFILE` could therefore escape the steering timer and terminate the child process.

Treat a failed scan as temporarily unavailable, leaving request files untouched for the next poll. This matches the existing acknowledgement consumer behavior.

## Validation

- confirmed the regression test fails before the production change with `ENFILE` from `readdirSync()`
- `node --experimental-strip-types --test test/unit/control-channel.test.ts` — 23/23 passing
- `npm run test:unit` — 1,464 passing, 1 skipped


nicobailon:
Thanks @hughcars — I landed this via maintainer replacement PR #674 so the branch could pick up the current base and visible changelog credit. The production fix and regression test are preserved, with credit in CHANGELOG.md.

nicobailon:
Closing as superseded by #674, which has been merged.

PR #671, opened 2026-07-28
fix: retry zero-activity subagent startup exits

felipeteodorocw:
## What changed

- classify only short, non-zero child exits with no model messages, tool activity, usage, output, diagnostics, process signal, mutation evidence, or lifecycle interruption as startup failures
- retry those failures on the same model with bounded 250 ms / 750 ms / 1500 ms backoff before considering ordinary model fallback
- apply the same policy to foreground and background execution, with cancellation-aware waits and per-launch `modelAttempts`
- retain bounded raw stdout diagnostics and foreground process signals so diagnosed or signaled exits are never replayed
- synthesize actionable exhaustion errors and `[startup-retry]` notes without including task or environment contents
- add unit, foreground integration, and background integration coverage

## Why

Concurrent child Pi launches can occasionally exit before model invocation. These runs currently report exit code 1 with zero turns, tokens, tools, or assistant output, and no error, so existing provider/model fallback does not apply. Serializing all subagents avoids the race but removes useful parallelism.

The retry gate intentionally fails closed to avoid duplicate side effects: any evidence that model or tool work started prevents a startup retry.

## How it was validated

- `npm run test:unit` — 1469 passed, 1 platform skip
- focused foreground startup-retry integration cases — 6 passed
- standalone background-runner validation with the repository mock Pi — passed (same-model retry, metadata, notes, and successful second launch)
- `git diff --check` — passed before commit
- three read-only reviewer passes, including a final fresh-context review — no blocking findings

The full integration file is not reliable in this local Node test-runner environment because the harness sends `SIGKILL` to mock child processes when the complete file runs; the focused foreground cases and standalone background validation avoid that harness limitation.

## Security / residual risk

- retries are capped at three retries after the initial launch
- process signals, stderr, bounded raw stdout, protocol errors, messages, tools, usage, mutation evidence, timeout, stop, detach, and interrupt state all disable startup retry
- retry notes contain only the model id, attempt count, and delay
- `npm audit --audit-level=high` reports the existing transitive `brace-expansion` high-severity advisory under `@earendil-works/pi-coding-agent`; this PR does not change dependencies or the lockfile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Retried short, zero-activity child startup exits on the same model with bounded backoff, reducing concurrent subagent launch races without replaying model or tool work.
+
 ## [0.37.2] - 2026-07-28
 
 ### Changed
--- a/src/runs/background/subagent-runner.ts
+++ b/src/runs/background/subagent-runner.ts
@@ -65,6 +65,13 @@ import { readChildToolDiagnosticError } from "../shared/tool-availability.ts";
 import { collectDynamicResults, DynamicFanoutError, materializeDynamicParallelStep, validateDynamicCollection } from "../shared/dynamic-fanout.ts";
 import { nestedSummaryFromAsyncStatus, projectNestedEvents, resolveNestedAsyncDir, writeNestedEvent } from "../shared/nested-events.ts";
 import { formatModelAttemptNote, isRetryableModelFailure } from "../shared/model-fallback.ts";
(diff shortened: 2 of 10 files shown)


nicobailon:
Thanks @felipeteodorocw — I landed this via maintainer replacement PR #676 after rebasing it onto the current queue and resolving the CHANGELOG.md conflict. The startup retry implementation and tests are preserved, with credit in CHANGELOG.md.

nicobailon:
Closing as superseded by #676, which has been merged.

PR #673, opened 2026-07-29
fix: isolate intercom stable IDs for child sessions

nicobailon:
## Summary

- use pi-intercom's current runtime session ID when building unnamed supervisor fallback targets
- prevent parent `PI_INTERCOM_STABLE_ID` / `PI_INTERCOM_SESSION_ID` values from leaking into child Pi processes
- assign each child its own stable intercom ID when pi-subagents already generated a child intercom target

## Validation

- `npm run test:unit` — 1465 passed, 1 skipped
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/intercom-result-delivery.test.ts` — 31/31 passed
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Kept pi-intercom stable IDs from leaking into child sessions and used the current intercom runtime ID for unnamed supervisor targets.
+
 ## [0.37.2] - 2026-07-28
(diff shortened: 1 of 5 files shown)


nicobailon:
Reviewed the intercom stable ID isolation change. The env clearing behavior is covered and Node omits undefined env entries at spawn time; focused tests and existing CI are passing. No issues found.

PR #734, opened 2026-08-03
fix: fall back to pid-scoped temp dirs on Windows EPERM/EACCES at load

franktheglock:
## Summary

This addresses the other Windows EPERM failure the rename fix (#731) didn't cover: a **`mkdir` EPERM at extension load**. On Windows, when the shared temp async result/run directories are persistently blocked by `EPERM`/`EACCES` (typically corrupted NTFS ACLs after wake-from-sleep on Azure AD/Entra machines), `ensureAccessibleDir` exhausts its retry loop and throws out of the pi extension loader — killing the whole Pi process before the extension can register.

```
Error: Failed to load extension ... EPERM: operation not permitted, mkdir '...Temp\pi-subagents-user-...\async-subagent-results'
```

## Changes

- **`ensureAccessibleDir`** now returns the usable path: it recreates the directory in place, and if a persistent `EPERM`/`EACCES` still blocks it, falls back to a fresh **pid-scoped sibling path** (`<dir>-<pid>`).
- Added a **mutable `DIRS` container** (`results`/`async`/`chain`/`artifacts`) so a runtime fallback path propagates to all ES-module consumers despite immutable `const` bindings.
- Migrated every `RESULTS_DIR`/`ASYNC_DIR` consumer to `DIRS.results`/`DIRS.async`.
- **Unit tests** (`test/unit/ensure-accessible-dir.test.ts`) cover the pid-scoped fallback on mkdir and access EPERM, plus `DIRS` reassignment.

Ports the approach from upstream PR #232 (which was written for this exact condition but never merged).

## Verification
- 1542 unit tests pass, 0 fail.
- `node --experimental-strip-types --check` passes on all modified files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 - Routed foreground chain scratch files to user-scoped temp storage when `artifactDir` is `"session"` or `"temp"`, preventing `.pi-subagents/` clutter in the working directory. Thanks to @magoz for #729.
 - Retry transient Windows `EPERM`/`EBUSY`/`EACCES` locks when atomically replacing the async runner startup-control handshake file, so a transient antivirus/indexer lock no longer makes the parent believe the runner never reached the ready state. Thanks to @franktheglock for #731.
 - Kept successful background subagent completions quiet so inactive Pi tabs are not marked unread, while failed and paused completions still notify the originating session. Thanks to @killianMei for #728.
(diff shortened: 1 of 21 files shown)


nicobailon:
Thanks for chasing this down. I landed a replacement in #736 that keeps the pid-scoped fallback approach from this PR, but avoids deleting the existing async/results temp roots when the primary path is blocked.

Your contribution is credited in the changelog as `@franktheglock for #734`, and the replacement merge commit passed post-merge CI on Ubuntu and Windows: abf07371d4470110ad21f701d896145e280a6faa.

Closing this one as superseded by #736.

PR #794, opened 2026-08-04
Export control-channel, intercom-bridge, pi-args, and shared types for extension consumers

shaneconner:
## Motivation

Extension packages that compose with pi-subagents (delegating through the prompt-template bridge, driving async runs via the control channel, resolving launch tool plans, or wrapping fork tasks) currently have to deep-import these modules by relative path, because the export map doesn't cover them:

- `src/runs/background/control-channel.ts` (`requestAsyncStop`)
- `src/intercom/intercom-bridge.ts` (`resolveIntercomSessionTarget`)
- `src/runs/shared/pi-args.ts` (`resolvePiLaunchToolPlan`)
- `src/shared/types.ts` (`wrapForkTask`, shared type definitions)

Deep imports break silently on any internal refactor. This PR adds the four subpaths to `exports` (mirroring the existing entry style) so downstream consumers can depend on a supported surface instead. Evidence of real demand: a production extension of ours uses exactly these four seams today via relative paths.

## Changes

- `package.json`: four additive export-map entries; no source changes, no version bump.
- `test/unit/package-manifest.test.ts`: expectation updated for the new entries (8/8 passing).

Happy to rename the subpaths, narrow `shared-types` to a curated re-export, or split this differently if you'd prefer a more deliberate public surface — the goal is just that composing extensions have something supported to import.


shaneconner:
Rebased the branch to tidy commit messages — your "fix: narrow extension entrypoint exports" is preserved verbatim (tree-identical, now 0daf9a8). No content changes.

PR #807, opened 2026-08-05
fix: validate acceptance and settle workflow siblings

ryanbbrown:
## Summary

- reject scalar `acceptance: "verified"` and commandless verified policies before a child starts
- make `runs.all` validate the full batch before dispatch, then return every child outcome in input order
- keep `runs.run` fail-fast and keep workflow timeout and cancellation behavior unchanged

## Why

A verified policy without runtime commands previously ran the child and failed only during acceptance evaluation. Scalar verified acceptance could never succeed because the scalar form cannot supply commands.

`runs.all` also used fail-fast promise behavior. One failed child rejected the workflow and aborted the shared child controller, which could terminate unrelated siblings. `runs.all` now returns `ok: false` results, including launch errors, so the workflow script decides what is fatal.

## Validation

- focused unit tests: 76 passed
- focused single-execution integration tests: 2 passed
- remaining unit suite, excluding the unchanged Herdr inspector baseline failure: 1,569 passed, 1 skipped
- remaining integration files: 508 passed
- real Pi-session E2E tests: 3 passed
- `npm run typecheck`
- `git diff --check`

The full unfiltered suites have two existing local failures that reproduce on `upstream/main`: six Herdr inspector tests are cancelled because the fake timeout leaves the event loop empty, and one managed-worktree handoff test intermittently sees one of two handoffs under suite-wide load but passes when run alone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@
 - Scheduled subagent runs are now enabled by default; set `{ "scheduledRuns": { "enabled": false } }` to opt out.
 
 ### Fixed
+- Reject scalar or commandless verified acceptance before spawning a child; verified policies now require object form with at least one runtime command.
+- Let every `runs.all` child settle and return ordered per-child outcomes instead of aborting siblings when one child fails; `runs.run` remains fail-fast.
 - Preserve successful async completion when project-local artifact or mission files are removed before final bookkeeping by recreating artifact directories and recording missing-mission warnings.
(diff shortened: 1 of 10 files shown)


nicobailon:
Thanks @ryanbbrown. I landed this via replacement PR #818 because this branch was fork-backed and could not be updated in place from the maintainer side after the base moved. Your implementation was preserved in #818, with visible changelog credit for @ryanbbrown and #807. The replacement passed fresh review, Greptile, Ubuntu, and Windows checks, then merged.

PR #822, opened 2026-08-05
fix(tui): unify subagent status presentation

xz-dev:
## Summary

- centralize subagent lifecycle presentation so compact and expanded rows agree
- render detached, stopped, and interrupted results as terminal states even when progress metadata still says `running`
- replace terminal-dependent circle selection glyphs with a simple focus marker

## Problem

Several expanded result paths rebuilt status labels independently. That allowed stopped or interrupted results to appear as `failed`, `done`, or still `running`, while compact rendering used different glyph semantics. The persistent Fleet selector also used `⏺`/`◯`, whose appearance varies substantially between terminal fonts.

## Validation

- focused render/Fleet tests: 47 passed
- integration suite: 681 passed
- E2E suite: 3 passed
- `npm run typecheck`
- `git diff --check`

The full unit command still has the same 12 environment/project-discovery failures as unmodified `main`; focused changed-surface tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Scheduled subagent runs are now enabled by default; set `{ "scheduledRuns": { "enabled": false } }` to opt out.
 
 ### Fixed
+- Unified subagent status presentation across compact, expanded, and Fleet views, including terminal interruption precedence and terminal-safe selection markers. Thanks to @xz-dev for #822.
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks @xz-dev. I landed this status-presentation work via replacement PR #829 so it could be integrated with the current renderer, FleetView, and terminal-state fixes.

The status semantics from this PR were preserved in #829, with visible changelog credit for @xz-dev and #822. The replacement passed fresh review, Greptile, Ubuntu, and Windows checks, then merged.

PR #823, opened 2026-08-05
fix(tui): sanitize Fleet transcript display

xz-dev:
## Summary

- add a safe display boundary for persisted and directly rendered Fleet transcript events
- escape terminal controls, bidi controls, private-use code points, lone surrogates, and Unicode noncharacters
- replace clearly binary content with a safe placeholder
- sanitize transcript warnings before other Fleet views consume them
- preserve normal Unicode, Markdown, code highlighting, newlines, tabs, and terminal width bounds

## Scope

The patch is limited to Fleet transcript display and adds no dependency. Binary detection is deliberately conservative: obvious binary content is omitted; other unsafe characters are rendered as visible `[U+XXXX]` markers.

## Validation

- focused Fleet transcript tests: 14 passed
- integration suite: 678 passed
- E2E suite: 3 passed
- `npm run typecheck`
- `git diff --check`

The full unit command still has the same 12 environment/project-discovery failures as unmodified `main`; all changed-surface tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Scheduled subagent runs are now enabled by default; set `{ "scheduledRuns": { "enabled": false } }` to opt out.
 
 ### Fixed
+- Sanitize Fleet transcript content and warnings before terminal display while preserving normal Unicode and formatting. Thanks to @xz-dev for #823.
 - Keep the under-editor async status widget visible by default while FleetView is enabled, including after active runs restore following reload. Thanks to @nicobailon for #804.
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks @xz-dev. I landed this via replacement PR #827 so it could be rebased cleanly with the current Fleet transcript code and the Windows-safe test fixes.

The transcript sanitization behavior was preserved in #827, with visible changelog credit for @xz-dev and #823. The replacement passed fresh review, Greptile, Ubuntu, and Windows checks, then merged.

PR #824, opened 2026-08-05
fix(tui): preserve actionable tool errors

xz-dev:
## Summary

- forward Pi's authoritative `context.isError` flag into the custom subagent result renderer
- keep multi-line actionable tool errors visible instead of applying the normal collapsed success summary
- preserve the existing collapsed behavior for successful structured results

## Root cause

Pi intentionally provides error state through the custom renderer context. The result passed to `renderResult` does not retain that flag, so `renderSubagentResult` could mistake an error for a normal result and collapse it.

## Validation

- registered-tool boundary regression: 15 passed
- integration suite: 678 passed
- E2E suite: 3 passed
- `npm run typecheck`
- `git diff --check`

The full unit command still has the same 12 environment/project-discovery failures as unmodified `main`; the changed registration test passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Scheduled subagent runs are now enabled by default; set `{ "scheduledRuns": { "enabled": false } }` to opt out.
 
 ### Fixed
+- Preserve actionable multi-line subagent tool errors in collapsed result rendering. Thanks to @xz-dev for #824.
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks @xz-dev. I landed this via replacement PR #830 because the original branch conflicted with the current result-renderer code after the status and summary-renderer changes landed.

The behavior from this PR was preserved in #830, with visible changelog credit for @xz-dev and #824. The replacement passed fresh review, Greptile, Ubuntu, and Windows checks, then merged.

PR #890, opened 2026-08-07
feat: expose scheduledRuns.storeRoot to keep schedules out of project…

ProCleiton:
… repos

schedules are persisted per project under <cwd>/.pi-subagents/schedules, which leaves runtime state inside project repositories. The schedule store authority (scheduledRunStorePath with a root, ScheduledRunManager deps.storeRoot) already supported an external store but it was not reachable from config.

Add an optional scheduledRuns.storeRoot to the extension config. When set, each project's schedules live under <storeRoot>/<sha256(cwd).slice(0,20)>, so multiple projects never collide and nothing pollutes the repo.

- config.ts: resolveScheduledStoreRoot expands ~/ via os.homedir and rejects project-relative paths (a global store must not move with a transient worktree); validateScheduledRunsConfig validates the option.
- index.ts: expand storeRoot and thread it into createScheduledRunManager.
- Docs + unit tests for expansion, rejection, config round-trip, and external-root persistence.

Backward compatible: without storeRoot the behavior is identical to before.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -116,6 +116,14 @@ A user may explicitly call `subagent({ action: "grant-spawn-budget", additional:
 
 Durable schedules are enabled by default and stored per project under `.pi-subagents/schedules/<id>/`. See [missions.md](missions.md#schedules) for usage.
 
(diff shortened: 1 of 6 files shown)


ProCleiton:
Sorry, PR sent to the wrong repository. I meant to send it to the fork before proposing this.

PR #902, opened 2026-08-08
fix: cap portable Codex tool IDs

alexei-led:
Fixes Codex requests rejected when a composite tool id is base64url-encoded past the provider 64-character limit.

- preserve existing portable IDs at or below 64 characters
- preserve existing base64url mapping when it fits
- deterministically hash oversized IDs while retaining call/result linkage
- add regression coverage for the observed composite id


alexei-led:
Closed at the requester preference. The error is reported in #903 for maintainer review.

PR #904, opened 2026-08-08
fix: cap oversized composite tool IDs in child context

alexei-led:
## Summary

- cap inherited tool-call IDs at OpenAI's 64-character limit
- preserve short composite IDs for APIs that normalize them
- deterministically hash oversized IDs while keeping assistant calls and tool results correlated

## Root cause

`pi-subagents` skipped tool-ID sanitization for `openai-codex-responses`. Generated composite IDs such as `call_*|fc_*` can be 116 characters. On a resumed/detached child turn, OpenAI rejects the raw ID before the model runs:

```
Invalid 'input[2].call_id': string too long. Expected a string with maximum length 64, but got a string with length 116 instead.
```

## Verification

- `HOME=$(mktemp -d) npm run typecheck` — pass
- `HOME=$(mktemp -d) npm run test:all` — pass
- replayed a copy of a real failed parallel-tool session — `SESSION_REPLAY_OK`
- replayed a copy of a real failed supervisor/reviewer session — `SUPERVISOR_REPLAY_OK`

Original session files were not modified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Serialize and merge each mission workflow-state write with the latest file so separate workflows do not drop unrelated keys.
 - Preserve `workflowScript` worktree children that detach for supervisor coordination instead of cleaning a live managed worktree. Thanks to @astarktc for #896.
 - Accept schema-valid structured output after a child recovers from an earlier tool error. Thanks to @white-hat for the report in #888.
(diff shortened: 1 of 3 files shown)


alexei-led:
Closing this: the root cause was localized to my pi-sub-aliases extension, so an upstream pi-subagents change is not needed.

PR #915, opened 2026-08-08
feat: surface terminal completions in subagent_wait details

lucasgrecco:
## Problem

Async completions reach the parent only as text. The result file (`<runId>.json`) already carries a structured `results` array — per-child agent, success, and `artifactPaths` — but the watcher consumes and deletes it after delivering the text notice, and the management-path tool results hardcode `details: { mode: "management", results: [] }` (`subagent-wait.ts:310` and the equivalents in `subagent-executor.ts`). `SingleResult.index` is documented as "Pair with `Details.runId` for cross-run correlation" — the results array is the correlation surface by design — yet in async mode nothing structured ever reaches it.

Concrete case, observed in the field: an extension that records a launch ledger from `tool_result` events (to correlate runs with their on-disk artifacts) got nothing for `async: true` + `subagent_wait`, while the sync workflow path populates `details.results` from child results (`subagent-executor.ts`) — the same traversal, mode-dependent observability.

## Change

`subagent_wait` now returns a slim projection of each terminal payload it covered in `details.completions`:

```ts
interface WaitCompletion {
  runId: string;
  agent?: string;
  mode?: string;
  state?: string;
  success?: boolean;
  results?: Array<{ agent?; runId?; success?; outputState?; error?; model?; artifactPaths? }>;
}
```

- Output text is deliberately excluded — it already travels in the tool-result content; details carry identity and the artifact trail only.
- The result watcher records the payload (new optional `SubagentState.completedResults` map, pruned with the same completion TTL that dedupes notifications) before dedupe and unlink; `subagent_wait` reads that record first and falls back to a direct, read-only result-file read for the pre-consumption window. Result files are written atomically, so the direct read never observes a torn write, and the wait never deletes — the watcher keeps ownership of notification and cleanup.
- **Workflow result files now record each child's `runId`** (`subagent-executor.ts`, success and failure paths). The workflow engine already knows it (`WorkflowScriptChildResult.runId`); the result-file mapping dropped it, leaving identity to be guessed from paths — and a workflow child's `artifactPaths` collects whatever paths the run produced (saved output under `outputs/<runId>/…`, session file, async dir), so which one lands first is run-shape-dependent and parsing ids out of path segments is contract-free guesswork. The engine knows the id; carrying it explicitly ends the guessing.
- `details.results` stays `SingleResult[]` and untouched: result-file children are looser than `SingleResult` (no `task`/`usage`/`exitCode`), so fabricating those required fields to reuse the existing array seemed worse than a new optional field. Happy to reshape if you'd rather extend `results` instead.

## Testing

- `npm run typecheck` clean.
- `npm run test:unit`: 1664 pass / 0 fail / 1 pre-existing skip. Two new tests in `test/unit/subagent-wait.test.ts`: the file-source test asserts per-child `runId` and `artifactPaths` come through, output text stays out of details, and the wait leaves the result file in place for the watcher; the record-source test covers the path where the watcher already consumed and deleted the file.
- Field-validated against a real consumer (a coding-pipeline extension that keeps a launch ledger keyed by run id): across two live pipeline runs — one stopped early, one multi-phase — all 9 async workflow children observed at the time of writing were correlated to their artifacts via `details.completions`, including failed children and a fail→relaunch→success pair. Before the change the same consumer correlated zero.
- CHANGELOG entry under Unreleased; `docs/observability.md` updated where the result-file lifecycle is documented.


lucasgrecco:
Thanks for rebasing the branch and for the changelog credit.

On the review — both P1s look valid to me, and I plan to fix them on this branch:

1. **Terminal-status → result-file race**: the executor persists terminal status before writing the result file, so a one-shot lookup can genuinely land in that window and return no completion. I'll add a short bounded re-check of both sources (watcher record, then file) with an injectable sleep so the unit tests stay deterministic — a few attempts within ~1s total.

2. **Broad catch in `collectWaitCompletions`** — one question on preferred behavior for non-ENOENT read errors (malformed JSON, I/O errors) before I push:

   - The literal suggestion rethrows anything that isn't ENOENT. That fails the whole `subagent_wait` call over a corrupt result file, even though the text summary in the tool-result content is still valid and `details.completions` is an additive projection.
   - The alternative narrows the catch: ENOENT → re-check the watcher record (the real race); anything else → `console.error` with the path and continue without that run's completion — matching how `result-watcher.ts` handles malformed payloads today.

   I lean toward log-and-continue, since failing the wait would degrade existing behavior to protect a new optional field — but happy to implement the rethrow if you prefer strictness here. Which would you rather have?

The P2 on changelog attribution looks already covered by the line you added.

Issue #933, opened 2026-08-09
Steering request can remain unacknowledged after manual-compaction rejection


jtac:
## Summary

Independent review of #928 found a separate pre-existing loss/acknowledgement gap when the steering inbox calls `sendUserMessage` during **manual compaction**.

Current Pi rejects `sendUserMessage` before emitting the correlated extension `input` event while the manual-compaction controller is active. The ExtensionAPI is void and logs/swallow-catches the asynchronous rejection. The steering inbox has already consumed the request file, but because no correlated `input` arrives it remains in pending correlation state rather than the queued list. Session shutdown only fails entries that reached the queued list.

This is distinct from #928's `agent_end`/`agent_settled` false-idle race and is not introduced by that fix.

## Expected behavior

- An accepted steering-inbox request must reach exactly one terminal acknowledgement: delivered, queued then delivered, or failed.
- A `sendUserMessage` rejection before correlated `input` must not silently consume the request.
- Manual compaction should retain/requeue the request for safe follow-up delivery, or write a closed failed acknowledgement with a retry-safe reason.
- Session shutdown must fail both queued and pending-correlation requests.
- No duplicate delivery when compaction ends and the request is retried.

## Suggested tests

- request -> manual compaction rejection -> no `input`: request remains retryable/pending and is not lost;
- compaction ends -> one follow-up delivery -> one terminal acknowledgement;
- session shutdown while pending correlation -> failed acknowledgement;
- duplicate identical guidance remains FIFO-correlated.

## Context

- #928 exact-head review
- Current steering inbox: `src/runs/shared/subagent-prompt-runtime.ts`
- Current Pi manual-compaction guard rejects before `input` and ExtensionAPI cannot synchronously report the rejection.

PR #934, opened 2026-08-09
Fix auto steering between agent_end and agent_settled

### LINKED ISSUE #928: Steering inbox treats agent_end as idle before agent_settled
## Summary

The child steering inbox marks the agent idle on `agent_end`, before Pi emits `agent_settled`. A steering/auto request arriving in that interval can be dispatched through `sendUserMessage(..., { deliverAs: "steer" })` while Pi still considers the prompt active, producing:

```text
Extension "<runtime>" error: Agent is already processing a prompt. Use steer() or followUp() to queue messages, or wait for completion.
```

## Current code

`src/runs/shared/subagent-prompt-runtime.ts` currently projects:

```ts
onRuntimeEvent("agent_start", () => { agentRunning = true; return activate(); });
onRuntimeEvent("agent_end", () => { agentRunning = false; inTurn = false; return activate(); });
```

This makes `mode: "auto"` choose the idle/between-turn steering path after `agent_end`, even though the runtime is not yet settled. The foreground/background completion state machines already understand `agent_settled` under #444; the steering inbox state machine does not.

## Reproduction sequence

1. Child emits `agent_start` and enters a turn.
2. Child emits `turn_end` / `agent_end` while Pi still has post-turn processing or queued continuation work.
3. A fresh steering-inbox request with `mode: "auto"` arrives before `agent_settled`.
4. The inbox flush sees `agentRunning=false` and dispatches an immediate steer.
5. Pi rejects it as already processing a prompt.

The race was observed repeatedly while coordinating long-running worker completion, and stopped when fresh prompts were withheld until settlement.

## Expected behavior

- Treat `agent_settled` as the preferred idle watermark.
- Do not set the steering inbox to idle merely because `agent_end` fired.
- Preserve a bounded compatibility fallback for Pi versions without `agent_settled`.
- If `agent_end.willRetry === true`, remain running/queued.
- `auto` guidance in the `agent_end` -> `agent_settled` interval should use follow-up/queue semantics, not immediate steer.

## Suggested tests

- `agent_start` -> `turn_end` -> `agent_end` -> auto request -> `agent_settled`: request remains queued/follow-up until settled and no immediate steer is sent.
- `agent_end({willRetry:true})` -> request -> next turn: request survives and is delivered once.
- `agent_settled` -> auto request: immediate between-turn delivery remains supported.
- Legacy runtime without `agent_settled`: bounded fallback does not deadlock steering.

Related: #444 fixed lifecycle-aware child completion/drain; this report is the adjacent prompt-steering state machine.

jtac:
## Summary

Fixes the steering-inbox race where `agent_end` was treated as idle before Pi emitted `agent_settled`.

- use the real extension `agent_end` shape without relying on listener-only `willRetry`
- keep `mode:auto` on safe `followUp` semantics through delayed post-agent work, retries/new turns, and legacy runtimes without `agent_settled`
- restore immediate between-turn auto delivery only after actual `agent_settled`
- preserve explicit steer, FIFO correlation, queued/delivered acknowledgements, exactly-once behavior, and shutdown handling
- add deterministic lifecycle regressions and changelog entry

## Validation

- focused prompt-runtime/child-protocol unit tests: 58/58
- focused lifecycle integration: 2/2
- `npm run typecheck`
- full-unit base/head comparison: same two pre-existing failures and six cancellations; repaired head adds six passing tests and no new failure
- independent exact-head review PASS at `5fd97a9d3e56e1451b4b3fecb870787879e4e36b`

## Residual

A separate pre-existing manual-compaction pending-acknowledgement gap is tracked in #933.

Closes #928.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Keep `auto` steering queued between `agent_end` and `agent_settled`, including retrying turns, while retaining immediate settled delivery and a bounded legacy fallback. (#928)
+
 ## [0.45.1] - 2026-08-09
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Fixed
-- Keep `auto` steering queued between `agent_end` and `agent_settled`, including retrying turns, while retaining immediate settled delivery and a bounded legacy fallback. (#928)
+- Keep `auto` steering on safe follow-up semantics after `agent_end` until a real `agent_settled`, while preserving explicit steer, immediate settled delivery, and legacy progress without inferring idle from elapsed time. (#928)
 
 ## [0.45.1] - 2026-08-09
 
--- a/src/runs/shared/subagent-prompt-runtime.ts
+++ b/src/runs/shared/subagent-prompt-runtime.ts
@@ -67,6 +67,13 @@ const SUBAGENT_ORCHESTRATION_SKILL_NAME_PATTERN = /<name>\s*pi-subagents\s*<\/na
 const PROJECT_CONTEXT_HEADER = "\n\n# Project Context\n\nProject-specific instructions and guidelines:\n\n";
 const SKILLS_HEADER = "\n\nThe following skills provide specialized instructions for specific tasks.";
 const DATE_HEADER = "\nCurrent date:";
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks @jtac. The #928 fix has already landed through #929 at `94e59f103e3a4cfa9caf84701dfa921c2471d362` and is now on `main` via merge commit `685df610f346106d22b668831139db460e3e405d`.

I am closing this PR as superseded because its exact head conflicts with current `main` and covers the same behavior. Your report and contribution are preserved with visible `CHANGELOG.md` credit for #928.

Issue #953, opened 2026-08-10
worktree:true children run in the parent repo, not the managed worktree (confirmed on 0.45.2)


flopsi:
## Summary

`worktree: true` on parallel `runs.all` children **does not isolate the child's working directory** — the child runs in the parent session's main repo on its branch (`master`), not in the managed worktree. The worktree directory is created and the launch contract points at it, but the child's actual `cwd` is never switched to it. Confirmed on **0.45.2 (latest)**, Windows 11, Git Bash.

This is the same class of bug as #907 (fix #910 corrected only the *reported* cwd in async metadata, not the actual child process cwd).

## Repro (minimal, 2 minutes)

Throwaway repo, clean tree, `runs.all` with two `worktree: true` children that only run `pwd` / `git rev-parse --show-toplevel` / `git branch --show-current` and report:

```js
return await runs.all([
  { key: 'a', agent: 'worker', worktree: true, cwd: 'D:/tmp/wt-repro', task: probe },
  { key: 'b', agent: 'worker', worktree: true, cwd: 'D:/tmp/wt-repro', task: probe },
]);
```

## Observed (both children, 0.45.2)

```
pwd                       → /d/000projects/cotscope      (parent session repo)
git rev-parse --show-toplevel → D:/000projects/cotscope   (parent session repo)
git branch --show-current → master                        (parent branch, not a worktree branch)
```

Meanwhile the launch contract says the child should read from `D:\apps\TEMP\pi-worktree-<runId>-0\context.md` — the worktree **directory** is created and context staged there, but the child process is spawned with the parent session's cwd.

## Impact

- Parallel mutating children all write to the **same shared main repo** (confirmed in a real run: worker B saw worker A's uncommitted changes; a worker's pre-commit hook failed because another worker's edit was in "this shared worktree").
- A child can commit to the parent's `master` instead of an isolated worktree branch (same incident class as #907).
- Isolation silently fails — there is no error, no warning; the run reports success.

## Expected behavior

A child launched with `worktree: true` must have its actual process cwd set to the managed worktree (`worktreeSetup.worktrees[i].agentCwd`). If the runtime cannot guarantee the requested cwd, it should fail the launch before any mutation-capable work starts.

## Evidence detail

Child session files (first line) record the cwd the child was spawned with:

```
{"type":"session","version":3,"cwd":"D:\000projects\cotscope", ...}   ← main repo, not the worktree
```

[...]

PR #955, opened 2026-08-10
fix: name the child agent in workflow run status

albertgwo:
## Summary
- carry the resolved child agent name on the workflow run trace entry
- name workflow status steps by that agent, keeping the caller's stable key as the label only when no agent name is available
- leave run targeting unchanged, since status, transcript, and nested addressing still key off `workflowKey`

Before this, `runs.run("budget-pot-oracle", { agent: "oracle" })` rendered in FleetView as `budget-pot-oracle (budget-pot-oracle)`, because the trace entry carried only the caller's key and the parent used it for both the agent and the label. Watching a fleet gave no way to tell which agent each row was without opening every run. Foreground rows were unaffected, so this only showed up for workflow children.

## Validation
- `npm run typecheck`
- `npm run test:unit` (1711 pass, 1 skip)
- targeted scripted-workflow and fleet tests (84 pass)

One open question for review: the key is dropped as the label whenever an agent name exists, so several parallel children of the same agent render identically, matching how foreground parallel rows already behave. Happy to always show both if you would rather keep the key visible.

### DIFF
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -4427,7 +4427,7 @@ export function createSubagentExecutor(deps: ExecutorDeps): {
 								if (entry.durationMs === undefined) delete existing.durationMs;
 								else existing.durationMs = entry.durationMs;
 							} else {
-								status.steps?.push({ agent: entry.key, label: entry.key, workflowKey: entry.key, parentWorkflowRunId: workflowRunId, status: mapped, startedAt: Date.now() });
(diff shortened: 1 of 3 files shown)


albertgwo:
Withdrawing: opened prematurely on my side. Closing rather than leaving it in the maintainer's queue.

albertgwo:
Reworked the title and description to match the repo's conventions after closing this earlier; the branch is rebased on current `main` (v0.45.2). Sorry for the churn on your queue.

nicobailon:
Thanks @albertgwo. I landed a current-base replacement in #959, preserving the workflow-status agent-name behavior from this PR and adding the extra guards and tests needed for #953 and #957. Closing this one as superseded by #959.

PR #972, opened 2026-08-11
fix: sanitize tool previews to one terminal line

xz-dev:
## Summary

Keep async tool previews on one physical terminal line so repeated widget updates replace the existing frame instead of leaving stacked rows in terminal scrollback.

- Normalize canonical tool argument previews before persistence and rendering.
- Strip 7-bit and C1 terminal control strings with a bounded linear scanner.
- Preserve readable text while truncating without splitting UTF-16 surrogate pairs.
- Reuse the same display normalization for the RPC Fleet projection.

## Root cause

Multiline shell commands could pass through `extractToolArgsPreview()` with embedded CR/LF, tabs, or terminal controls. The async widget then placed that preview inside one `Component.render()` array element. Pi TUI treats each element as one physical row, while the terminal rendered the embedded newline as additional rows. Differential redraw therefore moved relative to the wrong row count and left earlier live frames stacked below the current one.

## Validation

- Exact-baseline regressions fail on `80a4cc6`:
  - canonical preview controls remain;
  - widget current/recent tool rows contain embedded physical line breaks;
  - RPC output can split a surrogate pair.
- `test/unit/foreground-tool-call-compaction.test.ts`: 11/11 passed.
- Focused physical-line widget regression: 1/1 passed.
- RPC/Fleet focused suite: 80/80 passed during independent review.
- Focused committed-head suite: 49/49 passed.
- Linear sanitizer regression repeated without flakes; independent scaling remained near-linear.
- `git diff --check`: passed.

## Review status

This PR is intentionally a **draft**. The change passed independent publication review with explicit residual risks, but it will remain draft until it has been exercised in the reporter's real terminal workflow.

## Residual risks

- Existing persisted previews created before this fix are not rewritten.
- Incomplete terminal control strings fail closed by discarding the unterminated remainder.
- A few downstream display-only slices outside this canonical persistence path remain unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Persist workflow child attempts, status heartbeats, session paths, and artifacts in their enclosing mission, and add explicit mission decision resolution.
 
 ### Fixed
+- Keep tool argument previews on one physical terminal line so live widget updates do not leave stacked terminal frames. Thanks to @xz-dev for #972.
 - Suspend subagent status widgets during automatic compaction to prevent duplicate terminal frames.
 - Make live foreground workflow children visible as workflow-owned Fleet rows, route Herdr inspection to their workflow parent, and report their active and needs-attention state to Herdr. Thanks to @lukechen526 for #965.
 - Wait for retained-child resumes inside `workflowScript` to finish and return completed output before the script continues (#961).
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks for the draft and the clear repro direction here. I rebuilt the fix on current `main` with a smaller shared display-text path, kept the same goal, and credited you in the changelog.

Replacement PR: #974

nicobailon:
Merged the replacement implementation in #974. Thanks again for the draft and repro direction here.

PR #986, opened 2026-08-11
perf: retain async status cache entries

### LINKED ISSUE #982: Async run scans re-read every status.json once the tree passes 50 runs; the eviction policy is not the lever
First, thank you for pi-subagents — orchestrating a fleet of agents from inside a pi session is a genuinely hard problem, and it has been stable enough to lean on daily. This is a small report against something I use constantly.

## Summary

Once the async run tree holds more than 50 runs, every full scan re-reads and re-parses **every** `status.json` on disk. The `readStatus` memo (`src/shared/utils.ts:103`) is capped at 50 entries, and its consumers sweep run directories in order — a cyclic access pattern over more entries than the cache holds, which drives the hit rate to zero rather than merely degrading it.

The important part, and the reason I am filing rather than sending a one-line patch: **the eviction policy is not the lever.** I measured FIFO (current), LRU, and random eviction at the same cap; all three are within noise of each other at realistic tree sizes. Only capacity relative to the working set matters. An agent picking up this issue will very reasonably reach for "make it an LRU" — I did — and that change measurably does nothing.

Because `listAsyncRuns` (`src/runs/background/async-status.ts:371`) is reached from turn-boundary hooks rather than a timer — `agent_end` via `listRetainedChildren` (`src/runs/background/retained-children.ts:35`), and once per subagent tool result — the cost arrives as a burst of synchronous work on the parent's event loop during an ordinary turn.

Nothing prunes the directories that make the tree grow: `ASYNC_DIR` (`src/shared/types.ts:1935`), `RESULTS_DIR` (`:1934`) and `NESTED_EVENTS_DIR` (`src/runs/shared/nested-events.ts:34`) have no cleanup call anywhere in `src`, while `artifacts/` and `chain-runs/` both do (`cleanupOldArtifacts`, and `cleanupOldChainDirs` at `src/shared/settings.ts:202`). So the working set only grows, and 50 is crossed early.

This is distinct from #579. That issue was about how many nested sidecar files get *written* and the `processedEvents` replay cap; this one is about the cache in front of `readStatus` failing structurally, and it would behave identically with #579's write volume already fixed.

## Versions

- pi-subagents 0.46.0; defect confirmed present on `main` @ `c386b25`, and all line numbers below are `main` line numbers
- pi-coding-agent 0.84.1
- node v26.5.0, macOS 26.5 arm64

## Mechanism (verified against the code, not inferred)

1. **Eviction is insertion-order, and `Map.set` on an existing key does not reorder it** — `src/shared/utils.ts:174-178`:
   ```ts
   if (statusCache.size > 50) {
     const firstKey = statusCache.keys().next().value;
     if (firstKey) statusCache.delete(firstKey);
   }
   ```
2. **The consumers sweep in that same order.** `listAsyncRuns` walks run directories and calls `readStatus` per directory. With N > 50, the entry evicted while the sweep is at #51 is the one requested first on the next sweep.
3. **The scans are turn-driven.** `listRetainedChildren` passes no `entryLimit`, so `agent_end` re-scans the whole tree every turn although it only needs `states: ["complete"]` for the current session.
4. **A second unbounded scan rides along.** `buildNestedRouteIndex()` (`src/runs/shared/nested-events.ts:554`) `readFileSync`s every route directory on each reconciling `listAsyncRuns`, whether or not the route set changed.

## Reproduction

Self-contained, no real tree needed. Save as `status-cache-cliff.ts` in a checkout and run `bun status-cache-cliff.ts`:

```ts
import * as fs from "node:fs";
import * as os from "node:os";
import * as path from "node:path";
import { readStatus } from "./src/shared/utils.ts";

const root = fs.mkdtempSync(path.join(os.tmpdir(), "status-cache-cliff-"));
const dirs: string[] = [];
for (let i = 0; i < 1200; i++) {
  const dir = path.join(root, `run-${String(i).padStart(4, "0")}`);
  fs.mkdirSync(dir);
  fs.writeFileSync(
    path.join(dir, "status.json"),
    JSON.stringify({ id: `run-${i}`, state: "complete", agent: "reviewer", startedAt: 0, updatedAt: 0 }),
  );
  dirs.push(dir);
}

console.log(" N runs   per sweep (ms)   per run (us)");
for (const n of [40, 49, 50, 51, 60, 123, 400, 1200]) {
  const sub = dirs.slice(0, n);
  for (let warm = 0; warm < 3; warm++) for (const dir of sub) readStatus(dir);
  const samples: number[] = [];
  for (let i = 0; i < 11; i++) {
    const started = performance.now();
    for (const dir of sub) readStatus(dir);
    samples.push(performance.now() - started);
  }
  samples.sort((a, b) => a - b);
  const median = samples[5]!;
  console.log(`${String(n).padStart(6)}   ${median.toFixed(3).padStart(14)}   ${((median / n) * 1000).toFixed(1).padStart(12)}`);
}

fs.rmSync(root, { recursive: true, force: true });
```

Output on `main` — note the step between 50 and 51:

```
 N runs   per sweep (ms)   per run (us)
     40            0.556           13.9
     49            0.917           18.7
     50            0.688           13.8
     51            3.219           63.1
     60            3.747           62.5
    123           10.182           82.8
    400           25.813           64.5
   1200           90.358           75.3
```

## What does and does not fix it (all measured with the script above)

Per-run cost in µs, same tree, same machine:

| eviction policy | cap | N=51 | N=123 | N=400 | N=1200 |
|---|---|---|---|---|---|
| FIFO (current) | 50 | 86.1 | 85.6 | 76.0 | 77.9 |
| LRU (delete + re-set on hit) | 50 | 78.8 | 77.0 | 66.4 | — |
| random victim | 50 | 17.3 | 66.2 | 66.1 | — |
| FIFO | 512 | 17.7 | 17.3 | 17.8 | 79.6 |
| FIFO | 4096 | 17.6 | 17.5 | 16.7 | 17.6 |

Reading of that table:

- **LRU is not a fix.** A cyclic scan over more entries than the cache holds evicts precisely the entry needed next, under LRU exactly as under FIFO. This is the one thing I would most like to save an implementer from discovering after the fact.
- **Random eviction helps only when N is barely over the cap** (hit rate tends to capacity/N), so it is not a fix either.
- **Capacity is the whole story.** A cap comfortably above the working set gives ~17.5 µs per run instead of ~80. But raising the cap only relocates the cliff — at cap 512, N=1200 is back to 79.6 µs.
- Therefore the durable fix is **bounding the working set**, with a larger cap as the cheap immediate mitigation.

## Suggested fix, smallest first

1. **Bound what `agent_end` scans.** `listRetainedChildren` (`src/runs/background/retained-children.ts:35`) wants completed runs for the current session; give it the `entryLimit` treatment `collectFleetSnapshot` already uses. This removes the per-turn full sweep regardless of cache behaviour, and is the only change that helps at every tree size.
2. **Prune `async-subagent-runs/`, `async-subagent-results/` and `nested-subagent-events/`**, mirroring `cleanupOldChainDirs` (`src/shared/settings.ts:202`). Without this, every other fix postpones rather than solves.
3. **Raise the cache cap** from 50 to something above a realistic working set, ideally bounded by bytes rather than entries since `status.json` sizes vary. Cheap, immediate, and measured at ~4.9× on the per-run cost — but please treat it as mitigation, not the fix.
4. **Memoise `buildNestedRouteIndex`** (`src/runs/shared/nested-events.ts:554`) behind an mtime check on `NESTED_EVENTS_DIR`; it re-reads every route file even when the route set has not changed.

### Acceptance criteria for a fix

- The repro script above shows no step between N=50 and N=51, and per-run cost stays flat to N=1200.
- `agent_end` on a session with one completed child does not stat or read run directories belonging to other sessions.
- A focused regression covering the scan bound in (1) — narrow, on the changed file, not a new integration test.

### Honest limits of the evidence

At the ~123 status files my tree currently holds, the per-turn cost is roughly 10–30 ms: perceptible against an 80 ms spinner, but I could not reproduce a multi-hundred-millisecond stall at that size. Sampling event-loop lag at 80 ms over 12-second windows, the p99 was 22.5 ms at the real tree and 131.9 ms at a synthetic 10× tree, with 6 ticks over 100 ms. So the visible symptom today is mild; the reason I am reporting it is that nothing bounds the growth toward the size where it is not.

I also want to be clear about what is **not** the cause, so it need not be re-investigated: the twelve `setInterval` pollers contribute no measurable lag at any tree size (the 250 ms supervisor-channel poll reads an empty `supervisor-channels/`, `collectFleetStatusEntries` is pure in-memory, and the 3000 ms result-watcher interval is an `fs.watch` fallback that never starts on macOS), and child-process CPU contention is not a factor (0/1/2/4/8 busy children all gave p99 ≤ 1.2 ms on 12 cores).

---

*Investigated with Claude Opus 5 at max reasoning effort, driven by me. Every number above was executed against the real modules rather than estimated; the LRU and random-eviction rows exist because I tested my own proposed fix and it failed. Happy to open a PR for any of the four items.*

nicobailon:
This improves async status reads for large async-run directories without reducing freshness.

`readStatus()` no longer evicts cached entries at 50 paths. It still performs a `statSync()` on each read and only returns cached JSON when `mtimeMs`, `ctimeMs`, `size`, and `ino` still match, so the status view stays fresh while ordered sweeps over more than 50 runs can reuse parsed JSON.

This update also addresses the Greptile cache-retention concern. Complete async-root scans now prune cached `status.json` entries for run directories that no longer exist. Targeted scans and `entryLimit` subset scans do not prune, so partial views cannot discard valid entries for unscanned runs.

Validation:
- Focused cache regression for more than 50 statuses, same-mtime replacement invalidation, and removed-run pruning.
- `test/integration/async-status.test.ts`
- `npm run typecheck`
- `git diff --check`
- Fresh review found no blockers.

Fixes #982.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Move project-scoped pi-subagents storage from `.pi-subagents/` to `.pi/subagents/` for cleaner project roots. Thanks to @yceachan for #971.
 - Clarify the accepted mission launch object contract for tool callers.
 - Reduce repeated async status parsing and workflow trace projection work.
+- Keep parsed async statuses cached beyond 50 runs while preserving per-read freshness checks. Thanks to @bcanvural for #982.
 
 ### Fixed
 - Keep tool argument previews on one physical terminal line so live widget updates do not leave stacked terminal frames. Thanks to @xz-dev for #972.
--- a/src/runs/background/async-status.ts
+++ b/src/runs/background/async-status.ts
@@ -4,7 +4,7 @@ import { formatDuration, formatModelThinking, formatTokens, shortenPath } from "
 import { formatActivityLabel, formatParallelOutcome } from "../../shared/status-format.ts";
 import { type ActivityState, type AsyncJobStep, type AsyncParallelGroupStatus, type AsyncStatus, type CostSummary, type Details, type LaunchResolvedChildExtensionsV1, type RuntimeAcknowledgedChildExtensionsV1, type NestedRunSummary, type SteeringStatus, type SubagentRunMode, type TokenUsage, type TurnBudgetState, type UsageBudgetState, type ChainCheckpointState } from "../../shared/types.ts";
 import type { ResolvedSubagentCapabilityCeiling, SubagentCapabilityAudit } from "../shared/capability-ceiling.ts";
-import { readStatus } from "../../shared/utils.ts";
+import { pruneStatusCacheForAsyncRoot, readStatus } from "../../shared/utils.ts";
 import { attachRootChildrenToSteps, buildNestedRouteIndex, type NestedRoute, projectNestedEvents } from "../shared/nested-events.ts";
 import { formatNestedRunStatusLines } from "../shared/nested-render.ts";
 import { flatToLogicalStepIndex, normalizeParallelGroups } from "./parallel-groups.ts";
@@ -370,6 +370,7 @@ function sortRuns(runs: AsyncRunSummary[]): AsyncRunSummary[] {
 
 export function listAsyncRuns(asyncDirRoot: string, options: AsyncRunListOptions = {}): AsyncRunSummary[] {
 	let entries: string[];
+	let scannedCompleteRoot = false;
 	try {
 		if (options.runId !== undefined) {
 			const resolution = resolveTargetedAsyncRun(asyncDirRoot, options.runId, options.sessionId);
@@ -383,6 +384,7 @@ export function listAsyncRuns(asyncDirRoot: string, options: AsyncRunListOptions
 					: [];
 		} else {
 			entries = fs.readdirSync(asyncDirRoot).filter((entry) => isAsyncRunDir(asyncDirRoot, entry));
+			scannedCompleteRoot = true;
 		}
 	} catch (error) {
 		if (isNotFoundError(error)) return [];
@@ -392,6 +394,7 @@ export function listAsyncRuns(asyncDirRoot: string, options: AsyncRunListOptions
 	}
 
 	if (options.entryLimit !== undefined) {
+		scannedCompleteRoot = false;
 		const limit = Math.max(0, Math.floor(options.entryLimit));
 		entries = entries
 			.map((entry) => {
@@ -409,6 +412,8 @@ export function listAsyncRuns(asyncDirRoot: string, options: AsyncRunListOptions
 			.slice(0, limit)
 			.map((candidate) => candidate.entry);
 	}
+
+	if (scannedCompleteRoot) pruneStatusCacheForAsyncRoot(asyncDirRoot, entries);
 
 	const allowedStates = options.states ? new Set(options.states) : undefined;
 	const runs: AsyncRunSummary[] = [];
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -102,6 +102,23 @@ export function getAgentDir(): string {
 
 const statusCache = new Map<string, { mtime: number; ctime: number; size: number; ino: number; status: AsyncStatus }>();
 
+export function pruneStatusCacheForAsyncRoot(asyncDirRoot: string, runIds: Iterable<string>): number {
+	const root = path.resolve(asyncDirRoot);
+	const currentStatusPaths = new Set(
+		Array.from(runIds, (runId) => path.resolve(root, runId, "status.json")),
+	);
+	let removed = 0;
+	for (const statusPath of statusCache.keys()) {
+		const resolved = path.resolve(statusPath);
+		const relative = path.relative(root, resolved);
+		if (relative && !relative.startsWith("..") && !path.isAbsolute(relative) && !currentStatusPaths.has(resolved)) {
+			statusCache.delete(statusPath);
+			removed++;
+		}
+	}
+	return removed;
+}
+
 function getErrorMessage(error: unknown): string {
 	return error instanceof Error ? error.message : String(error);
 }
@@ -128,7 +145,10 @@ export function readStatus(asyncDir: string): AsyncStatus | null {
 	try {
 		stat = fs.statSync(statusPath);
 	} catch (error) {
-		if (isNotFoundError(error)) return null;
+		if (isNotFoundError(error)) {
+			statusCache.delete(statusPath);
+			return null;
+		}
 		throw new Error(`Failed to inspect async status file '${statusPath}': ${getErrorMessage(error)}`, {
 			cause: error instanceof Error ? error : undefined,
 		});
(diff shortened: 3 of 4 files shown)


bcanvural:
Confirmed on my side, and thank you for the quick turnaround. I ran the repro script from #982 unchanged against `pr986` and the cliff is gone:

| N runs | `main` (per run) | this PR (per run) |
|---|---|---|
| 40 | 13.9 µs | 14.6 µs |
| 49 | 18.7 µs | 19.4 µs |
| 50 | 13.8 µs | 13.5 µs |
| **51** | **63.1 µs** | **17.5 µs** |
| 60 | 62.5 µs | 16.8 µs |
| 123 | 82.8 µs | 17.6 µs |
| 400 | 64.5 µs | 17.6 µs |
| 1200 | 75.3 µs | 17.8 µs |

Full-sweep time at 1200 runs goes 90.4 ms → 21.3 ms. That meets both acceptance criteria I listed in the issue: no step between 50 and 51, and per-run cost flat out to 1200. Clearing the entry on a missing status file is a nice catch too — that was the leak path an uncapped map would otherwise have.

**One measured trade-off, offered as information rather than an objection.** The cache is now unbounded, and real `status.json` files are much larger than the ones in my repro — on this machine, 371 of them have a median of 8,994 B and a mean of 12,704 B (max 267 KB). Retained heap, measured under `node --expose-gc` with 10.3 KB status files:

| entries cached | retained | per entry |
|---|---|---|
| 100 | 1.3 MB | 13.4 KB |
| 400 | 5.2 MB | 13.4 KB |
| 1000 | 13.1 MB | 13.4 KB |
| 1500 | 19.6 MB | 13.4 KB |

Linear at about 1.3× the on-disk size of every status file read during a session. At my current tree that is roughly 5 MB, which is nothing; the reason I mention it is that its ceiling is set by something with no ceiling — see below. If you ever want a bound that does not reintroduce a cliff, a byte budget rather than an entry count is the shape that survives a scan (an entry count re-breaks the moment the working set passes it, whatever the eviction policy).

**One process note.** The PR says `Fixes #982`, which will auto-close the issue on merge, but the issue covered four things and this addresses the first. Still open after this lands:

- nothing prunes `async-subagent-runs/`, `async-subagent-results/` or `nested-subagent-events/` (`src/shared/types.ts:1934-1935`, `src/runs/shared/nested-events.ts:34`), while `artifacts/` and `chain-runs/` both have cleanup. This is what makes the tree — and now the cache — grow without limit;
- `listRetainedChildren` (`src/runs/background/retained-children.ts:35`) passes no `entryLimit`, so `agent_end` sweeps the whole tree every turn;
- `buildNestedRouteIndex` (`src/runs/shared/nested-events.ts:554`) re-reads every route file on each reconciling `listAsyncRuns` — 10.5 ms at 204 routes here, and I am at 612 now.

Happy to open a separate issue for those three so they do not disappear with the close, or to leave it to you if you would rather triage them differently. Either way this PR is the right first cut — it is the one that helps at every tree size.

---

*Same disclosure as the issue: measurements run with Claude Opus 5 at max reasoning effort, driven by me. The latency table is the issue's own script against your branch; the memory table is `node --expose-gc` with synthetic files sized to this tree's median.*

PR #1001, opened 2026-08-11
fix: carry the requested agent on started workflow trace entries

albertgwo:
## Summary
- carry the requested child agent name on `started` workflow trace entries, not just the terminal one
- keep launch resolution authoritative, so the canonical agent name still overwrites the requested one when the child settles

Follow-up to `2439218 fix: harden workflow child launches`, which attaches the resolved agent to the terminal trace entry. Because `started` carries no agent, the parent falls back to the caller's stable key while a child runs, so a fleet shows rows like `budget-pot-oracle` for every in-flight child and real agent names appear only as rows complete. Observed on v0.46.0:

```
state=started    key=budget-pot-oracle    agent=(undefined)
state=completed  key=budget-pot-oracle    agent=oracle
```

With this change the `started` entry reports `oracle`, so a child is identifiable for its whole lifetime. Aliases and retained resume are unaffected, since the terminal entry still reports the resolved name.

## Validation
- `npm run typecheck`
- `npm run test:unit` (1764 pass, 1 skip)
- harness replaying the parent's step projection and widget naming over a three-child `runs.all`, confirming in-flight rows read `oracle, worker, worker` rather than `budget-pot-oracle, w1, w2`

### DIFF
--- a/src/workflows/scripted-workflow.ts
+++ b/src/workflows/scripted-workflow.ts
@@ -469,10 +469,13 @@ function validateKey(value: unknown, owner = "runs.run"): string {
 	return value;
 }
 
-function workflowStringMetadata(params: Record<string, unknown>): Pick<WorkflowScriptTraceEntry, "phase" | "label"> {
+function workflowStringMetadata(params: Record<string, unknown>): Pick<WorkflowScriptTraceEntry, "phase" | "label" | "agent"> {
 	return {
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks again for the fix. I landed the same behavior in #1009 so we could carry it on current `main` and add the required visible changelog credit for your contribution.

The replacement keeps your workflow trace change and credits @albertgwo for #1001 in `CHANGELOG.md`.

PR #1049, opened 2026-08-13
feat: add builtin tool inheritance

### LINKED ISSUE #1047: Allow builtin roles to inherit Pi's full toolset without shadow agent files
## Problem

Builtin roles declare a strict `tools` allowlist. `subagents.agentOverrides.<role>.tools` can replace that list, but cannot remove it. As a result, there is no settings-only way to let a builtin child inherit the same ambient Pi tools and extensions as a normal Pi session.

A concrete failure: the bundled `researcher` requires `web_search`, `fetch_content`, and `get_search_content` from `pi-web-access`. A configured Exa extension exposes equivalent `exa_*` tools, but the builtin role will not launch because its fixed allowlist names are absent.

## Current workaround

Shadow every builtin role with a user agent file (for example `~/.pi/agent/agents/researcher.md`) that has the same `name` but omits `tools`. This makes pi-subagents omit `--tools`, so the child receives normal builtin tools and ambient extensions.

That workaround is surprisingly heavy for a common policy: it duplicates/shadows role definitions, requires one file per role, and can drift from upstream role prompts.

## Requested behavior

Please add a settings-only opt-out of a builtin role's tool allowlist. For example, either:

```json
{
  "subagents": {
    "agentOverrides": {
      "researcher": { "tools": "inherit" }
    }
  }
}
```

or a global opt-in for all builtins.

The result should be the same as an agent definition with no `tools` frontmatter: do not pass `--tools`; load normal Pi extensions; and let the child use the normal ambient tool registry. Explicit `tools: []` should continue to mean no tools.

This would let users keep the builtin role prompts/models while choosing an unrestricted-agent policy.

davidarny:
Closes #1047

Adds `subagents.inheritBuiltinTools`, a global opt-in that removes builtin role tool allowlists so child sessions inherit Pi's normal ambient tools and extensions.

- Keeps the default strict allowlists unchanged.
- Keeps explicit per-role `tools` settings within the winning config scope.
- Gives project settings precedence over user settings.
- Documents the setting and covers parsing and precedence with unit tests.

Checks:
- `npm run typecheck`
- `node --experimental-strip-types --test test/unit/agent-overrides.test.ts`

The full unit command is environment-sensitive here because the checkout discovers global user agent definitions under `~/.agents`; the focused tests pass in their isolated test home.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 - Add `debug.run` for async run lifecycle diagnostics without exposing prompts, secrets, or transcripts (#1037).
+- Add `subagents.inheritBuiltinTools` to let builtin roles inherit Pi's ambient tools and extensions. Thanks to @davidarny for #1049.
 
 ## [0.48.0] - 2026-08-13
 
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -90,7 +90,7 @@ To let every builtin role use Pi's normal tools and ambient extensions instead o
 }
 ```
 
-An explicit `agentOverrides.<name>.tools` entry still wins for that role. Project settings take precedence over user settings.
+An explicit `agentOverrides.<name>.tools` entry in the winning configuration scope still wins for that role. Project settings take precedence over user settings, including `inheritBuiltinTools`.
 
 Disable and restore:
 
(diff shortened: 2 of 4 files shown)


nicobailon:
Thanks for putting this together. I opened #1052 as a narrower replacement that keeps the behavior per-role with `tools: "inherit"`, so strict builtin allowlists stay the default.

I kept credit for your work in the changelog and PR body. I am leaving this open until the replacement lands so the history stays easy to follow.

nicobailon:
#1052 has landed with the narrower per-role `tools: "inherit"` behavior. That preserves the useful settings-only opt-out from this PR while keeping strict builtin allowlists as the default.

Thanks again for the original implementation. The changelog keeps credit for #1049, so I am closing this as superseded.

PR #1078, opened 2026-08-13
feat(rpc): publish structured async job snapshots

yanqianglu:
## Summary

Publish the async job tracker’s authoritative runtime state through Pi’s existing RPC `setWidget` string-array channel while preserving the current component widget in TUI mode.

RPC clients currently receive no usable async widget because component factories have no RPC wire representation. This adds a bounded, versioned `PI_SUBAGENT_ASYNC_JSON:` payload derived directly from `AsyncJobState`, including dynamic workflow steps and nested descendants.

Terminal snapshots are also attached to the existing hidden completion notification details, so clients can restore completed child rows from session history after the live widget is cleared.

## Design

- `ctx.mode === "tui"`: unchanged component widget
- `ctx.mode === "rpc"`: structured `string[]` widget
- exact run identity via `asyncId`
- stable child identity via `workflowKey` or step index
- bounded runs, children, nesting depth, and text lengths
- excludes working directories, artifact/session paths, tool arguments, recent output, and error text
- records workflow task/label/phase on runtime steps instead of parsing workflow source

## Validation

- `npm run typecheck`
- async snapshot unit tests: 2 passed
- notification unit tests: 26 passed
- RPC renderer test passed
- terminal-snapshot-before-clear tracker test passed
- existing render-widget integration suite: 35 passed
- `git diff --check`

The broader local suite reaches 1,880 passing tests; remaining failures are existing environment-sensitive tests involving process inspection, user-profile fixtures, and filesystem behavior. This PR is draft pending maintainer feedback and clean hosted CI.

### DIFF
--- a/docs/extension-api.md
+++ b/docs/extension-api.md
@@ -56,6 +56,16 @@ The DTO intentionally never exposes run, async, or tool IDs. Clients must ignore
 
 `pi.events` is in-process only. It does not reach separate Pi processes or child subagents; use the file lifecycle artifacts or `pi-intercom` for cross-process coordination.
 
+### External Pi RPC async snapshot
+
+The in-process request/reply bus above is available only to extensions in the same Pi process. Host applications running Pi in `--mode rpc` receive async job state through Pi's standard fire-and-forget extension UI protocol instead.
+
+When `ping.capabilities.rpcAsyncSnapshot` is advertised, it identifies the transport, widget key, payload prefix, and schema version. Version 1 uses `setWidget("subagent-async", string[])`; the first line starts with `PI_SUBAGENT_ASYNC_JSON:` followed by a bounded JSON snapshot.
+
+The snapshot is derived from the same `AsyncJobState` used by the TUI widget. It contains exact `asyncId` run identity, stable workflow keys or step indexes, lifecycle status, agent/task/model metadata, and bounded nested descendants. It omits working directories, artifact/session paths, tool arguments, recent output, and error text. Terminal snapshots are also attached to hidden `subagent-notify` message details for session replay.
+
+DTO types and the prefix constant are exported from `pi-subagents/async-rpc-snapshot`. Consumers must check `version`, ignore unknown fields, and treat the live widget as replace-by-snapshot state rather than appending rows.
+
 ## Launch contract preflight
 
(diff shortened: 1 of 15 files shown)


nicobailon:
Thanks @yanqianglu. I reviewed this draft as part of the backlog pass.

This should stay draft/blocked for now:

1. `subagent-notify` messages can carry `details: { asyncSnapshot }`, but the existing renderer expects `SubagentNotifyDetails` and can read `resultPreview.trim()` from an incompatible shape.
2. The terminal replay snapshot is built from in-memory async job state before the result-watcher completion event updates that state, so fast completions can persist a stale or childless snapshot.
3. The new RPC/status snapshot serialization path is performance-sensitive. Before promotion, it needs focused measurement that disproves status/render cost, or explicit owner approval for the tradeoff.

If this moves forward, please keep the public shape and replay lifecycle fixes narrow and include visible changelog credit.

nicobailon:
I re-reviewed this draft against the current async notification and RPC snapshot paths.

The direction is useful, but this head should stay draft/blocked. The current shape puts a snapshot-only object into `subagent-notify` message `details`, while the existing renderer treats any details object as the old `SubagentNotifyDetails` shape and reads fields like `resultPreview`. That makes persisted notification replay unsafe.

There is also no final-state synchronization boundary before the durable replay snapshot is captured, so a fast completion can persist an old in-memory child snapshot. The serializer has per-array limits but no global node or byte bound, and the branch also publishes a new package export / RPC capability before the lifecycle and payload contract are settled.

Recommended split: first rework the internal live snapshot path with a strict global payload budget and measurement; then add durable terminal replay from one authoritative final-state capture point with renderer-compatible details; publish the package export/docs only after that protocol is accepted. Any replacement/rework should keep visible credit for @yanqianglu / #1078.

nicobailon:
Thanks for the original draft and for pushing the structured async snapshot direction forward.

I landed a maintainer replacement in #1100 as `7306703`. It keeps the same accepted direction, but uses a bounded current-status snapshot only, keeps terminal result files canonical, and preserves the existing `subagent-notify.details` shape. The changelog credits @yanqianglu for #1078.

I am closing this draft as superseded by #1100.

PR #1079, opened 2026-08-13
fix: warn once when a mission leaves the store

albertgwo:
## Summary
- treat a missing mission record as terminal for that binding: warn once, then stop attempting further updates for it
- keep the existing per-child warning for every other failure
- credit the fix in `CHANGELOG.md` under Unreleased

`recordMissionWorkflowChild` warns on each failed update, and a long-running workflow child calls it on every heartbeat. When the mission record is gone, that becomes one warning per heartbeat for the life of the run:

```
[pi-subagents] Failed to record mission workflow child 'gate-packet-002': Mission '8cfcfb31-...' was not found i
 ⠸ gate-reviewer · step 1/1 · 33 tool uses · 4m59s
[pi-subagents] Failed to record mission workflow child 'gate-packet-002': Mission '8cfcfb31-...' was not found i
```

Because the warning goes straight to the console while the TUI is drawing, it also interleaves with the live run widget and truncates its lines, so the display becomes unreadable, as above.

The record can disappear under normal operation: `pruneTerminalMissions` retains the newest 200 terminal missions and runs on every `createMission` and `updateMission`, so an active run's mission can be pruned while its children still report. No later update can succeed once the record is gone, so retrying per heartbeat cannot recover anything.

## Validation
- rebased onto current `main` (v0.49.0); the guard sits alongside the newer `durableUpdate` destructuring
- `npm run typecheck`
- `npm run test:unit` (1895 pass, 2 skip). One unrelated failure, `watchdog-lsp-diagnostics` "returns a failed result for malformed language-server JSON", reproduces identically on clean `main` at v0.49.0 under the full suite and passes in isolation, so it is a pre-existing load-dependent flake rather than a regression from this change.
- harness replaying the guarded path over 350 heartbeats across two missing missions: 2 store attempts and 2 warnings, against 350 of each before

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Warn once, instead of on every heartbeat, when a long-running workflow child outlives its mission record. Thanks to @albertgwo for #1079.
+
 ## [0.49.0] - 2026-08-13
 
 ### Added
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -108,7 +108,7 @@ import { inspectSubagentStatus } from "../background/run-status.ts";
 import { applyForceTopLevelAsyncOverride } from "../background/top-level-async.ts";
 import { handleMissionAction, MISSION_ACTIONS } from "../../missions/actions.ts";
 import { attachMissionToLaunchResult, prepareMissionLaunch, writeMissionAsyncBinding, type MissionLaunchBinding } from "../../missions/lifecycle.ts";
-import { updateMission } from "../../missions/store.ts";
+import { MissionNotFoundError, updateMission } from "../../missions/store.ts";
 import type { MissionWorkflowChildUpdate } from "../../missions/types.ts";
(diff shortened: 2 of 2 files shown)


albertgwo:
Rebased onto current `main` (v0.49.0) and added the changelog credit, matching what you asked for on #1001. The conflict was only against the newer `durableUpdate` destructuring; the guard now sits above it.

Also confirmed the one failing unit test, `watchdog-lsp-diagnostics` "returns a failed result for malformed language-server JSON", fails identically on clean `main` at v0.49.0 under the full suite and passes in isolation, so it is pre-existing and unrelated to this change.

Thanks for landing #955 and #1001.

nicobailon:
Thanks @albertgwo. I opened #1085 as a current-main replacement that keeps this fix and the changelog credit, since I could not update the fork branch from this account.

nicobailon:
Thanks again @albertgwo. #1085 has landed with this behavior and visible changelog credit, so I am closing this superseded PR.

nicobailon:
Closed in favor of #1085, which landed the same missing-mission warning fix on current main with @albertgwo credited in the changelog.

PR #1080, opened 2026-08-13
feat: mirror subagent progress into Orca tabs

hyein-cbio:
## Summary

- add an experimental, explicitly opt-in Orca progress-tab observer for subagent children
- preserve native Pi and `external-cli` as the authoritative runners while mirroring live progress best-effort
- keep one tab per logical foreground/background/parallel/chain child, reuse it across startup/model retries, and retain the tab and scrollback after completion
- allocate unique persistent tab sequence numbers per Git worktree, including across concurrent processes
- show session cleanup guidance only for successfully completed native Pi children with a verified session header

## Design

The integration is enabled only with:

```json
{
  "orcaProgressTabs": {
    "enabled": true
  }
}
```

Orca-specific behavior is localized in `src/runs/shared/orca-progress-tabs.ts`. Existing execution, retry, cancellation, status, result, and artifact paths are unchanged. Missing Orca binaries, unavailable runtimes, tab creation failures, malformed optional metadata, and mirror I/O failures are isolated from child execution.

`external-cli` remains Orca-agnostic and exposes only generic stdout/stderr observation callbacks.

## Validation

- `PI_SUBAGENT_ORCA_BINARY=/definitely/missing/orca npm run test:all`
  - unit: 1,895 passed, 2 skipped, 0 failed
  - integration: 769 passed, 0 failed
  - e2e: 0 failed
- `npm run typecheck`
- `git diff --check`
- CodeRabbit CLI review: no findings

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -67,6 +67,22 @@ With `"summary"`, a tool result looks like this:
 ✓ reviewer · completed
 ```
 
+## `orcaProgressTabs` (experimental)
+
+```json
+{
+  "orcaProgressTabs": {
+    "enabled": true
+  }
+}
+```
+
+Opt in to a best-effort Orca observer that creates one Orca terminal tab for each subagent child and mirrors its live tool, assistant, stdout, and stderr progress. Tab titles use a persistent worktree-local sequence (`subagent · <agent> · 1`, `... · 2`, and so on), so separate workflows and concurrent children do not reuse the same number. This does **not** replace Pi as the child runner: native Pi children keep the same process, lifecycle, status, control, artifact, and result paths. External CLI profiles also keep their existing runner and can mirror their stdout/stderr.
+
+The integration is off by default. When enabled, `pi-subagents` looks for executable `orca` on `PATH`, or uses the executable path in `PI_SUBAGENT_ORCA_BINARY`. If no executable is available, Orca is not running, the cwd is not an Orca-managed worktree, or `terminal create` fails, the authoritative subagent still runs normally. Tab creation is deliberately best-effort and never changes the child result.
+
+Set `enabled` to `false` (or remove the block) as a kill switch. In that state, `pi-subagents` does not invoke `orca` and creates no Orca tabs. The temporary mirror files contain child output, use private file modes where supported, and are removed shortly after the child finishes. On completion, the viewer exits back to the Orca terminal's shell prompt; the tab and its terminal scrollback remain open until the user closes the tab. A successfully completed native Pi child with a recorded session ends with a command that can remove that session; failed, stopped, timed-out, and sessionless children do not show the removal command.
+
 ## `asyncByDefault`
 
 ```json
--- a/docs/tool-reference.md
+++ b/docs/tool-reference.md
@@ -315,6 +315,20 @@ The parser canonicalizes known enum synonyms, snake_case report keys and wrapper
 
 Acceptance fences are removed from normal output artifacts, while the raw child transcript remains intact and per-child metadata stores the complete acceptance ledger and parsed report. Explicit failed gates fail the run. Inferred gates remain observable without failing the run.
 
+## Orca progress tabs (experimental observer)
+
+Orca progress tabs are a global, opt-in observer, not an agent runner. Enable them in the extension config:
+
+```json
+{ "orcaProgressTabs": { "enabled": true } }
+```
+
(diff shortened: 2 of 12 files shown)


nicobailon:
Thanks @hyein-cbio. I reviewed this against the current backlog gate.

This is blocked for now on two points:

1. Raw child/tool/external-runner output reaches the Orca terminal without terminal-control sanitization. That matches the current Greptile security finding and needs a source fix before this can land.
2. The feature adds enabled-mode work on progress delivery paths: per-child Orca process creation, mirror writes, temp-file/state checks, and a 150 ms viewer poll. Under the repo’s current performance gate, that needs focused measurement that disproves a slowdown or explicit owner approval for the opt-in tradeoff before implementation or merge.

Also, any landed version needs visible changelog credit for your contribution.

nicobailon:
I did the focused performance/design pass for this PR.

The performance risk is not disproved. Enabled mode adds one Orca terminal plus one persistent polling Node viewer per observed child, duplicate output writes, synchronous setup/sequence allocation, and no observer backpressure bound. Since the feature is opt-in, this can move forward only if the owner explicitly accepts that resource model.

The smallest safe implementation path after that decision is to fix the terminal-control issue at the single viewer sink in `orca-progress-tabs.ts`, with a stateful sanitizer that handles CSI/OSC/control sequences even when they are split across poll reads. Sanitizing individual call sites would miss fragmented sequences.

So this remains blocked on two gates: owner acceptance of the enabled-mode resource model, then a focused sanitizer rework with tests and visible changelog credit for @hyein-cbio / #1080.

hyein-cbio:
Thanks for the focused pass. I agree that the terminal-control finding is valid, and that the sanitizer must be stateful at the single viewer sink so split CSI/OSC/control sequences are handled correctly.

I also agree that the enabled path necessarily performs additional work, so I do not think it would be honest to claim zero slowdown. One Orca terminal, one viewer per active observed child, and duplicate mirror output are inherent to the requested per-child-tab behavior. The 150 ms polling, synchronous setup, and unbounded observer buffering are implementation choices rather than inherent requirements, and I can quantify or mitigate those separately.

Before I rework the implementation, could you confirm whether the irreducible opt-in resource model is acceptable in principle? If it is, please also let me know whether you prefer the smallest path you outlined (stateful sink sanitizer plus focused tests and measurements of the current enabled mode), or whether acceptance should be conditioned on replacing polling and bounding observer backpressure first.

If the model is acceptable, I will also address the remaining review/process items in the same follow-up:

- use the exact verified session-file path in cleanup guidance;
- add the required changelog credit for @hyein-cbio / #1080;
- resolve the Windows coverage finding with portable fixtures, or an explicit platform gate if Windows is outside the supported Orca surface;
- rebase and run the focused and required validation on the final PR head.

If the per-child terminal/viewer model is not acceptable even as an explicit opt-in, I will close this PR rather than add more implementation churn.

nicobailon:
Thanks for the clear framing.

The per-child Orca tab/viewer model is acceptable in principle only if the enabled path is made performance-safe before merge. The current head is still blocked.

For the next revision, please treat these as required:

1. Add the stateful terminal-control sanitizer at the single viewer sink, so split CSI/OSC/control sequences cannot leak through.
2. Bound buffering/backpressure for mirrored output. No unbounded in-memory or file growth path should be introduced by slow viewers.
3. Remove the unbounded per-child polling cost, or provide focused measurements that prove the chosen polling model is cheap enough with several parallel children.
4. Keep disabled-mode overhead effectively zero and include focused measurements for disabled mode and enabled mode.
5. Use the exact verified session-file path in cleanup guidance.
6. Add the required changelog credit for @hyein-cbio / #1080.
7. Resolve the Windows coverage item with portable fixtures, or make the supported-platform boundary explicit if Orca support is intentionally non-Windows.

Once the branch is rebased and those points are covered, I will re-review the exact head.

hyein-cbio:
Thanks, that is clear. I’ll proceed with the performance-safe revision: remove per-child polling, bound observer buffering and mirror-file growth, keep disabled-mode gating outside the adapter, add the stateful viewer-sink sanitizer, and address the remaining cleanup-path, changelog, and Windows coverage items. I’ll rebase first and return with focused disabled/enabled measurements and exact-head validation.

nicobailon:
Thanks @hyein-cbio. I landed this through #1117 so the Orca progress-tab behavior could be rebased onto current `main` with the added safety fixes.

The replacement preserves the progress mirroring contribution and keeps visible credit for #1080 in the changelog.

PR #1084, opened 2026-08-14
fix: survive a schedule deleted while its timer is armed

albertgwo:
## Summary
- return early when a scheduled run fires for a schedule that no longer exists, instead of throwing from the store
- contain any other firing failure to that one schedule, so a timer callback can never exit Pi
- add `ScheduleStore.find`, a lookup that returns undefined rather than throwing; `get` keeps its throwing contract

Pi exits with an uncaught exception when a scheduled run fires for a deleted schedule:

```
pi exiting due to uncaughtException:
Error: Schedule 'topic-space-gate3-recertification-check' not found.
    at ScheduleStore.get (src/runs/background/scheduled-runs.ts:243:37)
    at ScheduledRunManager.fire (src/runs/background/scheduled-runs.ts:610:28)
    at Timeout.<anonymous> (src/runs/background/scheduled-runs.ts:603:61)
```

Schedules are project-scoped and shared, but `delete` clears the timer only inside the deleting process. Every other session holding an armed timer for that schedule then fires, throws from `store.get`, and dies. Because the throw happens in a timer callback it sits outside any caller's `try`/`catch`, so it reaches the process as an uncaught exception: deleting one schedule can take down every other session working on that project. A missing schedule has nothing to run and nothing to re-arm, so returning early is the correct behavior.

## Validation
- `npm run typecheck`
- scheduled-run unit tests (22 pass)
- harness reproducing the exact shape: firing for a deleted schedule rejects before the change and resolves cleanly after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Warn once, instead of on every heartbeat, when a long-running workflow child outlives its mission record. Thanks to @albertgwo for #1079.
+
 ## [0.49.0] - 2026-08-13
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Warn once, instead of on every heartbeat, when a long-running workflow child outlives its mission record. Thanks to @albertgwo for #1079.
+- Stop a deleted schedule from exiting Pi when another session still holds its armed timer. Thanks to @albertgwo for #1080.
 
 ## [0.49.0] - 2026-08-13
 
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks @albertgwo. I could not push to the fork branch, so I opened #1088 with the #1084 schedule deletion fix rebased onto current main and kept the changelog credit for you. I will leave this PR open until the replacement lands.

nicobailon:
Thanks again for this fix. I landed the maintained replacement in #1088, which preserves the deleted-schedule behavior and includes visible changelog credit for @albertgwo / #1084.

I’m closing this PR as superseded by the replacement so the queue has one source of truth.

PR #1094, opened 2026-08-14
feat: add FleetView default expansion

MattiasMTS:
## Summary

- add an opt-in `fleetViewDefaultExpanded` setting, preserving the compact default
- expand when a new active fleet appears while respecting manual collapse for the current fleet
- preserve lifecycle state across step transitions, widget suspension, and inspector ownership
- document and validate the setting with regression coverage

## Test plan

- `npm run typecheck`
- Node 24 unit suite: 1,939 passed, 0 failed, 0 cancelled, 2 skipped
- Node 24 integration suite: 788 passed, 0 failed
- E2E suite discovered successfully; runtime-dependent suite skipped because Pi runtime packages were unavailable
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- Add opt-in `fleetViewDefaultExpanded` configuration so active FleetView rosters can open automatically.
+
 ### Fixed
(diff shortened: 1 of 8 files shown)


MattiasMTS:
Closing because the interactive FleetView behavior did not work as intended.

PR #1131, opened 2026-08-15
fix: bound async tool-call index path segments

hlstwizard:
Use deterministic bounded path segments for oversized or non-portable tool-call IDs, and keep optional active-run aliases fail-open.

AI-assisted: commit prepared with AI help (model: gpt-5.6-sol)


nicobailon:
Thanks for the fix, @hlstwizard. I landed the bounded tool-call index behavior in #1134 on the current base, including changelog credit for your contribution.

I am closing this PR as superseded by #1134 so the queue has one source of truth.

PR #1137, opened 2026-08-15
fix: fall back from unavailable default fork context

hyein-cbio:
## Summary

Packaged `worker`, `oracle`, and `advisor` prefer `defaultContext: fork`. On a brand-new parent that does not yet have a persisted session file or current leaf, that implicit preference used to fail the first launch and force a retry.

This treats agent-level `defaultContext: fork` as a preference:

- if the parent session file and current leaf are available, launch a real fork
- otherwise start `fresh` immediately
- explicit `context: "fork"` stays fail-fast
- explicit `context: "fresh"` still wins

The same availability check is shared by execution and launch-contract preflight, so a contract no longer reports `context: "fork"` when the runtime would already fall back.

## Design

- `canPreferFork` / `canPreferForkFromSnapshot` live next to the real fork preconditions in `src/shared/fork-context.ts`
- implicit fork requires a session path, a current leaf, and a file that already exists on disk
- preflight emits the existing `host_required` diagnostic whenever the *resolved* context is `fork`, including implicit defaults

## Validation

- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test --test-name-pattern='falls back|keeps explicit fork fail-fast|uses agent defaultContext fork when launch|fails fast when context=fork' test/integration/fork-context-execution.test.ts`
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/unit/fork-context.test.ts test/unit/preflight.test.ts test/unit/schemas.test.ts`
- `coderabbit review --base-commit HEAD~1` after the follow-up commit: no new findings


nicobailon:
Thanks for the fix, [@hyein-cbio](https://github.com/hyein-cbio). I landed this through #1140 as a maintainer replacement so it could be rebased onto the current queue and include the required changelog credit.

The replacement preserves your authored commits and the behavior from this PR: implicit `defaultContext: "fork"` now falls back to `fresh` when the parent session cannot be forked, while explicit `context: "fork"` remains fail-fast.

Closing this PR as superseded by #1140 / 81fb6894acd1d2b70570e2184731fe028385dc4c.

PR #1151, opened 2026-08-15
fix: accept no-edit challenge passes

### LINKED ISSUE #1110: No-edit retained-writer challenges should not deliver generic failure notifications
During backlog maintenance, a retained-writer implementation challenge correctly made no edits and returned a useful report, but the runtime still delivered it as a failed subagent notification:

```text
subagent failed

worker failed in run d5d1976d.

Subagent failed: worker
Run: d5d1976d step 1
Signal: worker completed without making edits for an implementation task
```

For implementation challenge loops, a no-edit pass can be a valid result when the retained writer confirms the current candidate is already the smallest correct shape. The parent should still be able to inspect the report and continue to fresh review, but the user-facing delivery should not look like a hard worker failure.

Expected behavior:

- A no-edit retained-writer challenge result should be reported as an implementation-challenge/no-change outcome, not as a generic `subagent failed` delivery.
- The report should preserve the no-edit explanation and validation evidence.
- Real implementation tasks that accidentally return only planning text should still fail.
- The behavior should distinguish challenge/follow-up context from ordinary implementation tasks.

Observed context:

- Run: `d5d1976d`
- Challenge output said the writer kept the current implementation, explained why no additional edits were needed, and reran the focused regression.
- The runtime still emitted `worker completed without making edits for an implementation task` as a failure notification.

nicobailon:
Fixes #1110.

This keeps explicit no-edit retained-writer challenge passes successful when the follow-up allowed no file changes and the child reports `Implemented no further changes.` Ordinary implementation tasks with no edits still fail the mutation guard.

Validation on rebased head `ef0dbfc9`:
- `node --experimental-strip-types --test test/unit/completion-guard.test.ts`
- `npm run typecheck`
- `git diff --check HEAD^..HEAD`

Fresh exact-head review is being rerun after the final rebase.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### Fixed
 - Skip fallback models that are unavailable in the active registry, so shared agent configs still run where their primary model is available. Thanks to [@JPFrancoia](https://github.com/JPFrancoia) for #1147.
 - Sweep expired wait subscriptions armed by another session, so a record left behind by a session that never returns stops accumulating in the subscriptions directory. Thanks to [@MarcusNeufeldt](https://github.com/MarcusNeufeldt) for #1142.
+- Keep explicit no-change retained-writer implementation challenge passes successful instead of reporting a missing-edit failure (#1110).
 - Keep `mcp:<server>` direct tools available when pi-mcp-adapter cache identity includes a request-header command. Thanks to [@xz-dev](https://github.com/xz-dev) for #1141.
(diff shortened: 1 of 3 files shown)


nicobailon:
Closing this rather than landing it.

The bug report came from our backlog-maintainer challenge loop, and this PR encoded that workflow's challenge wording into `pi-subagents` core. That is too specific for the package contract.

For now, I moved the handling to the backlog-maintainer process: when a retained-writer challenge explicitly says no changes are needed and the runtime reports the generic missing-edit failure, treat the child report as the challenge result and continue to review. If we later want core support, it should be designed as a generic explicit-no-edit task contract, not tied to issue/backlog prompt wording.

PR #1154, opened 2026-08-15
fix: pass trusted session roots to fleet transcripts

aliceisjustplaying:
## Summary
- pass current-session and tracked-job session roots into async fleet transcript rendering
- allow trusted session JSONL fallback to render instead of showing a false “without a trusted root” warning
- add regression coverage at the fleet component seam

## Root cause
The fleet filtered runs to the current session correctly, but `asyncDetail()` dropped `state.trustedSessionRoots` when calling `formatAsyncRunTranscript()`.

## Tests
- `npm test` (2026 passed, 3 skipped)
- `npm run typecheck`
- `git diff --check`


nicobailon:
Thanks for the fix. I landed the replacement in #1174 and kept your credit in the changelog, so I am closing this superseded PR.

PR #1167, opened 2026-08-15
fix(schedules): tolerate a project dir deleted after session bind

colinb4987:
## Problem

`assertScheduleRoot` (`src/runs/background/scheduled-runs.ts:150`) calls `fs.realpathSync(projectCwd)` unconditionally. When a pi session stays bound to a project directory that gets deleted (e.g. a repo is removed mid-session), every extension rebind — `/clear`, `newSession`, `finishSessionReplacement` — throws a raw `ENOENT` out of `bindSession → selectProject → restore`, which kills the **entire extension**, not just schedule restore.

Repro:
1. Open pi in a directory, then delete the directory from another shell.
2. `/clear` (or any rebind).
3. Extension dies with `ENOENT: no such file or directory, lstat '<deleted dir>'` at `assertScheduleRoot`.

## Fix

Treat a missing project cwd as "no schedules": catch `ENOENT` on the realpath and return early. `ids()` then sees a missing root and returns `[]`, so bind-time restore is a no-op instead of a crash. All subsequent `realpathSync` calls in the function only run when the project dir resolved successfully.

## Test

Added a regression test that writes a schedule record into a temp project, deletes the project, then asserts `listScheduledRunSummaries(deletedCwd)` returns `[]` rather than throwing — this exercises the exact `restore → list → ids → assertScheduleRoot` path that crashed.

Full unit suite: 2029 tests, 2015 pass. The single failing test (`herdr-inspector.test.ts`, "event loop has already resolved") is pre-existing on a clean tree (2028 tests / 2014 pass) and unrelated to this change.

### DIFF
--- a/src/runs/background/scheduled-runs.ts
+++ b/src/runs/background/scheduled-runs.ts
@@ -152,7 +152,17 @@ function assertScheduleRoot(root: string, projectCwd: string | undefined, create
 		if (create) fs.mkdirSync(root, { recursive: true, mode: 0o700 });
 		return;
 	}
-	const projectPath = fs.realpathSync(projectCwd);
+	let projectPath: string;
+	try {
+		projectPath = fs.realpathSync(projectCwd);
+	} catch (error) {
+		// The project directory no longer exists (e.g. it was deleted while a
+		// session bound to it stays open). Treat it as having no schedules:
+		// ids()/list() then see a missing root and return empty, and the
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the deleted-project schedule restore fix. I landed the combined current-main version in #1175, with the behavior preserved and changelog credit kept for you. Closing this PR as superseded by #1175.

nicobailon:
Superseded by #1175.

PR #1171, opened 2026-08-16
Skip orphaned schedule directories so schedule restore cannot crash the extension

ELA718:
## Problem

`ScheduleStore.list()` did `ids().map(get)`, and `get()` throws when a schedule directory has no `schedule.json`. A directory can lose only its `schedule.json` — for example a git checkout that removes a file the repository previously tracked while leaving the untracked `events.jsonl` / `history.json` / `runs/` behind. Every session start then crashed in `ScheduledRunManager.restore` → `bindSession` with:

```
error: Schedule '<id>' not found.
  at ScheduleStore.get
  at ScheduleStore.list
  at ScheduledRunManager.restore
  at ScheduledRunManager.selectProject
  at ScheduledRunManager.bindSession
```

## Fix

`list()` now maps through the existing `find()` helper and skips directories whose `schedule.json` is gone. `get()` still throws for explicit lookups, so `schedule.show` / `schedule.delete` keep their not-found errors.

## Test

New unit test creates an orphaned schedule directory and asserts `bindSession` + `listScheduledRunSummaries` tolerate it. Fails on main with the exact production error, passes with the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [Unreleased]
+### Fixed
+- Keep schedule restore from crashing the extension at session start when a project has an orphaned schedule directory whose `schedule.json` is gone (for example after a git checkout that previously tracked it), by skipping orphaned directories instead of throwing.
 
 ### Added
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for the orphaned schedule-directory restore fix. I landed the combined current-main version in #1175, with the behavior preserved and changelog credit kept for you. Closing this PR as superseded by #1175.

nicobailon:
Superseded by #1175.

PR #1117, opened 2026-08-14
feat: mirror subagent progress into Orca tabs

nicobailon:
Supersedes #1080.

This preserves @hyein-cbio's Orca progress-tab contribution and adds the safety work needed to land it on current `main`.

The implementation keeps Orca opt-in, mirrors child progress into bounded private files, sanitizes terminal output before display, applies backpressure at 1 MiB, emits an exact cleanup command only for a completed Pi session, and limits support to macOS/Linux. It also covers native Pi children and external CLI children.

Follow-up fixes keep observer failures from delaying subagent completion, use a detached watchdog for hung `orca terminal create`, isolate Orca config loading from host-peer runtime imports, support standalone Pi executable host shapes via the existing Node resolver, use portable Node command fixtures for Orca tests, and ensure completed mirrors are eventually cleaned by a detached cleanup watcher.

Validation at `f229b51`:
- `test/unit/host-peer-runtime-imports.test.ts`
- `test/unit/orca-progress-tabs.test.ts`
- `test/integration/orca-progress-tabs.test.ts`
- `test/integration/external-cli-runner.test.ts`
- `test/unit/pi-coding-agent-dir.test.ts`
- `npm run typecheck`
- `git diff --check origin/main...HEAD`

Review gate:
- Fresh exact-head review at `f229b51` found no issues.

Known gap:
- Real-Orca enabled-mode measurement was explicitly owner-approved as not required for this landing.

PR #1196, opened 2026-08-16
fix: serialize Orca progress-tab creation order

hyein-cbio:
## Summary
- serialize same-worktree `orca terminal create` so numbered progress tabs appear left to right (`1`, then `2`, then `3`)
- keep child execution non-blocking; only the optional observer waits for the previous create
- wait only while the immediate predecessor is still pending, and skip pruned/missing markers immediately
- time out a stalled predecessor so later tabs are not stuck
- add unit and concurrent-process coverage for create order, missing markers, and queued predecessors

Follow-up to #1080 and #1117. Sequence numbers were already unique, but fire-and-forget creates could still land out of order in the Orca tab strip.

## Validation
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/orca-progress-tabs.test.ts`
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/orca-progress-tabs.test.ts`
- `npx tsc --noEmit --pretty false`
- `git diff --check`

## Notes
This does not add a create-position flag to Orca. The observer just stops racing `terminal create` for the same worktree.


nicobailon:
Thanks for this fix. I could not push the small follow-up to this fork branch even though GitHub reports maintainer edits are enabled, so I published a replacement PR that preserves your commits and credit: #1198.

The follow-up fixes the fresh-review blocker where queued same-worktree Orca create watchdogs could all age out from their spawn time and overlap after the 25s fallback. #1198 keeps the same scope and adds the regression.

PR #1205, opened 2026-08-16
feat: add explicit non-isolated subagent mode

### LINKED ISSUE #1203: Support an explicit `isolation: "none"` mode for non-Git directories
## Problem

I want to run subagents in ordinary directories that are not Git repositories, such as:

- documentation directories
- Obsidian vaults
- temporary research directories
- configuration directories
- new projects before Git initialization

A normal subagent run appears to support non-Git directories when `isolation` is omitted.

However, the Agent tool schema only exposes one explicit isolation value:
```json
{
  "isolation": "worktree"
}
```
Some schema-driven clients, wrappers, and LLM tool callers materialize this optional single-value field automatically. This makes the agent attempt worktree isolation and fail with:

```text
Cannot run with isolation: "worktree" — not a git repo, no commits yet, or `git worktree add` failed.
Initialize git and commit at least once, or omit `isolation`.
```
The instruction to “omit isolation” is difficult to enforce reliably when the tool schema has only one possible explicit value.

## Request

Could pi-subagents support an explicit non-isolated mode?

For example:

```json
{
  "isolation": "none"
}
```

Desired semantics:

- `"none"`: run directly in the requested `cwd`, including non-Git directories.
- `"worktree"`: retain the current strict Git worktree behavior.
- omitted: preserve the current default behavior for backward compatibility.
- Do not silently fall back from `"worktree"` to `"none"` when worktree creation fails.

Possible schema:

```ts
isolation: Type.Optional(
  Type.Union([
    Type.Literal("none"),
    Type.Literal("worktree")
  ])
)
```

Internally, `"none"` could normalize to the existing non-worktree execution path.

This would make non-Git execution explicit and deterministic instead of relying on omission semantics.

It would also help:

- structured tool clients that require explicit values
- LLMs that repeatedly populate the only available enum value
- cross-extension RPC callers
- users who want subagents for non-code or pre-Git work

nicobailon:
## Summary

Adds explicit `isolation: "none"` for schema-driven clients that need to request shared-cwd execution without relying on omission semantics. `isolation: "worktree"` maps to the existing strict managed Git worktree path and does not fall back.

Closes #1203.

## Validation

- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/public-execution.test.ts test/unit/schemas.test.ts`
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test --test-name-pattern='runs isolation none outside Git and keeps worktree isolation strict|runs a direct single child in a managed worktree|inherits workflow-level worktree isolation and allows a child opt-out' test/integration/single-execution.test.ts`
- `npm run typecheck`
- `git diff --check`

## Credit

Thanks to [@tlsneo](https://github.com/tlsneo) for #1203.


tlsneo:
👍

Issue #1190, opened 2026-08-16
Workflow worker can fail before launch when acorn is unavailable


nicobailon:
While using the subagent tool from surf-cli, a simple async fresh-context reviewer launch failed before the child started with:\n\n```\nWorkflow worker failed: Cannot find module 'acorn'\nRequire stack:\n- /Users/nicobailon/Documents/development/surf-cli/[worker eval]\n```\n\nThe parent request was a normal single-agent review launch. The failure appears to come from the workflow worker resolving its parser dependency from the project cwd instead of a bundled/runtime location. Expected behavior: the worker should have its own parser dependency available, or fail with an actionable setup message before accepting the launch.\n\nImpact: it blocked the required fresh-context review lane, so I had to use an external Codex review fallback.

PR #1214, opened 2026-08-17
fix: fall back to manifest resolution for workflow parser entry

xz-dev:
## Summary

`runWorkflowScript` resolves its parser via `createRequire(import.meta.url).resolve("acorn")`. That call throws `MODULE_NOT_FOUND` under **Bun-compiled single-file Pi binaries** even when `node_modules/acorn` is present right next to `package.json`, so every `workflowScript` launch fails before the worker starts with:

```
Workflow parser dependency 'acorn' is unavailable from pi-subagents. Reinstall pi-subagents dependencies before launching workflowScript.
```

Reinstalling cannot help — the dependency was never missing. Observed under a Pi `pi-native` binary reporting `{ node: "v24.3.0", bun: "1.3.14" }`:

| resolution attempt | result |
| --- | --- |
| `createRequire(import.meta.url).resolve("acorn")` | ❌ `MODULE_NOT_FOUND` |
| same with explicit `{ paths: [pkgDir] }` | ❌ `MODULE_NOT_FOUND` |
| `resolve("acorn/package.json")` | ✅ works |
| `require("<abs path>/node_modules/acorn/dist/acorn.js")` | ✅ works |
| `await import("acorn")` | ✅ works |

Bare package-specifier resolution through `createRequire` is broken in that runtime while subpath resolution still works.

## Fix

Add `resolveWorkflowParserEntry()`: keep the existing bare-specifier resolution as the primary path, and on failure fall back to resolving `acorn/package.json` and deriving the CommonJS entry from the manifest's `main` field. If the manifest is unavailable too, the original error is rethrown so the actionable setup message is unchanged for genuinely missing dependencies.

## Validation

- `tsc --noEmit` clean
- `test/unit/scripted-workflow.test.ts`: 71/71 pass
- End-to-end in the affected runtime (Bun-compiled Pi binary, extension loaded from source): `runWorkflowScript` with top-level `await`, `Promise.all`, `console.log` capture and `emit()` completes and returns the expected value; before the patch the same probe reproduced the exact reported error.

Refs #1190.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@
 - Resolve `/subagents-generate-profiles` provider probes through the shared Pi executable resolver so configured and Windows-specific Pi commands work. Thanks to [@Wumpf](https://github.com/Wumpf) for #1199.
 - Serialize same-worktree Orca progress-tab creation so numbered tabs appear left to right in sequence instead of racing in the UI. Thanks to [@hyein-cbio](https://github.com/hyein-cbio) for #1196.
 - Route Fleet inspector steering for live in-process workflow children through their foreground routes instead of the detached async queue. Thanks to [@ViktorBarzin](https://github.com/ViktorBarzin) for #1218 and #1216.
-- Resolve the workflowScript parser from pi-subagents instead of the caller's working directory, so workflows start in projects that do not install Acorn.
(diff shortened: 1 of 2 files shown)


nicobailon:
Resolved the Greptile credit thread. The current head now credits [@xz-dev](https://github.com/xz-dev) for #1214 and references #1190 in `CHANGELOG.md`.

Issue #1217, opened 2026-08-17
Fleet inspector steering dead-letters for live in-process workflows


ViktorBarzin:
# Fleet inspector steering dead-letters for live in-process workflows

## Affected version

`pi-subagents` 0.48.0.

## Reproduction

1. In one Pi session, launch an async workflow with two live children.
2. Open `/subagents-fleet` while both children are running.
3. Select the workflow row.
4. Press `s`, enter a message, and press `Enter`.

## Observed behavior

The fleet inspector reports the request as queued/successful, but neither child receives it. In a reproduction on 2026-08-17, the inspector wrote this request under `<asyncDir>/control/steer-requests/`:

```json
{"type":"steer","id":"6bb9d5ae-a80b-4ce9-9c24-efe5b5cf3fc0","ts":1786984374767,"message":"test steering","targetIndexes":[0,1],"source":"steer-action"}
```

The file remained unconsumed. The workflow directory had no root-level `steer-capabilities/` or `steer-acks/`, and the children did not log or receive the message.

## Expected behavior

The inspector should route the message through the live workflow's foreground steering channel and report the resulting acknowledgment. Steering for detached runs should continue using the async control-file channel.

## Root cause

The fleet inspector's default steer handler always calls `steerAsyncRun` (`src/tui/fleet.ts:1165-1173` in the npm-shipped 0.48.0 source). That function writes to `<asyncDir>/control/steer-requests/` through `requestAsyncSteer` (`src/runs/foreground/async-steering-action.ts:118-127`; `src/runs/background/control-channel.ts:314-347`).

Only detached runner processes consume that queue and route requests to child inboxes (`src/runs/background/subagent-runner.ts:2651-2700`). In-process workflow children instead register foreground controls and use per-child routes under `control/workflow-foreground/<childRunId>/`.

The `subagent` tool already distinguishes these cases. Its steer action resolves workflow or foreground targets with `resolveWorkflowForegroundSteeringTarget` and sends through `steerWorkflowForegroundTarget` (`src/runs/foreground/subagent-executor.ts:5494-5543`). This path writes to the live child's foreground inbox and waits for its acknowledgment (`src/runs/foreground/workflow-foreground-steering.ts:47-160`).

## Fix sketch

In the fleet inspector's default steer handler:

1. Read the selected run's status.
2. If it is a workflow with a live in-process workflow controller, resolve the foreground target and call `steerWorkflowForegroundTarget`.
[...]

PR #1218, opened 2026-08-17
Fix fleet steer routing for in-process workflow children

ViktorBarzin:
# Fix fleet steering for live workflow runs

## What and why

The fleet inspector currently sends every steer action through the detached-run control-file channel. For a live workflow whose children run in the extension process, that queue has no consumer, so the inspector can report a queued request that never reaches a child.

This change routes a workflow row through the foreground steering path when its workflow controller is live. Other runs continue using `steerAsyncRun`.

## Root cause

The default handler in `src/tui/fleet.ts` calls `steerAsyncRun` without checking whether the selected async directory represents an in-process workflow. `steerAsyncRun` writes to `control/steer-requests/`, which detached runner processes consume.

Live workflow children use `control/workflow-foreground/<childRunId>/` and are reachable through `resolveWorkflowForegroundSteeringTarget` plus `steerWorkflowForegroundTarget`. The `subagent` tool's steer action already uses this distinction; the fleet handler did not.

## Scope

- Detect workflow status only when a matching live workflow controller exists.
- Reuse the foreground resolver and acknowledged delivery path.
- Preserve the existing detached-run path.
- Do not change fleet inspector controls or input behavior.

## Test notes

- Manually verified that fleet steering reached live workflow children and appeared in their logs.
- Manually verified that detached-run steering still used the existing async channel.
- Verified that the patch applies cleanly to the npm-shipped 0.48.0 source. The npm artifact does not ship `test/`; run `npm test` in the upstream checkout.


ViktorBarzin:
Related: #1217.

PR #1222, opened 2026-08-18
fix: isolate extension cleanup across concurrent Pi sessions

ryanbbrown:
## Summary

- scope extension cleanup to the owning Pi `AgentSession`
- preserve cleanup and visible control-notice deduplication across reloads of the same session
- keep independent sessions' identity, event subscriptions, timers, parent-session environment ownership, and slash snapshots isolated
- add public-boundary regression coverage for concurrent sessions and real Pi reload order

## Root cause

`registerSubagentExtension()` stored one cleanup callback and event-unsubscribe list on `globalThis`. Registering any second extension instance in the same Node process ran the first instance's cleanup. That set the first instance's `currentSessionId` to `null`, so its unchanged headless `agent_end` auto-drain failed with `Cannot auto-drain background work without an active session identity.`

The new global registry binds live runtime entries to `ctx.sessionManager`. Independent `AgentSession` objects use different keys. A reload of the same session replaces only the entry for that session manager and uses one complete, idempotent cleanup path.

## Validation

- Regression test failed before the fix with the exact missing-session-identity error.
- Focused extension tests: 25 passed, 0 failed.
- `npm run test:integration`: 670 passed on rerun.
- `npm run test:e2e`: command passed; the runtime suite was skipped because real Pi runtime packages are unavailable in the local fixture setup.
- `npm run test:unit`: 2,103 passed, 1 unrelated temporary-directory teardown race, 10 existing Herdr inspector cancellations, and 3 skipped. The teardown test passed alone; the Herdr cancellations reproduced outside this change.
- `git diff --check`: passed.

## Remaining risk

The registry assumes one stable `sessionManager` object per Pi `AgentSession`, which matches Pi's current lifecycle. `PI_SUBAGENT_PARENT_SESSION` remains a process-wide compatibility fallback, but shutdown now deletes it only when the value still belongs to that runtime; current launch paths pass the owning parent session explicitly.


nicobailon:
Thanks @ryanbbrown. I landed this through maintainer replacement PR #1237, rebased onto current `main` and adjusted to preserve the later demand-gated runtime hooks and completion-owner behavior.

Your original patch direction is credited in `CHANGELOG.md`. I am closing this draft as superseded by #1237.

PR #1226, opened 2026-08-18
🤖: Default Herdr panes to no focus

boggylp:
🤖

## Summary

Unattended project and inspector panes moved Herdr focus unless the caller remembered to opt out. Subagents open those panes on their own and omit `focus`, so the disruptive case was the default. Both paths now pass `--no-focus` and take focus only on an explicit `focus: true`.

- Keep persisted focus timestamps aligned with actual focus requests.
- Request focus explicitly for the FleetView inspect key, the one caller whose pane open is a deliberate keypress.
- Correct the `focus` default in the tool reference, which still documented the old focused default.

## Validation

- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/herdr-inspector.test.ts test/unit/fleet.test.ts`
- `npm run typecheck`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Add `PI_SUBAGENT_FS_RETRY_MAX_TOTAL_MS` so a host embedding the extension can bound how long a contended filesystem retry blocks its thread. Unset by default. Thanks to [@MarcusNeufeldt](https://github.com/MarcusNeufeldt) for #1143.
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


nicobailon:
Thanks again @boggylp. I landed this through maintainer replacement PR #1232 so the branch could be rebased onto current `main` and the changelog conflict resolved while preserving the behavior and changelog credit for #1226. I am closing this PR as superseded.

nicobailon:
Closed as superseded by #1232, which preserves the contribution and credits @boggylp in the changelog.

PR #1235, opened 2026-08-18
fix: sanitize runId in foreground subagent executor output paths

maxime-louward-shift:
## Description

When subagents are dispatched (such as via foreground workflows or MCP tool calls), the `runId` passed to `resolveSingleRunOutputBaseDir` and `workflowChildDefaultOutput` in `subagent-executor.ts` can be the caller tool-call ID (e.g. `call_VfdHQygxGeL1L49ez04A4tf7|WtnJ9gVpB...`).

On Windows, these IDs often contain characters that are illegal in file/directory paths (such as `|`, `+`, `/`, `\`), or base64 segments that cause deep directory nesting or exceed path lengths, resulting in `ENOENT: no such file or directory, mkdir` errors when a subagent attempts to write its output to `subagent-artifacts/outputs/...`.

This PR adds `sanitizeRunPathSegment()` in `subagent-executor.ts` to ensure `runId` and `workflowRunId` path components are cleaned before joining into artifact output directory paths.

## Note for Maintainer
Maintainers might also want to review whether other areas where raw `toolCallId` / external `runId` values are joined into filesystem paths on Windows (across other background or nested runner handlers) need similar sanitization or hashing.

### DIFF
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -2861,8 +2861,11 @@ function resolveConfiguredSingleRunOutputBaseDir(deps: ExecutorDeps): string | u
 		: undefined;
 }
 
-function sanitizeRunPathSegment(value: string): string {
-	return value.trim().replace(/[^A-Za-z0-9._-]+/g, "_").replace(/^_+|_+$/g, "") || "unknown";
+export function sanitizeRunPathSegment(value: string, maxBytes = 120): string {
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks @maxime-louward-shift. I landed this through maintainer replacement PR #1238, rebased onto current `main` and with changelog credit added for your contribution.

I am closing this PR as superseded by #1238.

PR #1274, opened 2026-08-19
fix: hide console windows for spawned bookkeeping processes

MarcusNeufeldt:
## Problem

On Windows, running a workflow makes console windows flash up and steal focus. Under a busy fan-out or a test run it is a steady stream, which makes the machine hard to work on while pi-subagents is doing anything.

The cause is that `child_process` defaults `windowsHide` to `false`. When the spawning process has no console of its own, Windows allocates a **new console window** for each child. pi-subagents spawns a lot of short-lived helpers purely for bookkeeping, so each one flashes a window.

This is most visible when the host runs headless — for example a long-lived server process started by a Windows Scheduled Task, which has no console to inherit. In an interactive terminal the children inherit that console and nothing is visible, which is why this is easy to miss.

## What this changes

Adds `windowsHide: true` to the thirteen spawn sites that can run on Windows:

| File | Sites | Spawned for |
| --- | --- | --- |
| `src/missions/workflow-state.ts` | 1 | PowerShell process-liveness probe |
| `src/runs/background/async-retention.ts` | 1 | PowerShell process-liveness probe |
| `src/runs/background/subagent-runner.ts` | 2 | `gh auth status`, `gh gist create` |
| `src/runs/shared/acceptance.ts` | 4 | `git status`, `git rev-parse` ×2, `git diff` |
| `src/runs/shared/worktree.ts` | 2 | `git -C …`, acceptance hook |
| `src/watchdog/change-signature.ts` | 1 | `git -C …` |
| `src/watchdog/lsp-diagnostics.ts` | 1 | language server |
| `src/workflows/chat-progress.ts` | 1 | `git -C …` |

The package already uses this flag for its orca watchdog spawns, so this only makes the treatment consistent with existing practice rather than introducing a new convention.

## Deliberately not changed

- **POSIX-only probes**, where the flag would be inert: the `ps` / `/bin/ps` calls in `workflow-state.ts`, `async-retention.ts`, `owned-process-tree.ts`, `session-lease.ts` and `external-job-bridge.ts`.
- **`orca-progress-tabs.ts`** — already sets the flag on all of its spawns.
- **Test files.** There are ~124 unhidden spawn sites under `test/`, which is why running the suites on Windows is noisy. I left them alone to keep this diff to runtime behaviour; happy to do them as a follow-up if you would take it.

`windowsHide` only affects whether a console window is created. It does not change stdio, exit codes, or process trees, so there should be no behavioural change beyond the windows not appearing.

## Testing

- Targeted run of the suites covering every touched module (`acceptance`, `async-retention`, `watchdog-change-signature`, `watchdog-lsp-diagnostics`, `workflow-chat-progress`, `worktree`, `orca-progress-tabs`, `session-lease`, `owned-process-tree`): **148 tests, 122 pass, 3 fail**, identical to the same selection on unmodified `main`. The three failures are pre-existing in my environment and unrelated (`acceptance` host-verification cases).
- Re-ran that selection three times to confirm stability; one earlier run also failed `returns a failed result for malformed language-server JSON`, which passes in isolation both with and without this change and passed on all three re-runs. It looks timing-sensitive under load rather than affected here.

I did not run the full suites for this change, since the point of it is that those runs are disruptive on Windows.

## Note

The same pattern exists in the pi host packages themselves, which I am reporting separately upstream — this PR only covers pi-subagents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Improve bundled role and parent prompts for source-first discovery in noisy codebases (#1247).
 
 ### Fixed
-- Stop Windows opening a console window for each child process spawned for bookkeeping. `spawnSync` and `execFileSync` default to `windowsHide: false`, so the Git, `gh` and PowerShell probes used for process liveness, acceptance snapshots, worktree commands, chat progress and LSP diagnostics each flashed a console window on Windows, which made a busy run disruptive to work alongside. POSIX-only probes are left untouched. Thanks to [@MarcusNeufeldt](https://github.com/MarcusNeufeldt).
+- Stop Windows opening a console window for each child process spawned for bookkeeping. `spawnSync` and `execFileSync` default to `windowsHide: false`, so the Git, `gh` and PowerShell probes used for process liveness, acceptance snapshots, worktree commands, chat progress and LSP diagnostics each flashed a console window on Windows, which made a busy run disruptive to work alongside. POSIX-only probes are left untouched. Thanks to [@MarcusNeufeldt](https://github.com/MarcusNeufeldt) for #1274.
 - Keep completed inspect RPC output available from the durable completion replay after result delivery consumes its one-shot payload, including per-child inline result tails (#1254).
 - Wake the idle parent when an async workflow child needs attention, and persist that control event on the enclosing workflow. Status already showed the stall; the parent notice did not. Thanks to [@Yibo-Zhang](https://github.com/Yibo-Zhang) for #1266.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Improve bundled role and parent prompts for source-first discovery in noisy codebases (#1247).
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


nicobailon:
Went through this adversarially and it holds up. I independently inventoried every child_process call site in src/ — your count of thirteen is right, the POSIX-only exclusions are sound (those probes either sit behind a platform branch or ENOENT on Windows before a window can exist), and every hunk is a pure options addition with no stdio/timeout/arg changes.

Two sites you missed that can still flash on Windows: the `execSync("npm root -g", ...)` calls in `src/agents/agents.ts` and `src/agents/skills.ts`. `execSync` goes through cmd.exe, so same headless scenario applies. Both are cached so it's one flash per process, not a stream — fine to fold in here or leave for the follow-up you offered.

Only real blocker is the changelog conflict — a couple of entries landed since you branched, so this needs a rebase onto main. Happy to merge once that's resolved.

nicobailon:
Landed via #1275 with your commits and authorship preserved (merge commit 750a7d978). I folded in the two `execSync("npm root -g")` sites from my earlier comment so the follow-up isn't needed. Thanks for the careful writeup — the site inventory and POSIX exclusions all checked out exactly as described.

PR #1313, opened 2026-08-20
feat: add agent output mode defaults

### LINKED ISSUE #1305: Feature: per-agent default for outputMode (frontmatter or agentOverrides)
## Problem

`outputMode: "file-only"` exists and works (shipped in #132). Nothing lets an agent author set it as that agent's default. Every caller must remember to pass it per call, and the documentation barely mentions it.

Confirmed on v0.49.0:

- `agents.ts:1661` maps `frontmatter.output` into the agent config. No sibling line maps `frontmatter.outputMode`. The frontmatter parser swallows an `outputMode:` key in an agent's YAML into `extraFields` without an error. It does nothing.
- `settings.ts:309` and `settings.ts:492` both resolve the mode as `stepOverrides.outputMode ?? "inline"` / `task.outputMode ?? "inline"`. The default is hardcoded at the step-resolution layer, not read from agent config.
- `subagentsObject.agentOverrides` parsing (`agents.ts`, `parseBuiltinOverrideEntry`) accepts `description`, `model`, `thinking`, `systemPromptMode`, `inheritProjectContext`, `inheritSkills`, `defaultContext`, and a few more fields. It has no `outputMode` field either. `settings.json` cannot set this default.
- `execution.ts:253` already drops `result.messages` when `outputMode === "file-only"` and `savedOutputPath` is set. The mechanism a default would rely on is already built. Only the default itself is missing.

This matters more than it looks because the fallback isn't a smaller inline payload, it's an uncapped one. `execution.ts:1716` gates truncation on `if (options.maxOutput)`, and `maxOutput` was removed from the LLM-facing tool schema per the changelog. A caller that forgets `outputMode: "file-only"` on a research or recon agent gets the full report back with no cap at all.

## What we tried as a workaround

We shipped two user-level agent overrides (`scout`, `researcher`) with rewritten `description:` text asking the calling model to request the file-only mode itself, since the tool schema (`schemas.ts`) does expose `outputMode` as a parameter the driver can pass. In one live test (n=1, not a benchmark), the driver read the guidance, passed `outputMode: "file-only"`, and the parent received only the pointer line. That's the best case: it depends on the calling model reading and following a paragraph of agent metadata, call after call, forever. It's not a substitute for a default that can't be skipped.

## Ask

A per-agent default for the output mode, in whichever place already carries per-agent config:

- Frontmatter: `outputMode: file-only` alongside the existing `output:` key, mapped the same way `frontmatter.output` is mapped today.
- Or `agentOverrides.<name>.outputMode` in `settings.json`, next to the existing `model` and `thinking` override fields.

Either would let an agent author decide once, at the agent definition, that its output is a reference-sized artifact by default, without asking every caller to remember a parameter most callers never read the docs for.

nicobailon:
## Summary

Adds per-agent `outputMode` defaults from agent frontmatter and `subagents.agentOverrides`, while keeping call-level `outputMode` as the highest precedence. This closes #1305.

The change also includes focused parsing, override, preflight, and chain behavior coverage. Thanks to [@bbbRye007](https://github.com/bbbRye007) for #1305.

## Validation

- focused outputMode and preflight tests: 166 passing
- `npm run typecheck`
- `git diff --check origin/main...HEAD`
- fresh exact-head review: no issues found

Closes #1305.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 ### Added
+- Let agent definitions and `agentOverrides` set a default `outputMode`, while
+  call-level output mode stays higher priority. Thanks to [@bbbRye007](https://github.com/bbbRye007) for #1305.
 - Add `context: "profile"` for workflow children that should use the selected
   agent profile's declared context instead of the global default (#1303).
 - Add durable keyed async workflow receipts and resume-by-key selectors for
--- a/src/agents/agent-management.ts
+++ b/src/agents/agent-management.ts
@@ -213,6 +213,7 @@ export function editableAgentConfig(agent: AgentConfig): AgentConfig {
 	const base = agent.override?.base;
 	const {
 		override: _override,
+		outputMode: _outputMode,
 		model: _model,
 		fallbackModels: _fallbackModels,
 		thinking: _thinking,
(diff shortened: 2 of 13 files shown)


bbbRye007:
Thanks for the fast turnaround. Frontmatter plus agentOverrides with call-level precedence covers both cases I had in mind. I'll pick it up in the next release.

PR #1347, opened 2026-08-21
feat: expose model exclusions as a public package export

srcKod:
The TTL-backed model exclusion store (#1318) is exercised inside the dispatch pipeline, but extensions that run their own model-backed work have no way to reach it: nothing in the package `exports` map surfaces the store, so external dispatchers must fork a parallel exclusion system that drifts from the package's own fallback selection.

This adds a `pi-subagents/model-exclusions` subpath (mirroring the `pi-subagents/capability-ceiling` pattern from #1317) that re-exports the existing, already-tested store. A failure recorded by an extension lands in the same `<TEMP_ROOT_DIR>/model-exclusions.json` store the dispatcher filters against, so a 429 observed by an extension also protects the built-in fallback pool.

Purely additive: new `src/api/model-exclusions.ts` re-export file, one exports-map entry, a docs section, a CHANGELOG entry, a public-API unit test, and the `RecordModelFailureOptions` type gaining an `export`. No runtime behavior changes.

## Changes

- `src/api/model-exclusions.ts` (new): re-exports `recordModelFailure`, `isExcluded`, `filterFallbackCandidates`, `parseModelKey`, `getExcludedCount`, `clearExclusions`, `clearExpiredExclusions`, `setDefaultTTL`, `flushPersist`, `reloadFromDisk`, `EXCLUSIONS_PATH_ENV`, and the `ModelExclusion` / `RecordModelFailureOptions` types.
- `src/runs/shared/model-exclusions.ts`: `RecordModelFailureOptions` type gains an `export` so public callers can type their options object.
- `package.json`: `"./model-exclusions"` exports entry.
- `docs/extension-api.md`: "Model exclusions" section with usage example and semantics.
- `CHANGELOG.md`: entry under `[Unreleased] → Changed`.
- `test/unit/model-exclusions-public-api.test.ts` (new): 5 tests covering the public surface, shared store instance with the internal module, record/filter behavior, persistence, and provider/model key parsing.
- `test/unit/package-manifest.test.ts`: expected exports map updated with the new entry.

## Verification

- `npm run typecheck` clean.
- 33/33 tests pass across the new public-API suite, the existing `model-exclusions.test.ts` (19 tests), and `package-manifest.test.ts`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Changed
 - Make scripted-workflow helper support and stale-session recovery visible in `doctor` and the workflow guide (#1344).
 - Keep structured single-child execution receipts quiet by removing an internal conversion log from public workflow output.
+- Expose the TTL-backed model exclusion store as a public `pi-subagents/model-exclusions` package export so extensions can record model failures and skip excluded models in their own fallback selection.
 - Add per-agent model restrictions and a current-parent `inherit` allow-list alias. Thanks to [@hieudmg](https://github.com/hieudmg) for #1328.
 - Show package names, versions, and external-job provider status in subagent list and detail output so package agents such as Surf's `gpt-pro` are easier to find and use.
 - Document Council Mode routing in the main `pi-subagents` skill so natural-language requests for advisor councils, plan critique, cross-exam, or multiple model perspectives load the Council Mode protocol.
--- a/docs/extension-api.md
+++ b/docs/extension-api.md
@@ -238,6 +238,29 @@ Semantics:
 
 Schedules created while a ceiling is active are rejected until durable schedule persistence is available; unrestricted schedules remain subject to any policy active when they fire. Public status exposes bounded audit counts and sources, never full extension paths.
 
+## Model exclusions
+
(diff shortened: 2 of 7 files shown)


nicobailon:
Thanks for the follow-up. I’m going to close this one as not planned.

The model-exclusion store from #1318 already benefits pi-subagents internally. This PR mainly adds a public package entrypoint so other local extensions can mutate and query that store. That would make the TTL behavior, persistence path, global mutable state, clear/reload/flush controls, and key parsing part of the supported extension API.

I do not want to take on that public API maintenance burden without clear demand from a real integration. If this need comes back, the better next step is a design issue with the concrete consumer, why existing fallback/preflight/delegation APIs are not enough, and the smallest stable API needed. I would expect that shape to be much narrower than exporting the full store, likely closer to a single report-model-failure hook.

srcKod:
Thanks for the clear guidance — I agree the full store surface is too much to support without a concrete consumer. I'll follow the path you outlined: a design issue with the concrete consumer, why the current APIs fall short, and a minimal API proposal shaped like the single report-model-failure hook you mentioned. I'll reference this PR there for context.

PR #1349, opened 2026-08-21
feat(agents): support output and defaultReads overrides

mevatron:
## Summary
- allow `subagents.agentOverrides.<agent>` to set or clear `output` and `defaultReads`
- preserve custom-agent frontmatter precedence and prevent settings-derived values leaking into managed Markdown
- document the fields and add focused regression coverage

## Validation
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/agent-overrides.test.ts` (46 passing)
- focused agent-management preservation tests
- `npm run typecheck`
- local fork smoke test via uncommitted Nix config: scout output was configured as `scout-context.md`; worker received configured `defaultReads` (`scout-context.md`, `plan.md`) and reported the marker from the scout output without its task naming that file.

## Context
Revives the focused override portion of previously closed #231, reworked against current `main`. It intentionally excludes that PRs unrelated run-history work and does not add `defaultProgress` override support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Added
+- Let `agentOverrides` set or clear default `output` paths and `defaultReads`, while
+  preserving explicit custom-agent frontmatter and preventing settings-derived values
+  from being serialized into custom definitions. Thanks to [@mevatron](https://github.com/mevatron) for #1349.
+
 ## [0.54.0] - 2026-08-21
 
 ### Highlights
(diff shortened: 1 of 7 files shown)


nicobailon:
Thanks for this contribution, @mevatron. I landed the preserved behavior in replacement PR #1350 because this branch conflicted with current main and the replacement includes the follow-up layered-override fix from review. The changelog credit for #1349 is included in the merged change.

mevatron:
Thanks @nicobailon I fixed the conflict, but looks like you beat me to it! :smile:

PR #1359, opened 2026-08-22
fix: skip oversized async event tails across polls

fl4p:
## Summary
- detect when the async control-event cursor points inside a JSONL record
- continue skipping oversized event lines until their terminating newline
- prevent malformed-event warnings when large tool payloads are appended across polls

## Test
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/async-job-tracker.test.ts` (26 passed)

### DIFF
--- a/src/runs/background/async-job-tracker.ts
+++ b/src/runs/background/async-job-tracker.ts
@@ -213,11 +213,15 @@ export function createAsyncJobTracker(pi: Pick<ExtensionAPI, "events">, state: S
 					});
 				}
 			};
+			const previousByte = Buffer.allocUnsafe(1);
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the fix. Current `main` already has the same mid-line resynchronization behavior in `src/runs/background/async-job-tracker.ts`, and the integration coverage now includes `resynchronizes when a saved cursor resumes inside an oversized JSONL record`.

Because this PR is now merge-conflicted and no unique behavior remains to preserve, I am closing it as already fixed on current `main` rather than reworking the branch. I appreciate the report and patch.

PR #1405, opened 2026-08-23
feat: expose bounded current-work projection

### LINKED ISSUE #1404: Expose bounded current-session direct-work projection for SDK hosts
## Outcome
Expose one supported, versioned, read-only projection that SDK hosts can use to present direct subagent work without scraping transcripts, parsing human status text, or reading private runtime state.

## Context
Pi SDK hosts can retain a shared EventBus and call the existing `subagents:rpc:v1` bridge. Current installed/public projections split information across a limited active fleet and an async-only snapshot; foreground activity, terminal lifecycle, hierarchy semantics, and workflow ownership are not consistently available as safe structured data.

Consumer: Pi Interface issue Tom2906/pi-interface#30.

## Requirements
- Add a bounded `currentWork` v1 projection to the existing read-only `status` RPC response and advertise it in `ping` capabilities.
- Scope exclusively to the authoritative current native session.
- Include direct foreground and async work with opaque reconciliation keys, caller-facing goal, agent/role, mode, authoritative lifecycle state, safe activity/current tool, start/update/end timing, bounded token usage, and chain/parallel child structure.
- Include active, attention-requiring, and bounded recent terminal direct work so an open inspector can reconcile completion/failure/pause without timers.
- Exclude workflow runs and every workflow-owned foreground/async child at the projection source using explicit ownership fields; do not require consumers to infer ownership.
- Never expose run/async/tool IDs, paths, transcript/session files, prompts, reasoning, current-tool arguments, credentials, raw output, or callbacks.
- Sanitize and truncate all strings; cap roots, children, depth, and serialized bytes; report omissions.
- Preserve existing RPC v1 `fleet` and `asyncSnapshot` compatibility.
- Add public type/constants export for the projection contract so SDK consumers do not deep-import private source.

## Proof
- Current-session and cross-session attribution tests.
- Foreground and async single/chain/parallel lifecycle tests.
- Workflow and workflow-owned child exclusion tests.
- Attention/current-tool, redaction, omission, and byte-bound tests.
- RPC capability/response compatibility tests.
- Typecheck and full unit suite.

## Boundaries
Read-only projection only. No new execution controls, transcript access, scheduler, workflow state machine, or browser UI.

Tom2906:
## Summary
- add a public, bounded `currentWork` v1 status projection for SDK hosts
- cover current-session foreground and async lifecycle, activity, timing, usage, and hierarchy
- exclude workflows and workflow-owned children at the source, including restored foreground history
- export the contract and document the RPC capability

Closes #1404

## Validation
- `npm run typecheck`
- focused current-work/RPC/package/fleet suite: 72 passed
- `git diff --check`
- independent focused review: PASS

## Full-suite note
The full local unit command also exercised unrelated environment-sensitive tests and reported existing agent-discovery, symlink-permission, and acceptance-fixture failures. All tests covering this change and its direct integration surfaces pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 - Child launches fail fast with clear reasons when requested models or write tools are unavailable.
 
 ### Added
-- Add a bounded `currentWork` v1 RPC projection for SDK hosts, covering current-session direct foreground and async lifecycle, activity, timing, usage, and hierarchy while excluding workflow-owned work and private runtime data (#1404).
+- Add a bounded `currentWork` v1 RPC projection for SDK hosts, covering current-session direct foreground and async lifecycle, activity, timing, usage, and hierarchy while excluding workflow-owned work and private runtime data. Thanks to [@Tom2906](https://github.com/Tom2906) for #1405 (tracking issue #1404).
 - Add `subagents.maxThinking` to enforce a thinking ceiling across native subagent launches. Thanks to [@alex-real14](https://github.com/alex-real14) for #1397.
 - Add `subagents.defaultProvider` and per-agent `defaultProvider` overrides so bare subagent model ids can prefer a configured provider. Thanks to [@swingtempo](https://github.com/swingtempo) for #1393.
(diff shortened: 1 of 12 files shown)


Tom2906:
Withdrawn. This contribution was opened in error and will not be pursued.

PR #1429, opened 2026-08-23
feat: add session-scoped launch authority API

meowsigma:
## Summary
- add bounded, two-phase, one-use launch permits for trusted session policy extensions
- enforce exact static workflow manifests at public, delegated, RPC, and scheduled ingress before side effects
- add structured durable terminal-result replay for missed completion recovery
- document the API, capability advertisement, threat model, and schedule/nested behavior

## Validation
- `npm run typecheck`
- `HOME=<isolated> USERPROFILE=<isolated> npm run test:all`
  - unit suite passed
  - integration suite: 741 passed
  - E2E suite skipped because optional real Pi runtime fixtures were unavailable


nicobailon:
Thanks for the substantial PR. I reviewed the current head during backlog maintenance.

This cannot merge as-is. The current unresolved launch-authority finding is real: the PR commits authority against a preflight manifest, then the ordinary executor can still resolve mutable launch state again before the final spawn. That means the consumed permit can authorize a different runtime launch contract than the one the authority approved.

The branch is also conflicting against current `main`, and this is a large public/security API surface, so it needs a clean redesign/replacement before it can move forward. The useful idea is recorded, and any replacement that uses this work will credit #1429 / @meowsigma.

I am leaving this blocked rather than merging or reworking it in-place.

nicobailon:
Thanks for the launch-authority work here. I did a current-head backlog pass and a separate read-only architecture review.

Owner decision for now: keep this blocked and do not adopt a public session-scoped launch-authority API yet. The branch still has a real time-of-check/time-of-use gap because authority is decided against one preflight view, while the final launch recomputes agent, model, and tool state before spawn.

If we revisit this later, the safer shape is a clean replacement with a two-phase invariant: reserve an opaque token for the exact request digest at ingress, then verify and consume that token against the final launch contract, resolved agent identity, and model candidates immediately before spawn. Any mismatch fails closed, and the token is tombstoned after use.

Leaving this open as intentionally blocked rather than merging or reworking it in place.

nicobailon:
Closing this as not planned for the current API direction.

The backlog review kept this open while the rest of the queue landed, but the owner decision is now to not adopt a public session-scoped launch-authority API from this branch. The current design still has the material time-of-check/time-of-use gap called out above: authority is committed against a preflight view, while the final child launch can resolve mutable agent, model, and tool state again before spawn. That means a consumed permit can authorize a different launch contract than the one that was approved.

The branch also conflicts with current `main`, and the accepted replacement shape, if this is revisited later, should be a clean two-phase design: reserve an opaque permit for the exact request digest at ingress, then verify and consume it against the final resolved launch contract immediately before spawn. Mismatches must fail closed and permits must be one-use.

Thanks @meowsigma for the substantial work here. If a future replacement uses this design work, it should credit #1429 and @meowsigma.

PR #1439, opened 2026-08-24
feat(config): add configurable model exclusion TTL

### LINKED ISSUE #1438: Model-exclusion cache silently skips a recovered primary model for 24 hours
## Summary

A transient retryable failure of a primary subagent model creates a 24-hour entry in the TTL-backed model-exclusion cache. Until that entry expires, later launches can silently remove the recovered primary model from the candidate list and run the first fallback instead.

This makes a configured primary model appear unable to start even after it is healthy again.

## Environment

- `pi-subagents`: `0.56.0`
- Pi: `0.84.2`
- OS: macOS
- Primary reviewer model: a custom-provider `glm-5.3` model with `xhigh` thinking
- A fallback model was configured for the reviewer

## Reproduction

1. Configure a subagent with a primary model and a fallback model.
2. Let the primary model return one transient retryable error, such as `503 status code (no body)`.
3. Verify that the same primary model now succeeds in a direct Pi probe, including with the reviewer's normal read-only tool list.
4. Launch the subagent again.
5. Observe that it starts on the fallback without attempting the recovered primary model. In this case, the recorded `attemptedModels` contained only the fallback.
6. Inspect the model-exclusion cache and find an active entry for the primary model with an expiry roughly 24 hours after the transient failure.

## Actual behavior

`buildModelCandidates()` ultimately applies `filterFallbackCandidates()`. A live entry in `model-exclusions.json` removes the primary candidate before the run begins. With a fallback configured, the run can succeed on the fallback with no clear indication that the configured primary was skipped due to a cached failure.

The default duration is hard-coded in `src/runs/shared/model-exclusions.ts`:

```ts
let defaultTTLMs = 24 * 60 * 60_000;
```

There is an internal `setDefaultTTL(ms)` function, but no documented setting, CLI command, or environment variable for configuring the duration. `PI_MODEL_EXCLUSIONS_PATH` changes the cache location only.

## Expected behavior

A transient failure should not cause a recovered primary model to be silently bypassed for 24 hours with no operator control or visibility.

## Proposed improvements

1. Expose a supported setting such as `subagents.modelExclusionTtlMs` (or per-reason TTLs) for future exclusion entries.
2. Provide a supported command/API to list and clear a specific provider/model exclusion.
3. Surface a launch diagnostic when a candidate is skipped, including the cached reason and expiry time.
4. Consider a shorter TTL for transient `5xx` / network errors, or retrying the primary before falling back after a configurable cooldown.

## Related issue

Related to #1420, which describes an incorrect model-verification failure causing a 24-hour exclusion. This report is distinct: even a genuine transient provider `503` has no supported TTL configuration or scoped reset mechanism, and it can silently bypass a recovered configured primary model.

mithyer:
## Summary

- Add `modelExclusions.defaultTtlMs` to the extension config at `~/.pi/agent/extensions/subagent/config.json`.
- Apply the configured TTL when the extension starts or reloads, before child launches can record model failures.
- Validate the value, document its scope and existing-cache behavior, and add unit/integration coverage.

The setting controls only newly recorded exclusions. Existing `model-exclusions.json` entries retain their persisted `expiresAt` values.

## Configuration

```json
{
  "modelExclusions": {
    "defaultTtlMs": 300000
  }
}
```

## Testing

- `npm run typecheck`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/pi-coding-agent-dir.test.ts test/unit/model-exclusions.test.ts`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/index-child-registration.test.ts`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/orca-progress-tabs.test.ts`

The full unit/integration runs also contain existing failures in unrelated watchdog/fork-context tests; those same failures reproduce on the base commit. See issue #1438 for the motivating behavior.

Closes #1438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Added
-- Add `modelExclusions.defaultTtlMs` configuration for controlling the default lifetime of newly recorded model exclusions.
+- Add `modelExclusions.defaultTtlMs` configuration for controlling the default lifetime of newly recorded model exclusions. Thanks to @mithyer for #1439.
 - Add the built-in `codex-exec` profile with code-owned read-only argv, bounded JSONL terminal proof, and a final-message artifact.
 - External CLI runs now publish code-owned capability limits and compact workflow receipt metadata for adapter identity, artifacts, handoff mode, supervisor support, and resumability.
 - External one-shot runners now support bounded parser hooks and logs, environment allowlists, cached launch preflight, coalesced parser progress, and process-tree cleanup.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Added
+- Add `modelExclusions.defaultTtlMs` configuration for controlling the default lifetime of newly recorded model exclusions.
 - Add the built-in `codex-exec` profile with code-owned read-only argv, bounded JSONL terminal proof, and a final-message artifact.
 - External CLI runs now publish code-owned capability limits and compact workflow receipt metadata for adapter identity, artifacts, handoff mode, supervisor support, and resumability.
 - External one-shot runners now support bounded parser hooks and logs, environment allowlists, cached launch preflight, coalesced parser progress, and process-tree cleanup.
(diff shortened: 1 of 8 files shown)


nicobailon:
Thanks for the model-exclusion TTL work. This branch is now conflicting and the issue needed one more piece: launch-time visibility when a candidate is skipped by a cached exclusion.

I opened a clean replacement as #1450. It keeps the TTL config work, adds skipped-candidate warnings, and preserves credit for #1438/#1439.

I am closing this PR as superseded by #1450.

PR #1455, opened 2026-08-24
fix: resolve bare HTTPS git URLs from Pi settings for package agent discovery

trancikk:
## Summary

Pi stores git-installed packages in `~/.pi/agent/settings.json` as bare HTTPS URLs like `https://github.com/user/repo.git`. `pi-subagents` only resolved `git:` prefixed sources, so agent directories exposed by git-installed packages (e.g. `gsd-package`) were never discovered.

## Root cause

In `src/agents/agents.ts`, `resolveSettingsPackageRoot` only handled sources starting with `git:`, `npm:`, `file:`, absolute paths, or relative paths. Bare HTTPS/HTTP git URLs fell through to `return undefined`, so `collectSettingsPackageRoots` skipped them.

## Fix

Add a fallback that reuses the existing `parseGitPackagePath` helper for bare HTTPS/HTTP git URLs, mapping them to the same git cache directory (`~/.pi/agent/git/<host>/<repo>`) as `git:` prefixed sources. The fallback also handles SSH-style URLs such as `git@host:repo.git`.

This is an additive change that only activates for strings that previously returned `undefined`; existing `git:`, `npm:`, and path sources are unaffected.

## Verification

- `npm run typecheck` passes.
- Added regression test `test/unit/package-agent-git-url-discovery.test.ts` which asserts that an agent from a package installed via `https://github.com/...` is discovered.
- Manually verified that all 21 `gsd-*` agents from `https://github.com/trancikk/gsd-package.git` are now discovered.

## Test notes

Running the full unit suite on Windows shows several unrelated pre-existing failures (symlink permission errors and package-agent management tests that also fail on clean `main`). This change does not introduce new failures.

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -382,6 +382,13 @@ function resolveSettingsPackageRoot(source: string, baseDir: string): string | u
 	if (normalized === "." || normalized === ".." || normalized.startsWith("./") || normalized.startsWith("../")) {
 		return path.resolve(baseDir, normalized);
 	}
+	// Pi stores git-installed packages as bare HTTPS/HTTP URLs (e.g.
+	// "https://github.com/user/repo.git"). Treat those as git package roots so
+	// they resolve to the same git cache directory as "git:" prefixed sources.
+	const parsedGit = parseGitPackagePath(`git:${trimmed}`);
+	if (parsedGit) {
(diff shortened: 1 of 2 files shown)


trancikk:
duplicate of https://github.com/nicobailon/pi-subagents/pull/1454
closing

PR #1459, opened 2026-08-24
feat: show active task labels in status surfaces

phoenixdam:
## Summary

- show a bounded one-line task in compact foreground cards and prefer explicit workflow labels in multi-child and Fleet status
- share the Fleet shortcut constant so live-detail hints accurately advertise `Ctrl+Alt+F`
- propagate only bounded, sanitized workflow labels into Herdr summary/title metadata, restore the prior task after overlapping runs finish, and clear metadata at zero active runs
- keep raw task and goal prompts out of Herdr metadata

## Validation

- `npm run typecheck`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/render-helpers.test.ts test/unit/herdr-status-bridge.test.ts` (32 passing)
- `git diff --check`

The complete unit suite was also compared against an untouched `origin/main` worktree under the same Node 26 host. Existing environment-sensitive failures reproduce on the baseline; the focused changed-path tests above are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Added
 - Expose a bounded, versioned workflow-child summary across workflow results, status, receipts, and completion replay. Thanks to [@rochecompaan](https://github.com/rochecompaan) for #1453.
 - Show live context-window usage separately from cumulative token spend in status and Fleet views. Thanks to [@nazzeDe](https://github.com/nazzeDe) for #1444.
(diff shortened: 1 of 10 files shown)


phoenixdam:
Fixed the stale-label finding in `9a23c6f`. Authoritative refreshes now treat an omitted `taskLabel` as absence instead of restoring the prior step label, while still stripping unsanitized label input. Added regressions for both labeled-to-unlabeled transitions and labels that sanitize to empty.

Validation: `npm run typecheck` and 34 focused rendering/Herdr tests pass.

PR #1470, opened 2026-08-25
feat(api): add invocation-scoped launch authorization

maroffo:
## Summary

- add a public `pi-subagents/launch-authorization` provider API for extensions that reserve exact agent identities
- authorize fully resolved foreground attempts, dynamic workflow children, structured delegation, schedules, resumes, and async singles before child startup
- authorize every bounded detached startup retry and model fallback through exact one-use approvals
- carry root invocation plus workflow/child/attempt identity outside model-visible tool arguments
- propagate bounded reservation descriptors to async and nested children so a missing provider fails closed without a session-global allow window

## Motivation

`workflowScript` can select agent names dynamically, so an extension cannot safely authorize protected package agents by inspecting the top-level script text. Session-scoped allowlist widening also creates a concurrency window. This API lets an extension compare the actual runtime `launchContractDigest` and canonical agent source at the common child-launch boundary.

## Contract

- providers are exact-session, reload-safe, bounded, and synchronous
- every provider reserving an agent must allow; denial, throw, Promise, malformed response, missing inherited provider, missing detached approval, or replay blocks startup
- requests are frozen and prompt-free: identity, source, digests, model candidates, invocation/workflow/attempt metadata only
- authorization receipts remain provider-owned and are never serialized to children
- detached runners receive only exact one-use approvals for the owning process's pre-authorized bounded attempt matrix
- unreserved launches return before request normalization or provider callbacks
- this is a same-process policy boundary, not an OS sandbox against other current-user code

## Validation

Current head: `26f3dd6c4a378d78e4b50bfcc3acb7e2ecc9a4e9`.

Fresh checks on this head:

- `npm run typecheck` — pass
- focused empty-candidate and detached-attempt authorization integration — 4 pass, 0 fail
- full integration suite — 774 pass, 0 fail, 6 skip
- `npm pack --dry-run --json` — pass, 271 entries
- `git diff --check` — pass
- Greptile re-review — 5/5, no blocking code finding

The aggregate local `npm run test:all` attempt is not represented as green: it encountered an Orca cleanup timing failure that passed in isolation and a malformed-LSP 500 ms timing failure that reproduces unchanged from an archived `upstream/main` checkout. GitHub Actions on the exact PR head is therefore still required.

Real Pi-session E2E is unavailable in the isolated checkout because runtime packages are not installed. Bun is not installed locally, so the existing Bun workflowScript parity job remains CI-only.

## Performance

A local warm microbenchmark of the no-provider authorization path measured about 0.15 microseconds per call; no provider callback or request normalization occurs on that path.


maroffo:
Review update for `6216676e`:

The two original blockers are addressed:

- detached startup retries and model fallbacks now require exact, one-use attempt approvals before child spawn;
- stale/repeated provider disposers can no longer delete a recreated session registry.

Fresh local evidence on this commit: typecheck passed, focused launch-authorization tests passed 20/20, focused async integration tests passed 7/7, and the full rerun passed with 2503 unit tests and 774 integration tests (0 failures; 3 and 6 skips respectively). Real Pi-session E2E remains unavailable in the isolated checkout, and Bun is not installed locally.

The latest Greptile re-review confirms those fixes but raises a new P1 for native async launches with an empty model-candidate list. I verified that finding is valid: the owner authorizes a digest containing `modelCandidates: []`, while the runner currently recomputes `actualLaunchContractDigest` from its execution sentinel `[undefined]`. Exact approval consumption therefore fails closed for an otherwise authorized launch.

This PR is **not merge-ready yet**. The follow-up should add an empty-candidate regression and hash the persisted contract representation (`step.modelCandidates ?? []`) rather than the runner-only execution sentinel, then rerun the full gates.

maroffo:
Follow-up for the empty-candidate review at `26f3dd6c`:

The earlier comment classified this as a runtime P1 too quickly. Executable inspection showed that the existing `stableJson()` canonicalizer maps both `modelCandidates: []` and the runner-only `[undefined]` sentinel to the same launch digest, and the reserved async no-model path already completed successfully. So there was no demonstrated authorization failure on `6216676e`.

The code was nevertheless ambiguous and relied on that incidental canonicalization. This commit removes the ambiguity by hashing the persisted contract representation directly:

```ts
modelCandidates: step.modelCandidates ?? []
```

The integration regression now explicitly proves that a reserved async single with no selected model:

- presents four bounded startup approvals with `modelCandidates: []` and no selected model;
- completes successfully;
- preserves the exact owner/runner `launchContractDigest`;
- propagates only the reservation descriptor to the child.

Fresh post-commit evidence:

- `npm run typecheck` — pass
- focused empty-candidate/attempt authorization integration — 4/4 pass
- full integration suite — 774 pass, 0 fail, 6 skip
- `npm pack --dry-run --json` and `git diff --check` — pass

The local aggregate `test:all` attempt also encountered two unrelated timing/filesystem-sensitive unit failures: the malformed-LSP 500ms timeout reproduces unchanged from an archived `upstream/main`, and the Orca cleanup case passed in isolation. I am not representing that aggregate attempt as green. Real Pi-session E2E and Bun remain unavailable locally.

Please re-review the new head.

maroffo:
@nicobailon, the code-review blockers are resolved on `26f3dd6c`; Greptile now reports 5/5 with no blocking finding. The PR has not received any GitHub Actions run for the exact head, apparently because this is a fork contribution awaiting maintainer workflow approval.

Could you please approve/run the **Tests** workflow and review the PR? The PR body now records the current-head evidence and local limitations accurately. No merge is being requested or performed by me.

nicobailon:
I completed a read-only exact-head backlog review.

The implementation fixes the earlier final-spawn verification concern from #1429 better than the old session-scoped shape, and I did not find a source-proven implementation blocker in the consume path. It is still blocked before merge for two reasons:

1. Exact-head GitHub Actions are still `action_required`, so required CI has not run.
2. This adds a public `pi-subagents/launch-authorization` API. The standing decision from #1429 was not to own a public launch-authority API yet unless the owner accepts the exact public/security contract. This PR uses provider-owned receipts rather than a package-owned opaque reserve token.

So the disposition is keep-open/blocked. If the owner accepts this API shape, the next gates are CI approval, exact-head CI, fresh exact-head security/API review, credit check, and normal merge gates. If not, this should be reworked or closed as not planned while preserving credit for the design/code.

nicobailon:
Closing this as not planned.

The current branch is too large and too broad for this security/API surface: about 1.9k added lines across public exports, docs, foreground launch, async launch, detached runner startup, shared launch contracts, types, and multiple test layers. Even though the latest head addresses earlier implementation findings better than #1429, the resulting public `pi-subagents/launch-authorization` API is still not an owner-accepted contract.

For this kind of launch-authority boundary, the bar is intentionally adversarial. A future proposal needs to start from a much smaller design and prove one narrow invariant at a time, with the public API, threat model, lifecycle states, detached retry semantics, and CI gates reviewed before the implementation expands across launch paths.

If this direction is revisited later, the prior work here should still be credited to @maroffo/#1470.

Issue #1471, opened 2026-08-25
status returns "Async run not found" for LIVE async children (late-status window)


rafafortes:
**Body:**

Context: orchestrator harvests scheduled children via background workflow events. Between spawn and the first persisted state flush, `subagent({action:"status", id})` returns the literal error `Async run not found. Provide id or dir.` even though the child is live and already consuming tokens.

Measured on 2026-08-25 (engineer session): **6 occurrences** across trend/probe schedule ticks (`60d8cb73`, `d1c7e9dd`, `55778a23`, `8990d9c7`, `24d73228`, `597dc71c`). All six later completed successfully with valid structured output — one had 5 turns / 72k tokens at the moment status claimed it did not exist. Two were initially misclassified as INFRA failures by the harvester before the completion event arrived.

Impact: orchestrators cannot distinguish "run died" from "state not flushed yet"; naive retry-on-not-found double-spawns children; attention nudges (`steer`/`resume`) fired against a "missing" id are lost or misdelivered.

Additive proposals:

1. **Registered-pending state**: register the run id in the capacity/state store at spawn time so `status` can return `{state:"starting"}` instead of `not found`.
2. **Soft-error semantics**: keep the error but add a machine-readable code (e.g., `code:"possibly_starting"`) plus `spawnedAt` so callers can implement backoff instead of treating it as terminal.
3. **Docs**: document the late-status window explicitly in tool-reference (§status) with recommended harvest protocol (single inspection + wait for workflow completion event).

Local mitigation adopted meanwhile (watcher role memory): single `status` inspection, classify `late-status/pending`, never steer/resume/retry on this signal alone.

PR #1481, opened 2026-08-25
fix(runs): notify the parent when a runner dies before naming its session

### LINKED ISSUE #1480: A run whose runner dies before writing anything never notifies its parent
# A run whose runner dies before writing anything never notifies its parent

## Summary

When an async runner process exits before it records a `sessionId`, the stale-run
reconciler correctly marks the run failed — but the parent agent is never told.
It finds out only if it happens to ask later.

The parent had already handed control back on the strength of "N agents
started", so the failure surfaces as silence: the work simply never comes back.
Observed twice in one session, with 4 and 3 children respectively.

Reported message:

```
Async runner process 85090 exited or disappeared before writing a result.
Marked run as failed.
```

## Where it goes wrong

Completion notifications are delivered from result files —
`src/runs/background/notify.ts`:

> Async result files call this notifier directly and are deleted only after
> `sendMessage()` accepts the notification.

The reconciler writes that result file conditionally
(`src/runs/background/stale-run-reconciler.ts:278`):

```ts
function writeFailedRepair(asyncDir, status, resultPath, now, reason) {
  const repair = buildFailedRepair(status, asyncDir, now, reason);
  if (repair.result.sessionId) writeAsyncResultFile(resultPath, repair.result);
  ...
}
```

A runner that dies early has no `sessionId` recorded yet, so:

- the status is repaired to `failed` ✅
- no result file is written ❌
- no completion notification is delivered ❌

The guard therefore drops the notification in exactly the case that most needs
one: the earlier the runner dies, the less state it has recorded, and the more
completely the failure is swallowed. A crash during startup is silent by
construction.

## Suggested fix

Make the repaired failure deliverable without a `sessionId`. The run id is
always known (`status.runId || path.basename(asyncDir)`, already computed in
`buildFailedRepair`), so it can identify the result:

- write the result file whenever a run is repaired to `failed`, using the run id
  when `sessionId` is absent; or
- if the result file genuinely requires a session, notify through the same path
  the notifier uses, so the parent is woken either way.

Either way the invariant worth keeping is: **a run that stops running produces
exactly one completion signal, success or failure.** Today that holds only for
runs that lived long enough to name themselves.

## Impact

An agent that launches async children and returns control cannot rely on being
told when they die. It has to poll defensively, which defeats the purpose of the
completion notification, or it reports "started" as if it were "finished" —
which is what happened here.

VincentHanxiaoDu:
Fixes #1480.

## Problem

A completion notification is delivered from a run's result file, and result files are addressed by session (`writeIndexedPendingResultFile` throws without a `sessionId`). So `writeFailedRepair` only writes one when the status already carries a session:

```ts
if (repair.result.sessionId) writeAsyncResultFile(resultPath, repair.result);
```

A runner that exits during startup has not recorded a `sessionId` yet. Its run is correctly marked `failed` in `status.json`, but no result file is written and no notification is delivered — the parent, which had already handed back control on "N agents started", never hears anything again.

The guard drops exactly the failures that most need reporting: the earlier the crash, the less the runner recorded, and the more completely the failure is swallowed. A crash during startup is silent by construction.

Observed twice in one session (4 and 3 children), reported as:

```
Async runner process 85090 exited or disappeared before writing a result. Marked run as failed.
```

## Change

The session that started the run is already known independently — the tracker passes it in as `options.startedRun.sessionId`. It now stands in when the status has none, so the repaired failure is deliverable rather than merely recorded.

The invariant this restores: **a run that stops running produces exactly one completion signal, success or failure.** Today that holds only for runs that lived long enough to name themselves.

## Verification

One test added to `test/unit/stale-run-reconciler.test.ts`, covering a status with a pid and no session. It fails on `main` (the result file is absent) and passes with the change.

```
$ npm run typecheck    # clean
$ node --test test/unit/stale-run-reconciler.test.ts
  before: 10 pass, 1 fail
  after:  11 pass, 0 fail
$ npm run test:unit
  2476 pass, 17 fail
```

Those 17 failures are pre-existing and unrelated (agent enable/disable management); `main` at the same commit reports 2475 pass / 17 fail.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -664,7 +664,6 @@
       "integrity": "sha512-3giAOQvZiH5F9bMlMiv8+GSPMeqg0dbaeo58/0SlA9sxSqZhnUtxzX9/2FzyhS9sWQf5S0GJE0AKBrFqjpeYcg==",
       "dev": true,
       "license": "Apache-2.0",
-      "peer": true,
       "engines": {
         "node": ">=8.0.0"
       }
@@ -1463,7 +1462,6 @@
       "integrity": "sha512-ytENFjIJFl2UwYglde2jchW2Hwm4GJFLDiSXWdTrJQBIN9Fcyp7n4DhxJEiWNAJMV1/BqWfW/kkg71UDcHJyTQ==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "funding": {
         "url": "https://github.com/sponsors/colinhacks"
       }
--- a/src/runs/background/stale-run-reconciler.ts
+++ b/src/runs/background/stale-run-reconciler.ts
@@ -351,8 +351,17 @@ export function reconcileAsyncRun(asyncDir: string, options: ReconcileAsyncRunOp
 	const now = options.now?.() ?? Date.now();
 	const status = readStatus(asyncDir);
 	const startedStatus = !status && options.startedRun ? buildStartedStatus(asyncDir, options.startedRun, now) : undefined;
-	const effectiveStatus = status ?? startedStatus;
-	if (!effectiveStatus) return { status: null, repaired: false };
+	const started = status ?? startedStatus;
+	if (!started) return { status: null, repaired: false };
+	// A result file is addressed by session, so a repair without a sessionId
+	// cannot be written - and the completion notification is delivered from that
+	// file. A runner that died before its child session existed has no sessionId
+	// in status.json yet, which silently swallowed exactly the failures that
+	// most need reporting: the earlier the crash, the less it recorded. The
+	// session that started the run is known independently, so it stands in.
+	const effectiveStatus: AsyncStatus = started.sessionId || !options.startedRun?.sessionId
+		? started
+		: { ...started, sessionId: options.startedRun.sessionId };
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks for the fix. I am keeping this open until #1482 lands because it covers the same #1480 failure at the shared launch/status seam and also fixes #1471.

I am not planning to merge this branch as-is. The current diff copies `sessionId` into the stale repair result, but the result watcher also requires the matching `completionOwnerId`, so the repaired result can still be rejected before parent delivery. #1482 persists both values in the launcher-owned initial status and credits @VincentHanxiaoDu for #1480 in the changelog.

Once #1482 lands, I will close this as superseded. If #1482 does not land, this branch would need owner-scoped delivery coverage, the missing `completionOwnerId`, removal of the incidental lockfile metadata churn, a rebase on current `main`, and the full repo CI surface.

nicobailon:
Closing as superseded by #1482, now merged as 43c71dca7e360410981afd1d0ff3929427eb7597.

The merged fix covers the same early-runner-death notification path with launcher-owned sessionId and completionOwnerId, and also fixes the related #1471 live-status window. @VincentHanxiaoDu is credited for #1480 in the changelog.

Issue #897, opened 2026-08-08
pi exiting due to uncaughtException:


pascalandy:
running ` DeepSeek: DeepSeek V4 Flash 0731  think:xhigh`
---

pi exiting due to uncaughtException:
Error: This extension ctx is stale after session replacement or reload. Do not use a captured pi or command ctx after ctx.newSession(), ctx.fork(), ctx.switchSession(), or ctx.reload(). For newSession, fork, and switchSession, move post-replacement work into withSession and use the ctx passed to withSession. For reload, do not use the old ctx after await ctx.reload().
    at ExtensionRunner.assertActive (file:///Users/andy16/Library/pnpm/store/v11/links/@earendil-works/pi-coding-agent/0.84.1/6e3a6bfab8e6fd1679d16fb86ebe0add38fc8e04cb5758a5c3fb4bcb6ab56eb9/node_modules/@earendil-works/pi-coding-agent/dist/core/extensions/runner.js:360:19)
    at get cwd (file:///Users/andy16/Library/pnpm/store/v11/links/@earendil-works/pi-coding-agent/0.84.1/6e3a6bfab8e6fd1679d16fb86ebe0add38fc8e04cb5758a5c3fb4bcb6ab56eb9/node_modules/@earendil-works/pi-coding-agent/dist/core/extensions/runner.js:477:24)
    at getQueueContext (/Users/andy16/.pi/agent/npm/node_modules/pi-powerline-footer/index.ts:1241:48)
    at Timeout._onTimeout (/Users/andy16/.pi/agent/npm/node_modules/pi-powerline-footer/index.ts:1363:51)
    at listOnTimeout (node:internal/timers:605:17)
    at process.processTimers (node:internal/timers:541:7)

nicobailon:
Thanks, this is a real Powerline Footer issue rather than a pi-subagents runtime issue.

I traced the stack to the delayed post-compaction queue delivery path in `pi-powerline-footer`. That timer could still hold an old extension context and read `ctx.cwd` after reload/session replacement.

I opened the fix here: https://github.com/nicobailon/pi-powerline-footer/pull/148

The PR snapshots the queue context before the delayed callback, skips stale session generations, and keeps queued items queued if a stale context race happens before delivery.

nicobailon:
Fixed in `pi-powerline-footer` by https://github.com/nicobailon/pi-powerline-footer/pull/148.

[...]

Issue #1472, opened 2026-08-25
Terminal async runs hold capacity slots indefinitely when proof is unknown or runner PID is gone


rafafortes:
**Body:**

Context: multi-session orchestration host running up to 6 interactive pi sessions, each dispatching waves of async children under `maxActiveAsyncRunsPerSession` (we run 20; measured 17–18 healthy concurrent).

While diagnosing throughput decay we found that the reconcile path in `src/runs/background/active-async-capacity.ts` only releases a slot in the narrow "failed before child startup completed" branch (~lines 181–185). When a run is already terminal but its proof lands as `unknown` with reason in {stale-repair, runner-candidate-missing, observer-unavailable, proof-write-failed}, the slot is retained indefinitely. Real incident from today: a scheduled child died with

```
"error": "Async runner process 3177497 exited or disappeared before writing a result. Marked run failed by stale-run reconciliation."
```

and the owning session's slot stayed held until the whole session was restarted (capacity pools of dead sessions also accumulate under `/tmp/pi-subagents-<uid>/session-active-async-capacity/<old-session>/`).

**Proposal (additive):**
1. In reconcile, release the slot when the run is terminal AND the proof is terminal-irrecoverable (`unknown` + reasons above), or when the runner PID no longer exists (`kill -0`).
2. Startup reaper for capacity pools whose owning session no longer exists.
3. Age-based eviction: `reservedAt` older than N hours (e.g., 24h) without any observed proof.

This turns "silent decay → session restart" into self-healing capacity.

---

nicobailon:
I tried to reduce this to a safe local fix while working the lifecycle cluster, but the obvious release predicate is not strong enough.

A dead runner PID proves only that the runner process is gone. It does not prove the Pi writer child or its descendant process tree is gone. The existing process-terminal contract waits for writer close records and observed process-tree cleanup before treating a run as strongly terminal.

So #1472 stays open and blocked for now. A future fix needs a durable failure-boundary artifact that proves all expected writer process trees are gone. I am not going to release capacity from `stale-repair`, `observer-unavailable`, `proof-write-failed`, or raw PID liveness alone because that can undercount live orphan work.

nicobailon:
I ran a deeper architecture pass on this after the initial blocker comment.

[...]

PR #1485, opened 2026-08-25
fix: survive stale session ctx on async process-terminal delivery

AdrianAcala:
## Problem

pi can exit with `uncaughtException` when an async runner exits after its owning session has been replaced or reloaded:

```
pi exiting due to uncaughtException:
Error: This extension ctx is stale after session replacement or reload. ...
    at Object.emit (.../chunk-*.js)
    at .../pi-subagents/src/runs/background/async-execution.ts:1776:32
    at ChildProcess.<anonymous> (.../async-execution.ts:603:26)
```

## Root cause

`spawnRunner()` registers a `proc.once("close", ...)` handler on the detached async runner. At the end of that handler it calls `onProcessTerminal?.(persisted)` (line 603), and both launch sites pass a callback that captures the extension command `ctx` from launch time:

```ts
(proof) => ctx.pi.events.emit(SUBAGENT_PROCESS_TERMINAL_EVENT, proof),  // lines 1243, 1776
```

If the session is replaced (`ctx.newSession()`, `ctx.fork()`, `ctx.switchSession()`) or reloaded while the runner is still running, pi-core invalidates the old ctx. When the runner later exits, the late `emit()` hits `assertActive()`, which throws — from inside a child-process close callback, outside any tool-handler call stack — so the error becomes an uncaughtException and takes the session down.

This is the same stale-ctx class of bug as the animation-timer crash fixed in v0.21 (#122), but the async process-terminal path was never guarded; the only existing guard (`subagent-executor.ts`) covers foreground runs.

## Fix

In `src/runs/background/async-execution.ts`:

1. Add `isStaleExtensionContextError()` using the same regex already used in `subagent-executor.ts`.
2. Add `emitProcessTerminalEvent(ctx, proof)` (exported for testing): when the ctx is stale, drop the event silently — the process-terminal proof is **already persisted to `process-terminal.json`** by the close handler before the emit, and the replacement session's extension instance reconciles run state from disk. Non-stale emit failures are logged instead of thrown.
3. Both launch sites (single + chain/workflow) use the helper instead of the raw `ctx.pi.events.emit`.
4. Wrap `onProcessTerminal?.(persisted)` in the close handler in try/catch so the detached child-process close path can never take down the host regardless of what the callback does.

No hot-path cost: this only touches the rare late-exit path after session replacement.

## Verification

- `npm run typecheck` — clean
- `npm test` — 2494 pass, 0 fail (includes 3 new unit tests for `emitProcessTerminalEvent`: live-ctx emit, stale-ctx drop, non-stale failure logged)
- End-to-end repro against the real `executeAsyncSingle` path: launch an async run, invalidate the ctx (simulated session replacement), SIGKILL the runner process group.
  - Unpatched v0.56.0: host dies with `uncaughtException: This extension ctx is stale after session replacement or reload.`
  - Patched: `process-terminal.json` proof written, stale event dropped, host process survives.
- Note: one pre-existing integration failure (`fork-context-execution.test.ts` — `uses request cwd for execution-time agent discovery`) also fails on unmodified upstream HEAD; unrelated to this change.

Closes the crash reported in #897's stack (that issue itself was attributed to pi-powerline-footer; this stack is the pi-subagents async path).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Repair bounded dead async run candidates before retention classifies them, so existing terminal retention guards can reclaim stale run directories without deleting ambiguous worktrees or branches. Thanks to [@rafafortes](https://github.com/rafafortes) for #1477.
 - Make async runs visible to exact status lookup as soon as launch succeeds, and deliver one completion when a runner dies before its normal status write. Thanks to [@rafafortes](https://github.com/rafafortes) for #1471 and [@VincentHanxiaoDu](https://github.com/VincentHanxiaoDu) for #1480.
 - Record explicit completed, failed, timed-out, stopped, and interrupted outcomes in run history. Thanks to [@rafafortes](https://github.com/rafafortes) for #1474.
+- Keep async runner process-terminal event delivery from crashing the session when the captured extension context is stale after a session replacement or reload; the terminal proof is already persisted to the run directory before the event, and the replacement session reconciles run state from disk. Thanks to [@AdrianAcala](https://github.com/AdrianAcala) for #1485.
 - Preserve the local user identity and temporary-directory environment needed by authenticated Claude Code adapter runs.
 - Preserve structured-output and related bounded child contract fields when foreground workflow children resume. Thanks to [@Livan-pro](https://github.com/Livan-pro) for #1460.
 - Preview runtime-recorded workflow child sessions in Fleet and inspect without trusting sibling transcripts. Thanks to [@JHa13y](https://github.com/JHa13y) for #1441.
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -127,7 +127,7 @@ function resolveJitiCliPath(): string | undefined {
 
 const jitiCliPath = resolveJitiCliPath();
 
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks for the stale-session-context fix. This is a distinct process-terminal delivery issue, so I am keeping it open rather than closing it as superseded by #1482, #1484, or #1486.

It cannot merge as-is yet because the branch is currently conflicting, it does not include changelog credit for @AdrianAcala/#1485, and it has not run the full exact-head CI surface. The change also overlaps `async-execution.ts`, so the clean path is to rebase or fold it after the active #1482 async-start fix settles.

I checked that this does not try to solve #1472 by releasing capacity from raw runner PID death, which is good. The next accepted shape should keep that boundary: contain stale extension-context event-bus failures, add the contributor credit, and preserve the existing capacity fail-closed behavior.

nicobailon:
Closing as superseded by #1488, now merged as 4868199a226aaaf967f80afb27ad203c8cd1d03a.

The merged maintainer branch preserves the useful stale session-context process-terminal event handling from this PR, rebased on top of the merged async-start work, and credits @AdrianAcala/#1485 in the changelog.

PR #1508, opened 2026-08-25
fix: retry provider connection errors

genkikadomatsu:
## Why

A provider can report the transient failure string `Connection error`. The model-fallback classifier recognizes adjacent network failures but not this wording, so configured fallback models are skipped.

## What changed

- classify `connection error`, `connection reset`, `connection closed`, and `connection aborted` as retryable provider failures
- preserve the existing guard that prevents network-flavored tool failures from retrying an entire task
- add an async dispatch regression proving `Connection error` advances to the fallback model
- add a public foreground-result regression proving a non-retryable provider reason is returned when no model output exists

## What didn't change

- tool-call failures remain non-retryable
- context-overflow and startup-retry behavior are unchanged
- failure-result propagation already existed on current `main`; this PR adds explicit coverage rather than duplicating that implementation

## Verification

- `npm test` with inherited Pi runtime state removed: unit suite passed
- focused model-fallback tests: 80 passed
- focused foreground/background integration tests: 2 passed
- `npm run typecheck`
- independent Terra review: PASS

Downstream report: HEDSWE-3060

### DIFF
--- a/src/runs/shared/model-fallback.ts
+++ b/src/runs/shared/model-fallback.ts
@@ -425,6 +425,7 @@ const RETRYABLE_MODEL_FAILURE_PATTERNS = [
 	/overloaded/i,
 	/service unavailable/i,
 	/temporar(?:ily)? unavailable/i,
+	/connection\s+(?:error|reset|closed|aborted)/i,
 	/connection refused/i,
 	/fetch failed/i,
 	/network error/i,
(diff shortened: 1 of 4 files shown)


nicobailon:
Closing this as superseded by #1514, which landed the provider connection fallback fix after rebasing onto current `main`, addressing the current Greptile findings, and passing the required exact-head gates.

Thanks [@genkikadomatsu](https://github.com/genkikadomatsu) for the original contribution in #1508.

PR #1510, opened 2026-08-25
fix: descend into symlinked directories in agent discovery (#1505)

### LINKED ISSUE #1505: Agent discovery skips symlinked directories under .agents/ (unlike skill discovery, cf. #262)
## Summary

`listFilesRecursive` (in `src/agents/agents.ts`) does not descend into **symlinked directories** when it discovers agent definition files. A `.md` agent file inside a symlinked directory is never found, so the agent silently does not exist: `subagent({ action: "list" })` omits it and `doctor` reports `project 0`.

Skill discovery already handles this — it checks `child.isDirectory() || child.isSymbolicLink()` — which is why skills symlinked the same way resolve but agents do not. It is the same scanner-disagreement bug fixed for skills in #262.

## Current behavior

`listFilesRecursive` only recurses when `entry.isDirectory()` is true:

```ts
for (const entry of entries) {
  const filePath = path.join(dir, entry.name);
  if (entry.isDirectory()) {
    if (!shouldPruneDiscoveryDir(rootDir, filePath, entry.name)) {
      files.push(...listFilesRecursive(filePath, predicate, rootDir));
    }
    continue;
  }
  if (!entry.isFile() && !entry.isSymbolicLink()) continue;
  if (!predicate(entry.name)) continue;   // predicate = name.endsWith(".md")
  files.push(filePath);
}
```

For a `fs.Dirent` that is a symlink to a directory, Node reports:

```
isDirectory: false   isFile: false   isSymbolicLink: true
```

So the entry falls through to the file branch, fails the `.endsWith(".md")` check (its name is e.g. `agents`), and is dropped. The walker never enters the symlink target, so no agent file inside is seen.

## Why this is a problem

Symlinking a shared `.agents/agents` (and `.agents/skills`) into a source-of-truth repo is a common way to reuse agents and skills across checkouts:

```
myproject/.agents/agents -> /path/to/toolkit/.agents/agents   # symlink to a dir
myproject/.agents/skills -> /path/to/toolkit/.agents/skills   # symlink to a dir
```

The skill scanner resolves the skills, but the agent scanner finds **zero** agents — though the frontmatter is valid and the files parse. Nothing warns you: there is no diagnostic, and the agent never shows up, not even as an invalid entry.

The two scanners should agree. A directory layout that resolves for skills should resolve for agents.

## Repro

1. Create an agent at `/tmp/toolkit/.agents/agents/my-agent.md` with valid `name` + `description` frontmatter.
2. In a project dir, symlink the directory: `ln -s /tmp/toolkit/.agents/agents .agents/agents`.
3. Run `subagent({ action: "list" })` (or `doctor`) — `my-agent` is absent and the project agent count is `0`.
4. Replace the directory symlink with a real directory holding a symlink to the `.md` **file** (`ln -s /tmp/toolkit/.agents/agents/my-agent.md .agents/agents/my-agent.md`) — the agent now resolves. Symlinked files load; symlinked directories are skipped.

## Proposed fix

Match the skill-discovery fix from #262: treat a symlinked directory as recursable. `fs.Dirent.isDirectory()` does not follow the link, so resolve the target with `fs.statSync` before deciding:

```ts
for (const entry of entries) {
  const filePath = path.join(dir, entry.name);
  let isDir = entry.isDirectory();
  if (entry.isSymbolicLink()) {
    try { isDir = fs.statSync(filePath).isDirectory(); } catch { /* dangling symlink */ }
  }
  if (isDir) {
    if (!shouldPruneDiscoveryDir(rootDir, filePath, entry.name)) {
      files.push(...listFilesRecursive(filePath, predicate, rootDir));
    }
    continue;
  }
  if (!entry.isFile() && !entry.isSymbolicLink()) continue;
  if (!predicate(entry.name)) continue;
  files.push(filePath);
}
```

Symlinked `.md` files still load. Guard against symlink cycles if arbitrary depth is a concern (e.g. track resolved real paths).

## Environment

- `pi-subagents` v0.56.0 (also present on `main` — `listFilesRecursive` is unchanged at current HEAD)
- macOS

## Related

- #262 — the same symlink/nested-directory disagreement, fixed for **skill** discovery. The agent walker never got the equivalent treatment.

robsdudeson:
## Summary

Fixes #1505. Agent discovery skipped symlinked directories under `.agents/`, so a symlinked `.agents/agents` directory (a common way to share agents across checkouts) registered zero agents while skills symlinked the same way resolved fine.

`listFilesRecursive` relied on `fs.Dirent.isDirectory()`, which does not follow symlinks — a symlinked directory reports `isDirectory: false`, falls through to the file branch, fails the `.md` predicate, and is dropped. The walker never enters the target.

## Fix

Resolve symlink targets with `fs.statSync` before deciding whether to recurse, guarding dangling symlinks. This mirrors the skill-discovery scanner (#262) so the two scanners agree: a layout that resolves for skills now resolves for agents.

## Tests

New `test/unit/agent-symlink-discovery.test.ts`:
- discovers agents inside a symlinked `.agents/agents` directory
- discovers agents inside a nested symlinked subdirectory
- ignores a dangling directory symlink without throwing

Confirmed 2/3 fail without the fix and all pass with it.

## Verification

- `tsc --noEmit`: clean
- Full unit suite: 2507 pass / 0 fail / 3 skipped

### DIFF
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -1603,9 +1603,20 @@ function shouldPruneDiscoveryDir(rootDir: string, dir: string, dirName: string):
 	return path.resolve(dir) !== path.resolve(rootDir) && isDiscoveryNestedProjectRoot(dir);
 }
 
-function listFilesRecursive(dir: string, predicate: (fileName: string) => boolean, rootDir = dir): string[] {
+function listFilesRecursive(dir: string, predicate: (fileName: string) => boolean, rootDir = dir, visited = new Set<string>()): string[] {
 	const files: string[] = [];
 	if (!fs.existsSync(dir)) return files;
+
+	// Track canonical paths so symlinks that point back to an ancestor cannot
+	// send recursion into an infinite loop.
+	let canonicalDir: string;
+	try {
+		canonicalDir = fs.realpathSync(dir);
+	} catch {
+		canonicalDir = path.resolve(dir);
+	}
+	if (visited.has(canonicalDir)) return files;
+	visited.add(canonicalDir);
 
 	let entries: fs.Dirent[];
 	try {
@@ -1628,7 +1639,7 @@ function listFilesRecursive(dir: string, predicate: (fileName: string) => boolea
 		}
 		if (isDir) {
 			if (!shouldPruneDiscoveryDir(rootDir, filePath, entry.name)) {
-				files.push(...listFilesRecursive(filePath, predicate, rootDir));
+				files.push(...listFilesRecursive(filePath, predicate, rootDir, visited));
 			}
 			continue;
 		}
--- a/src/agents/agents.ts
+++ b/src/agents/agents.ts
@@ -1616,7 +1616,17 @@ function listFilesRecursive(dir: string, predicate: (fileName: string) => boolea
 
 	for (const entry of entries) {
 		const filePath = path.join(dir, entry.name);
-		if (entry.isDirectory()) {
+		// isDirectory() does not follow symlinks; resolve the target so a symlinked
+		// directory recurses like a real one (matches skill discovery, cf. #262).
+		let isDir = entry.isDirectory();
+		if (entry.isSymbolicLink()) {
+			try {
+				isDir = fs.statSync(filePath).isDirectory();
+			} catch {
+				// dangling symlink: leave isDir false and fall through to the file branch
+			}
+		}
+		if (isDir) {
 			if (!shouldPruneDiscoveryDir(rootDir, filePath, entry.name)) {
 				files.push(...listFilesRecursive(filePath, predicate, rootDir));
 			}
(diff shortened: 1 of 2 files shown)


robsdudeson:
Addressed both Greptile findings in 786f36d:

1. **Symlink cycles** — `listFilesRecursive` now tracks canonical `realpathSync` paths in a visited set and skips any directory it has already entered, so a symlink pointing back to an ancestor terminates instead of recursing. Added a regression test for the ancestor-cycle case.
2. **Windows portability** — tests now create symlinks with the repo's platform pattern (`junction` on win32) and skip gracefully on Windows `EPERM` (no Developer Mode/elevation) rather than failing all cases.

Full unit suite: 2508 pass / 0 fail.

nicobailon:
Thanks for the update here. I landed #1506 for the same #1505 fix so we could keep the already-reviewed cycle protection, pruning, malformed-agent diagnostics coverage, and changelog credit together.

The changelog credits @robsdudeson for both #1505 and #1510. Closing this PR as superseded by #1506.

PR #1518, opened 2026-08-25
feat: add one-use workflow child permit

### LINKED ISSUE #1494: RFC: package-owned one-use permit for one exact workflow child spawn
## Decision requested

Would maintainers accept the invariant and first slice below as a design direction worth specifying before code?

This issue does **not** request implementation, reopening #1429 or #1470, or approval of a permanent public API. If this is not a current priority, closing the RFC as not planned is a complete answer.

## Downstream use case

Pi Forge has a policy extension that protects exact package-qualified agents. For an ordinary direct `subagent` call, the extension can inspect the structured request and use `pi-subagents/preflight` to reject shadowed agents or modified model/tool/extension contracts.

With `workflowScript`, the top-level request contains JavaScript rather than the child launch resolved later by `runs.run`. A script can construct the agent name dynamically, so script-text inspection cannot authorize the final child. Regex scanning, a session-global allow window, and model-visible `extensionBindings` are not safe substitutes. Generic workflows for unreserved agents must remain available.

The eventual downstream workflow uses several protected critics and a synthesizer. This RFC intentionally begins with one foreground child and does not claim to unblock that complete async/parallel workflow.

## One invariant

> Within one foreground workflow root, an identity declared reserved before execution cannot start unless `pi-subagents` synchronously consumes a package-owned, one-use permit whose stored request and expected contract exactly match the immutable final launch plan immediately before the single process-creation call.

For this slice, the comparison binds:

- owning session and workflow root invocation;
- one workflow child key/index;
- canonical package-qualified agent name, package source, and definition digest;
- caller task digest;
- final `launchContractDigest`;
- the single model candidate;
- native runner and context mode.

Absence, mismatch, prior consumption, or use from another root/child slot denies startup. Identities not declared reserved in that root retain current behavior.

## Package-owned permit and host-only carrier

For this first same-process slice:

- `pi-subagents` resolves the exact child request and creates an opaque, non-serializable in-memory permit object;
- the package retains the permit record and reservation membership; downstream code receives no mutable contract fields;
- a host-only execution entry point binds the permit to one workflow root and child key outside the `workflowScript` source and model-visible tool parameters;
- `runs.run` retrieves the reservation only from that package-owned root context;
- neither the script nor the child receives the permit through task text, extension bindings, environment, status, artifacts, or result payloads;
- final verification invokes no downstream callback and opens no session-global authorization window.

API-neutral flow:

1. Pi Forge asks `pi-subagents` to reserve one exact native child request.
2. `pi-subagents` returns an opaque same-process permit.
3. Pi Forge starts one foreground workflow through a host-only entry point and supplies the permit outside workflow-visible input.
4. The workflow calls `runs.run` with a child key and dynamically resolved request.
5. `pi-subagents` compares the final immutable launch plan and consumes the matching permit immediately before spawn.

If no existing execution entry point can carry this context, the RFC asks whether adding that one host-only carrier is an acceptable minimal API surface. Names and signatures remain undecided.

## Reservation and consumption lifecycle

The first slice has one authorizing transition:

```text
reserved --exact synchronous verify--> consumed --single spawn call--> terminal
```

`consumed` is permanently non-authorizing. Verification and `reserved -> consumed` occur synchronously in the package-owned execution frame. Between that transition and the one process-creation call there must be:

- no `await` or yield to workflow/model/downstream code;
- no callback;
- no re-resolution or launch-affecting filesystem/configuration read;
- no second process-creation opportunity.

Every exit keeps the permit non-authorizing. A pre-spawn or spawn failure does not restore it; the host must reserve a new permit explicitly.

Reserved identity membership is separate from permit state for the lifetime of the foreground root. A missing or consumed permit therefore cannot make that identity fall through to unreserved behavior. When the foreground root exits, it can no longer launch, and all unused in-memory records are discarded.

## Small first implementation slice

If the design is accepted, the first implementation would prove only:

- one trusted same-process host extension;
- one foreground `workflowScript` root;
- one `runs.run` native Pi child;
- one exact package-qualified agent and child key;
- `async: false`;
- one model candidate;
- no startup retry or fallback.

Within that root, attempts to launch the reserved identity through any unsupported workflow operation must deny before spawn rather than fall through as unreserved.

Out of scope for this permit and requiring separate design approval:

- `runs.all` or multiple permits;
- top-level async or detached execution;
- startup retries and model fallback;
- schedules, resume, nested delegation, or nested authority;
- external CLI/job runners;
- persistence or recovery across reload;
- durable or cross-process permit representations.

The first permit cannot be serialized or widened to cover those paths. No behavior for those paths is proposed here beyond denial when they try to use this root’s reserved identity.

## Threat model

This is a same-process extension policy boundary, not an OS sandbox against arbitrary same-user code.

It is intended to prevent:

- model-authored or dynamically constructed workflow input from selecting a reserved agent without the exact host reservation;
- task, child-slot, agent identity, definition, model, capability, or context substitution;
- time-of-check/time-of-use drift caused by resolving launch-affecting state after permit consumption;
- replay, sibling substitution, or duplicate spawn from one permit;
- cleanup or permit loss from silently changing a reserved identity into an unreserved one while the root can still launch.

It does not claim to prevent a malicious same-user process from inspecting memory, replacing installed files, debugging the process, or altering the OS.

## Questions to settle before implementation

1. Is an opaque, non-serializable same-process object the appropriate first permit representation?
2. Which host-only execution entry point should bind that object to one foreground workflow root and child key without exposing it to the script?
3. Should the package define a new versioned final-permit projection, or is `launchContractDigest` plus explicit root/slot/identity metadata sufficient?
4. Which internal execution frame can construct an immutable launch plan and perform synchronous consume-plus-spawn without another launch-affecting read?

No API name or signature is proposed as accepted until these are resolved.

## Evidence and CI gates for the first slice

Before any broader lifecycle is proposed:

- document the exact permit projection and the `reserved -> consumed` ownership/release rule;
- unit-test root/child isolation, mismatch, prior consumption, duplicate disposal, and permanent consumption after spawn failure;
- integration-test a dynamically constructed reserved name denied without a permit and allowed exactly once with a matching permit;
- prove task, definition, model, context, or child-key drift denies before process creation;
- prove an unsupported workflow operation cannot launch the reserved identity or downgrade it to unreserved;
- use a spawn canary and barrier-based double-consume test to prove at most one process-creation call;
- run typecheck, unit, integration, package-export, and exact-head GitHub Actions checks;
- obtain fresh exact-head API and security review before proposing a second lifecycle.

## Non-goals

- parsing or restricting generic `workflowScript` source;
- replacing `preflight` or `extensionBindings`;
- adding a provider callback registry or provider-owned receipt;
- authorizing arbitrary same-user code;
- solving Pi Forge’s async panel in this slice;
- covering every launch path in one PR;
- specifying detached, retry, parallel, resume, schedule, or reload semantics here.

## Prior work and credit

- #634 established the public resolved launch-contract preflight used by the downstream policy.
- #1410 established bounded extension bindings, which carry child metadata but are intentionally not launch authority here.
- #1429 by @meowsigma explored two-phase launch permits and identified the exact-final-contract requirement. Its owner feedback is the source of the package-owned opaque-permit direction.
- #1470 by @maroffo proved final-spawn verification across more launch paths, but its provider-owned public API and approximately 1.9k-line scope were not accepted. This RFC starts over with one invariant, one same-process carrier, and one foreground child.

nicobailon:
Implements the accepted internal-only first slice for #1494.

This adds a package-private one-use permit for one exact native child spawn from a foreground `workflowScript` root. The permit is host-only and in-memory. It is not serialized into workflow state, status, receipts, artifacts, transcripts, environment, or model-visible params.

The final permit comparison still happens at the resolved launch-plan point before native spawn. The permit now also claims or consumes on the first distinct `runs.run` attempt, so a wrong child attempt cannot be followed by the expected child in the same root.

Validation:
- focused permit integration test passed
- `test/unit/scripted-workflow.test.ts` passed
- `npm run typecheck` passed
- `git diff --check origin/main..HEAD` passed
- fresh security/API review returned no issues

Fixes #1494.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@
 - Resolve package subagents from bare HTTP(S) Git URLs stored in Pi settings. Thanks to [@trancikk](https://github.com/trancikk) for #1452.
 
 ### Added
+- Add a package-internal one-use permit for one exact native child in a foreground `workflowScript`. Thanks to [@maroffo](https://github.com/maroffo) for #1494.
 - Add configured pruned fork sessions with budgeted transcript-overflow summaries, stable child-visible recovery refs, and private validated recovery sidecars while preserving normal fork links and safety behavior.
 - Load `workflowScript` from `workflowScriptPath` for execution, offline validation, and schedule creation. Thanks to [@elecnix](https://github.com/elecnix) for #1464.
 - Add offline `workflowScript` syntax and structural validation through the public subagent tool. Thanks to [@elecnix](https://github.com/elecnix) for #1462.
--- a/src/runs/foreground/execution.ts
+++ b/src/runs/foreground/execution.ts
@@ -106,6 +106,7 @@ import { appendTurnBudgetSystemPrompt, formatTurnBudgetOutput, initialTurnBudget
 import { initialToolBudgetState, toolBudgetState } from "../shared/tool-budget.ts";
 import { resolveWatchdogConfig } from "../../watchdog/settings.ts";
 import { agentDefinitionDigest, launchBindingDigest } from "../../shared/launch-contract.ts";
+import { consumeWorkflowChildPermit } from "../../shared/workflow-child-permit.ts";
 import { createBoundedByteTail, createBoundedLineReader, formatProtocolOutputLimit, MAX_CHILD_STDERR_BYTES, PI_AGGREGATE_EVENT_PROJECTOR, projectChildLifecycle, type ChildLifecycleAction, type ChildLifecycleState, type ProtocolOutputLimit } from "../shared/child-protocol.ts";
 import {
 	acceptChildWatchdogEvent,
@@ -540,8 +541,27 @@ async function runSingleAttempt(
 	let toolAvailabilityError: string | undefined;
 	let abortedBySignal = options.signal?.aborted === true;
 
+	const spawnSpec = getPiSpawnCommand(args);
+	if (options.workflowChildPermitLaunch) {
+		const permitError = consumeWorkflowChildPermit(options.workflowChildPermitLaunch.permit, {
+			workflowRunId: options.workflowChildPermitLaunch.workflowRunId,
+			childKey: options.workflowChildPermitLaunch.childKey,
+			agent: agent.name,
+			launchContractDigest,
+			context: options.context ?? "fresh",
+			runner: "pi",
+		});
+		if (permitError) {
+			cleanupTempDir(tempDir);
+			result.exitCode = 1;
+			result.error = permitError;
+			result.finalOutput = permitError;
+			progress.status = "failed";
+			progress.error = permitError;
+			return result;
+		}
+	}
 	const exitCode = await new Promise<number>((resolve) => {
-		const spawnSpec = getPiSpawnCommand(args);
 		const proc = spawn(spawnSpec.command, spawnSpec.args, {
 			cwd: options.cwd ?? runtimeCwd,
 			env: spawnEnv,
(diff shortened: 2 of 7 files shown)


maroffo:
@nicobailon Thanks for implementing and releasing this in 0.57.0.

One integration question: the permit currently appears package-internal. `createWorkflowChildPermit` is not exported, the structured delegation API cannot carry it, and the active `executeDelegated` entry point is not exposed.

Is there a supported same-process path for a trusted extension package to create and bind the permit, or is that carrier intentionally deferred? I want to avoid deep imports or any model-visible API.

PR #1600, opened 2026-08-27
fix: enforce globalConcurrencyLimit for workflowScript children

mateominato:
## Summary

- apply the existing per-run `globalConcurrencyLimit` semaphore to `workflowScript` children launched through `runs.run` / `runs.all`
- pass the configured limit through both foreground and detached workflow hosts
- re-check abort/stop state after a queued child acquires capacity
- document the expanded setting and add focused concurrency/validation tests

## Why

`globalConcurrencyLimit` currently reaches the durable legacy background runner, but `runWorkflowScript()` calls its host `launch` function directly. That leaves the primary modern orchestration path effectively unlimited even when operators configure a lower bound.

I reproduced this with 0.56.0 and 0.58.0: a `runs.all` workflow with four children and `globalConcurrencyLimit: 2` started all four child bash windows simultaneously. This is especially surprising because the config description presents the key as the global cap within one run.

## Validation

- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/scripted-workflow.test.ts` — **83 passed**
- `npm run typecheck` — pass
- full `npm run test:unit` — **2581 passed, 0 failed, 10 cancelled**; all cancellations are the pre-existing `herdr-inspector.test.ts` pending-promise group and are unrelated to this change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Show workflow child labels and phases in async status progress while preserving stable workflow keys.
 
 ### Fixed
+- Apply `globalConcurrencyLimit` to `workflowScript` children launched through `runs.run` and `runs.all`, not only legacy multi-child runners. Thanks to [@mateominato](https://github.com/mateominato) for #1600.
 - Ignore nested `.pi` and `sync-backups` directories during agent discovery so stale backup definitions cannot become executable agents. Thanks to [@arlishansenn](https://github.com/arlishansenn) for #1596.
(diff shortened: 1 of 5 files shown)


mateominato:
Thanks for merging this. VAPE currently pins npm:pi-subagents@0.56.0 and needs the workflow concurrency fix before running its cost benchmark. The current npm latest is 0.58.0, published before this merge and does not contain the scripted-workflow semaphore. Could you include #1600 in the next npm release and let us know the version when published?

PR #1663, opened 2026-08-28
fix: account for async workflow child usage

### LINKED ISSUE #1662: bug: /subagent-cost reports zero for completed async workflow children
## Summary

`/subagent-cost` reports `Children: 0` after a completed detached `workflowScript`, even though every child has complete usage in its persisted `_meta.json`.

Observed on `pi-subagents 0.58.0` with a six-child async workflow.

## Reproduction shape

1. Launch `subagent({ async: true, workflowScript: "...runs.all(...)..." })`.
2. Let the workflow complete and observe it through normal async completion / `subagent_wait`.
3. Run `/subagent-cost` in the owning parent session.

Actual:

```text
No subagent child usage found in this session.
Children: $0.0000
```

Persisted child metadata contains non-zero `input`, `output`, `cacheRead`, `cacheWrite`, `cost`, and `turns` for every child.

## Cause

The launch tool result is necessarily early and contains:

```text
toolName = subagent
details.results = []
```

The terminal observation uses:

```text
toolName = subagent_wait
details.completions[].results
```

`buildSubagentCostReport()` currently accepts only `toolName === "subagent"` and reads only direct `details.results`. It does not follow the terminal workflow receipt or child `_meta.json` artifacts. `subagent_wait`'s slim completion projection also omits `usage` and `sessionFile`.

Compaction is not the root cause: parent usage remains visible across compaction, while the async child usage was never attached to a direct `subagent.details.results` row.

## Expected

- Future async completions should carry a bounded usage projection per child.
- `/subagent-cost` should consume `subagent_wait.details.completions`.
- For already persisted async workflows, the command should use the workflow receipt's child run IDs and bounded `_meta.json` files instead of reporting a false zero.
- Parent compaction usage should be included in the parent total.

I have a narrow tested patch ready with a regression test for the async launch → receipt/meta path.

Geraldo-Morais:
Fixes #1662.

## Problem

Detached `workflowScript` launches persist an initial `subagent` tool result with `details.results = []`. Terminal children arrive through `subagent_wait.details.completions`, while `/subagent-cost` only inspected direct `subagent.details.results`. Completed async workflows therefore reported a false `Children: $0` despite complete child `_meta.json` accounting.

Parent compaction calls were also omitted because their usage lives on `type: "compaction"` entries rather than assistant messages.

## Change

- carry the bounded `Usage` and `sessionFile` projection in future wait completions;
- consume both direct `subagent` results and `subagent_wait` completions;
- discover only workflow IDs already present in the owning session branch;
- recover historical async child usage through the terminal workflow receipt and exact run/agent metadata path;
- deduplicate children by persisted run/session identity;
- report unresolved async metadata instead of silently treating it as zero;
- include parent compaction usage without counting it as an assistant turn.

The receipt/meta fallback is bounded: no recursive artifact scan is performed, run IDs are path-safe validated, and metadata must match both expected run ID and agent.

## Validation

- `npm run typecheck`
- targeted unit tests:
  - `test/unit/wait-completions.test.ts`
  - `test/unit/completion-replay.test.ts`
- targeted integration test:
  - `test/integration/slash-commands.test.ts`

The new integration regression exercises the actual async shape: empty launch results → terminal workflow receipt → persisted child metadata → non-zero `/subagent-cost`, including compaction and aggregate totals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
 - Give `runs.lanes(...)` stage-0 retained-resume validation actionable `runs.run(...)` guidance instead of a generic error (#1657).
 - Remove the remaining `lane:` prefix from operator-facing async status rows in the TUI (#1658).
 - Allow scheduled project roots shared through a registered Git worktree's `.pi` symlink while rejecting unrelated or unproven Git-layout escapes. Thanks to [@sususu98](https://github.com/sususu98) for #1656.
+- Include delegated child usage in subagent tool results and `/subagent-cost`, including completed async workflow children and parent compaction usage with persisted workflow-receipt recovery when needed (#1662, #1666). Thanks to [@Geraldo-Morais](https://github.com/Geraldo-Morais) for #1662 and [@jf88888](https://github.com/jf88888) for #1666.
 - Avoid false preflight mismatch warnings for generated `runs.lanes(...)` stage keys (#1649).
 - Accept long host tool-call ids in workflow child summaries, matching the existing 4,096-byte session id bound. Thanks to [@SudoKillMe](https://github.com/SudoKillMe) for #1653.
 - Keep an async `workflowScript` continuation live while an awaited child coordinates with its supervisor, so sequential tail steps continue after the child settles (#1634).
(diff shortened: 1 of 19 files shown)


Geraldo-Morais:
Validated the maintainer-updated head 611280d on Windows. The receipt fallback now exercises and sums every continuation.runIds attempt, and the changelog includes the requested contributor attribution. npm run typecheck passed; 78 focused unit tests passed; the new /subagent-cost receipt/metadata integration regression passed. The six remaining /run assertions in the full slash-commands file reproduce unchanged on current main (1f2abe1d), so they are baseline/environmental rather than introduced by this PR. Greptile's latest review is 5/5.

PR #1665, opened 2026-08-28
test(windows): hide child process console windows

### LINKED ISSUE #1664: test(windows): full suite repeatedly opens visible terminal tabs
## Summary

Running the full `test:unit` / `test:integration` suites from a headless Pi session on Windows repeatedly opens visible blank Windows Terminal tabs (often titled with the Git Bash `mingw64` path). The tabs appeared every few seconds while process-spawning tests ran and stopped when the test command ended/timed out.

Observed on Windows with Node 24 and Windows Terminal configured as the default console host.

## Evidence

During the popup window, process readback showed children rooted in the pi-subagents test checkout, for example:

```text
node .../pi-subagents/src/runs/background/subagent-runner.ts
     .../pi-subagents-test-root-*/async-cfg-*.json
```

with short-lived PowerShell/conhost and shell descendants. This lines up with the repeated process-spawning tests, not an unrelated application.

Production nested Pi spawn paths already use `windowsHide: true`, and `windows-hide-spawn.test.ts` covers those two paths. The test suite itself still has many direct `spawn`, `spawnSync`, `execFileSync`, and shell calls without `windowsHide`; when the suite is launched from a process without an attached console, Windows Terminal may create a visible tab for each console child.

## Expected

The test harness should be terminal-silent on Windows without requiring every individual test fixture to remember `windowsHide: true`.

## Suggested fix

Install a Windows-only test preload from the existing `isolated-temp-root.mjs` / `register-loader.mjs` path that wraps Node child-process APIs and defaults `windowsHide: true` while preserving explicit options and overloads. Keep production behavior unchanged and add a focused regression test for option normalization.

Geraldo-Morais:
Fixes #1664.

## Problem

The full test suites repeatedly launch child processes. On Windows hosts that use Windows Terminal as the default console host, test fixtures that omit `windowsHide` can create visible blank terminal tabs every few seconds. This is especially disruptive when the suite runs from a headless Pi tool call.

Production nested Pi paths already set `windowsHide: true`; the remaining gap is the test harness and its many direct child-process fixtures.

## Change

- install a Windows-only child-process guard from the existing unit/integration preload path;
- default `windowsHide: true` for `spawn`, `spawnSync`, `exec`, `execSync`, `execFile`, `execFileSync`, and `fork`;
- preserve overloads, callback placement, existing option objects, and caller input immutability;
- sync patched Node builtin ESM exports so named imports used by tests see the same guarded functions;
- keep non-Windows behavior unchanged and install idempotently.

This avoids a broad edit across dozens of process-oriented fixtures and keeps the policy centralized in test-only code.

## Validation

- `npm run typecheck`
- `test/unit/windows-hide-test-processes.test.ts`
- `test/unit/windows-hide-spawn.test.ts`
- representative process tests:
  - `test/unit/close-grace-timer.test.ts`
  - `test/unit/session-lease.test.ts`

All focused tests passed on Windows. No production source is changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 - Show workflow child labels and phases in async status progress while preserving stable workflow keys.
 
 ### Fixed
+- Keep the Windows test harness from opening visible terminal tabs for child-process fixtures (#1664).
 - Remove the remaining `lane:` prefix from operator-facing async status rows in the TUI (#1658).
 - Avoid false preflight mismatch warnings for generated `runs.lanes(...)` stage keys (#1649).
 - Accept long host tool-call ids in workflow child summaries, matching the existing 4,096-byte session id bound. Thanks to [@SudoKillMe](https://github.com/SudoKillMe) for #1653.
--- a/test/support/isolated-temp-root.mjs
+++ b/test/support/isolated-temp-root.mjs
@@ -1,6 +1,7 @@
 import * as fs from "node:fs";
 import * as os from "node:os";
 import * as path from "node:path";
+import "./windows-hide-child-processes.mjs";
 
 if (!process.env.PI_SUBAGENTS_TEMP_ROOT) {
 	const tempRoot = fs.mkdtempSync(path.join(os.tmpdir(), "pi-subagents-test-root-"));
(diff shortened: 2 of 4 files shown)


Geraldo-Morais:
Closing after better process correlation. The recurring visible Git Bash tabs were produced by a different active Pi session running ssh openclaw-vps on an approximately five-second loop, not demonstrably by this repository's test harness. Although this PR is defensive, its motivating causal claim was incorrect and the global child_process monkeypatch is not justified without an independent reproduction.

PR #1688, opened 2026-08-29
Animate running spinner on wall clock instead of render events

Pudgey:
The running spinner (braille frames in FleetView and inline status rows) only advances when a render event changes its seed/frame. During quiet stretches — a child thinking, a long tool call — the dots freeze mid-frame and read as a hang even though the run is healthy.

This PR makes the spinner free-running on wall clock:

- `runningGlyph` derives its frame from `Date.now() / 125` (~8 fps), keeping the existing `seed` term so parallel rows stay phase-offset rather than spinning in lockstep.
- FleetView's refresh gate (`renderKey === lastRenderKey` early return) now requests a repaint on its existing 500ms tick while any entry is `running`, so the clock-driven frame actually gets painted between state changes. Idle fleets keep the early return — zero extra renders when nothing runs.

No new timers, no API changes; running this locally the fleet reads as alive the whole way through a 5-lane parallel run instead of stuttering with model activity.


nicobailon:
Fixed the Greptile test-gate finding in 12e5b6d705756dfc82cd77d40b7c3dd65cb80fcc. Added deterministic tests for wall-clock glyph progression and unchanged running-vs-queued repaint behavior, plus the changelog credit for @Pudgey. Focused unit/integration tests and typecheck pass locally.

PR #1711, opened 2026-08-29
fix: hide Windows resume runner consoles

Zethu5:
## Problem

Retained-session resume runners flash short console windows on Windows.

Fresh workflow children do not show this behavior. The problem occurs when resume starts an outer async runner.

I reproduced the problem with five retained `worker` resumes. Each resume caused a visible console flash.

The outer runner used `detached: true` with `windowsHide: true`. Windows can create a new console before it applies the hidden state.

## Fix

Use `detached: false` for the outer async runner on Windows.

Keep these existing behaviors:

- Keep `windowsHide: true` on Windows.
- Keep `detached: true` on Unix.
- Keep `proc.unref()` for background execution.
- Keep the startup handshake and file control channel.
- Keep stop and resume behavior unchanged.

A small helper makes the platform rule easy to test.

## Verification

- Repeated the retained-resume test after the local patch.
- Confirmed that no console window appeared.
- Ran the focused platform tests: 2 passed.
- Ran the retained-resume workflow test separately: passed.
- Ran TypeScript type checking: passed.
- Ran `git diff --check`: passed.
- Ran the full unit suite: 2,672 passed, 20 failed, and 34 skipped.

The observed full-suite failures do not cover the changed helper or launch site. This Windows environment cannot create test symlinks, and the concurrent suite reports unrelated package fixture failures.

## Risk

The focused test verifies process options. It does not inspect Windows window visibility.

Manual Windows verification covers the visible behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Clarify workflowScript portability and runs.host working-directory limits, including the outer workflow cwd and trusted `cd ... && command` patterns (#1679).
 
 ### Fixed
+- Keep retained-session resume runners from flashing console windows on Windows while preserving Unix background detachment.
 - Hide mutation-evidence Git subprocess windows on Windows to prevent visible console flashes or terminal tabs. Thanks to [@dnnkeeper](https://github.com/dnnkeeper) for #1706.
 - Deliver immediate async workflow terminal failures on macOS without requiring a later status refresh (#1700).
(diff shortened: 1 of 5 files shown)


infectiousstupidity:
Haven't explored and pinpointed this issue myself, but I experience a lot of flashing console windows on Windows 11 too when the pi-subagent tool is doing something.

PR #1730, opened 2026-08-30
perf: reuse staged workflow widget projection

### LINKED ISSUE #1727: Avoid repeated staged workflow projection during widget rendering
Recent performance review of the last-day commits found avoidable repeated workflow projection work in the async widget render path. This is especially relevant after #1714, which added staged workflow visibility to the live widget.

Source evidence:

- `src/tui/render.ts` `workflowGraphStageNodes()` recursively flattens workflow graph nodes.
- One collapsed staged-workflow render can call this graph projection repeatedly through:
  - `widgetStats()` / `workflowStageProgress()`
  - `buildSingleWidgetLines()` count selection
  - `foregroundStyleWidgetDetails()` / `workflowWidgetSteps()`
  - `foregroundStyleWidgetDetails()` planned-key construction
  - `fitAdaptiveWidgetLines()` staged-workflow layout exception
- `workflowWidgetSteps()` also builds maps/sets, merged step objects, and sorts rows for planned/loaded stages.
- `widgetRenderKey()` is computed before and after status refresh in `src/runs/background/async-job-tracker.ts`, and can repeat graph flattening plus lane-step derivation.

Recommended fix shape:

- Compute flattened workflow stages, stage progress, planned-key set, and merged widget steps once per synchronous widget render/key calculation.
- Pass that local projection through `widgetStats()`, `foregroundStyleWidgetDetails()`, count selection, lane selection, and adaptive-layout checks.
- Keep the projection local to one render/key calculation. Avoid persistent caches that could make active-stage visibility stale.
- For bounded public snapshots only, consider stopping ordinary child projection once `maxChildrenPerNode` visible capacity is filled while still counting omitted inputs.
- For host widget rows, validate/count all host nodes but format only the first eight displayed rows.

Acceptance:

- Preserves `Plan:` display.
- Preserves active/current stage priority and attention/failure rows.
- Preserves correct stage numbering and hidden-line behavior.
- Does not cap `projectAsyncWorkflowRows()` generically because Fleet currently relies on full rows.
- Adds focused tests around staged workflow widget output to prove visible output is unchanged.

Non-goals:

- Do not pre-budget collapsed staged rows before priority ordering unless separately benchmarked and tested.
- Do not change FleetView's full-detail contract.

kkkhs:
## What
Fixes #1727. Collapsed staged workflow widgets were rebuilding the same workflow projection several times during one render/key calculation.

## Fix
Add a local `WorkflowWidgetProjection` lookup that reuses flattened stages, merged widget steps, stage progress, and planned keys while keeping the projection scoped to the current render/key calculation.

## Test
Adds regression coverage for staged workflow widget output and render-key generation to assert the workflow graph nodes are read once.

### DIFF
--- a/src/runs/shared/workflow-graph.ts
+++ b/src/runs/shared/workflow-graph.ts
@@ -15,7 +15,8 @@ export interface WorkflowGraphBuildInput {
 
 /** Return displayable workflow stages while hiding structural parallel groups and host monitors. */
 export function workflowGraphStageNodes(graph: WorkflowGraphSnapshot | undefined): WorkflowGraphNode[] {
-	if (!graph?.nodes?.length) return [];
+	const nodes = graph?.nodes;
+	if (!nodes?.length) return [];
 	const stages: WorkflowGraphNode[] = [];
 	const visit = (node: WorkflowGraphNode): void => {
 		if (node.kind === "parallel-group" || node.kind === "dynamic-parallel-group") {
@@ -24,7 +25,7 @@ export function workflowGraphStageNodes(graph: WorkflowGraphSnapshot | undefined
 		}
 		if (node.kind !== "host-step") stages.push(node);
 	};
-	for (const node of graph.nodes) visit(node);
+	for (const node of nodes) visit(node);
 	return stages;
 }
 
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks again for this contribution. I adopted the implementation by cherry-picking your commit with original authorship, added changelog credit, and landed it through #1750. Closing this draft as superseded by the merged maintainer PR.

PR #1757, opened 2026-08-30
fix: isolate corrupt async status during restore

### LINKED ISSUE #1756: Corrupt async status.json prevents restoration of all active async jobs
## Summary

A single corrupt persisted async `status.json` can make active-job restoration throw before the remaining valid runs are restored.

I encountered this on Windows with `pi-subagents` 0.60.0. One stale run was still present in `.active-runs`, while its `status.json` consisted entirely of NUL bytes. On startup, `restoreActiveAsyncRuns()` reached `listAsyncRuns()` / `reconcileAsyncRun()`, `readStatus()` threw while parsing that run, and the exception aborted restoration of the complete active-run set.

## Environment

- OS: Windows
- `pi-subagents`: 0.60.0
- Pi coding agent: 0.84.3

## Observed error

```text
Failed to restore active async jobs from '<temp>/async-subagent-runs':
Error: Failed to parse async status file
'<temp>/async-subagent-runs/<run-id>/status.json':
Unexpected token '', ""... is not valid JSON

    at readStatus (.../src/shared/utils.ts)
    at reconcileAsyncRun (.../src/runs/background/stale-run-reconciler.ts)
```

Inspection showed that the affected file had a nonzero length but every byte was `0x00`. The recorded runner PID was no longer alive.

## Impact

The corrupt run was not isolated. Because the read/reconcile exception escaped the per-entry loop in `listAsyncRuns()`, one bad persisted run prevented unrelated valid active runs from being restored.

Atomic status writes reduce the chance of partial JSON, but they cannot eliminate pre-existing corruption or every filesystem/crash failure mode. Startup recovery still needs a per-run corruption boundary.

## Minimal reproduction shape

Under an async run root:

```text
async-subagent-runs/
  .active-runs/
    good-run
    bad-run
  good-run/
    status.json   # valid running AsyncStatus
  bad-run/
    status.json   # NUL bytes or otherwise invalid JSON
```

Calling the active-run listing/restoration path currently throws on `bad-run`, so `good-run` is not returned/restored.

## Expected behavior

A malformed persisted status for one run should not prevent restoration of other runs:

1. Emit an actionable diagnostic naming the unreadable run.
2. Treat that run as corrupt/terminal or otherwise quarantine it.
3. Remove or release its stale active-run marker when safe.
4. Preserve the run directory and logs for diagnosis.
5. Continue restoring all other valid active runs.

## Suggested regression test

Create one valid active run and one active run whose `status.json` contains invalid JSON/NUL bytes. Assert that:

- the valid run is still listed/restored;
- the corrupt run is omitted or projected as a bounded corrupt terminal record;
- the corrupt active marker is released or quarantined according to the chosen policy;
- a diagnostic is emitted;
- the original corrupt run directory is retained.

## Related existing hardening

The changelog notes that async status/result JSON writes became atomic in 0.12.0 and that `readStatus()` now preserves parse failures with context. This report concerns the remaining recovery behavior: a real parse failure is correctly surfaced, but currently aborts restoration of the entire active set instead of being isolated to one run.

zhexulong:
## Summary

- isolate corrupt active-run `status.json` failures during async run listing;
- release the affected active-run marker while preserving the run directory;
- continue restoring unrelated valid active runs;
- retain strict failures for repair scans, targeted runs, and semantic status validation;
- add regression coverage for both the run-list and job-tracker restore paths.

Fixes #1756

## Validation

- `npm run typecheck`
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/async-status.test.ts test/integration/async-job-tracker.test.ts`

### DIFF
--- a/src/runs/background/async-status.ts
+++ b/src/runs/background/async-status.ts
@@ -173,6 +173,10 @@ function isNotFoundError(error: unknown): boolean {
 		&& (error as NodeJS.ErrnoException).code === "ENOENT";
 }
 
+function isAsyncStatusParseError(error: unknown): boolean {
+	return error instanceof Error && error.message.startsWith("Failed to parse async status file '");
+}
+
 function isAsyncRunDir(root: string, entry: string): boolean {
 	const entryPath = path.join(root, entry);
 	try {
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks @zhexulong for the report and the original fix in this PR. I landed the maintainer replacement in #1758, with the same behavior covered more completely and visible changelog credit preserved for you. Closing this PR as superseded by #1758.

PR #1760, opened 2026-08-30
feat(workflow): add per-run capacity overrides

RapierCraft:
## Summary
- expose workflow-only `globalConcurrencyLimit` and `maxSubagentSpawnsPerRun` overrides
- preserve explicit-call precedence in foreground and async workflow paths
- keep inherited nested run-fanout descriptors authoritative
- strip capacity controls from child defaults and reject them in `runs.run`/`runs.all` items

## Verification
- `npm run typecheck`
- `npm test` (2799 tests; 2793 passed, 4 skipped, 0 failed)
- focused schema, public-normalization, foreground, async, and inheritance tests
- independent API/correctness and authority reviews found no patch-caused blockers


nicobailon:
Thanks [@RapierCraft](https://github.com/RapierCraft) for #1760. I landed the smaller maintainer-owned replacement in #1761, preserving the workflow capacity override behavior while folding the coverage into existing tests. Closing this PR as superseded so we do not mutate your branch.

PR #1770, opened 2026-08-31
feat: add acceptance.report structured delivery switch

mapleluvr:
## Summary

Adds an opt-in `acceptance.report: "on" | "off"` switch to the acceptance config that changes **how** the child delivers its acceptance report: as schema-validated structured output instead of a fenced JSON text block parsed from prose.

Motivation: the fenced-JSON report channel is parsed from the child's final text with a tolerant regex pipeline (enum synonym folding, wrapper keys, underscore fence tags). A malformed or missing block fails late and ambiguously. `structured_output` delivery already exists for `outputSchema` deliverables — this change reuses that machine-validated channel for the acceptance report, with format failures surfaced in-session (cheap retry) instead of post-hoc parsing.

## Behavior

| `acceptance.report` | Delivery | Validation | Termination |
|---|---|---|---|
| omitted / `"off"` | fenced `acceptance-report` JSON block (legacy) | post-hoc regex parse (unchanged) | unchanged |
| `"on"` without `outputSchema` | standalone `structured_output` tool, `value` = report | real JSON Schema (enum-strict `criteriaSatisfied[].status`, `commandsRun[].result`) | only a valid delivery terminates the step |
| `"on"` with `outputSchema` | single `structured_output` with `value` + `acceptanceReport` | deliverable + report both schema-validated | missing/invalid report throws in-session; child corrects and re-calls |

## Implementation notes

- `resolveAcceptanceReportMode()` resolves the switch once; foreground (`subagent-executor.ts`) and async (`async-execution.ts`) orchestration create a **report-only runtime** when `"on"` and no `outputSchema` exists (no schema file materialized; report capture side file only).
- Child side (`subagent-prompt-runtime.ts`) registers the standalone tool when `PI_SUBAGENT_ACCEPTANCE_REPORT_REQUIRED=1` and the structured-output env vars are absent; the piggyback branch validates `acceptanceReport` against the real report schema and refuses delivery without it when required.
- Parent read sides (`execution.ts` foreground, `subagent-runner.ts` async) gain a `reportOnly` branch: skip `readStructuredOutput` (output.json is never written), read the report capture; missing required report fails the run with `MISSING_ACCEPTANCE_REPORT_ERROR`.
- `formatAcceptancePrompt` emits mode-correct instructions (standalone "value is the report" wording vs piggyback "include acceptanceReport" wording) and drops the fenced template for structured delivery.
- The report schema includes the `notes` field accepted by the legacy parser (`manualNotes` alias).
- Backward compatibility: omitted `report` leaves every default template and behavior byte-identical.

## Testing

- New `test/unit/acceptance-report-mode.test.ts` (15 tests): mode resolution matrix, report-only/piggyback runtime shapes, schema validation positive/negative, prompt templates for all four mode combinations, `notes` schema acceptance.
- `npm run typecheck` passes.
- Full unit suite: normalized failure-set diff vs clean HEAD shows zero new failures.

### DIFF
--- a/docs/design/acceptance-report-structured-channel.md
+++ b/docs/design/acceptance-report-structured-channel.md
@@ -1,0 +1,199 @@
+# Design: Acceptance Report Delivery via `acceptance.report` Switch (tool + prompt)
+
+```text
+Status:   draft — implementation not started
+Scope:    local change to pi-subagents (this repository), not upstream yet
+Context:  evolved from issue #499 / PR #577 post-mortem discussions (2026-08)
+```
+
+---
+
+## 1. Problem
+
+Acceptance reports today reach the parent through two mechanisms:
+
+1. **Fenced JSON text (default)** — `formatAcceptancePrompt` (src/runs/shared/acceptance.ts:407)
+   injects a ` ```acceptance-report ` fence template into the child prompt; the parent recovers it
+   with a fault-tolerant regex/normalization pipeline (acceptance.ts:487–749: wrapper-key aliases,
+   inline `ACCEPTANCE_REPORT:` markers, fence-scanning fallbacks). This is the #499 lesion:
+   free-text delivery means fields can be dropped, and malformed reports become parse errors.
+2. **`structured_output` piggyback (conditional)** — when the task already has `outputSchema`,
+   the child-side `structured_output` tool grows an optional `acceptanceReport` parameter
+   (structured-output.ts:64, subagent-prompt-runtime.ts:735–757), validated as an opaque object
+   and captured to `acceptance-report.json` (structured-output.ts:140). The parent prefers this
+   capture over text parsing (execution.ts:1479–1481, 2159–2160).
+
+The gap: the piggyback path only exists when the parent supplies an `outputSchema`. Tasks
+**without** `outputSchema` are locked to the text channel regardless of harness capability.
+The piggyback `acceptanceReport` parameter is also only `type: "object"` — it is never
+validated against the acceptance report shape.
+
+## 2. Goal
+
+Add a task-level switch `acceptance.report: "on" | "off"` (default unchanged → current behavior):
+
+- **`on`** — the child MUST deliver the acceptance report through the structured channel
+  (`structured_output` tool with a schema-validated `acceptanceReport` parameter) **plus** the
+  text prompt contract; if the task has no `outputSchema`, a dedicated `structured_output`
+  tool is still registered solely to carry the report.
+- **`off`** — equivalent to today's no-`outputSchema` behavior: no structured channel for the
+  report; the text fence channel remains the only self-report path (and hosts that prefer
+  verify/gate-only evidence can stop requesting child reports at the `acceptance` level).
+
+Key rules from the design discussion:
+
+- **The switch has exactly two values.** No `prompt`/`tool` split: `on` means
+  tool *and* prompt together (prompt explains the contract, tool enforces it); `off` means the
+  report is delivered only if an existing text/manual path already would be, with no new
+  machinery.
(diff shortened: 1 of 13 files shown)


nicobailon (changes requested):
Thanks for the PR. I took a fresh pass against current `main`; this should not merge as-is yet.

Concrete blockers I see:

1. `acceptance.report: "off"` with `outputSchema` still creates structured acceptance-report capture and prompts/tool schema. That contradicts the documented `off` behavior, which should force the legacy fenced-text channel.

2. The structured acceptance-report schema accepts `{}`. That can terminate in-session even though the parent acceptance parser later rejects the same empty report as semantically invalid.

3. Required piggyback reports are not fail-closed at the parent read boundary. If `output.json` exists but the required acceptance-report capture is missing or unreadable, foreground/async paths can continue past structured-output validation and fail later with a weaker prose/attestation parse path.

4. `docs/design/acceptance-report-structured-channel.md` is a stale draft plan in shipped docs. It says implementation has not started, while this PR implements the feature and `package.json` publishes `docs/**/*`.

Suggested shape: resolve report mode once, suppress structured capture/prompt/tool augmentation for `report: "off"`, enforce the same semantic acceptance-report validity before `structured_output` termination, fail closed on missing required piggyback captures in both foreground and async readers, and remove or rewrite the stale draft doc. Please add focused regressions for `outputSchema` + `report: "off"`, empty structured reports, and missing required piggyback capture.

nicobailon:
Thanks for putting this together. I am going to close this PR rather than keep iterating on it.

The useful idea is the explicit `acceptance.report` switch, but this implementation is too broad for that contract: 13 files, +2481/-559 lines, including large rewrites in `acceptance.ts` and `types.ts`, a new report-only channel, and a committed draft design doc. That scope does not earn its size under this repo's review policy, and the current head also has concrete correctness blockers around `report: "off"`, empty structured reports, and required piggyback report enforcement.

If we pursue this, the smaller shape should be a narrow maintainer-owned replacement: resolve report mode once; for `outputSchema` runs only, make `report: "off"` disable structured acceptance capture/prompting, and make `report: "on"` require and validate the piggyback `acceptanceReport`. Avoid a standalone report-only channel and broad type/doc rewrites unless there is a separate accepted design for them.

Thanks [@mapleluvr](https://github.com/mapleluvr) for the proposal. I will preserve credit if a smaller replacement lands.

mapleluvr:
Thanks for turning this around so quickly — and for the credit in the changelog and commit message.

Reviewed the landed implementation (7281103); the narrow shape is clearly the right call. Two parts of it I want to call out as genuinely better than my original:

1. Enforcing the required report at the tool-schema level (`required: ["value", "acceptanceReport"]`) instead of throwing inside `execute` — invalid submissions never even reach the tool body, which kills the invalid-write-then-retry window my version had.
2. Preflight rejecting `report: "on"` without `outputSchema` (`acceptance.report requires outputSchema`) — that removes the whole standalone-channel question at the validation boundary instead of gating it at runtime.

One observation for future reference: my original patch's large diff was mostly incidental reformatting from a local automated formatter layered over the real changes (~1.9k of the 2.4k lines), which made the semantic diff much harder to weigh than it should have been. Noted for any future contributions — they'll come in smaller and formatter-clean.

One small follow-up I hit while using this today, unrelated to the switch itself: when a model passes `acceptance` as a *stringified* JSON object (e.g. `"{\"level\": \"checked\"}"`) — which models occasionally do when constructing complex tool args — the anyOf schema rejects it with five branch errors and no hint about the actual mistake. A one-line addition to the schema description ("must be an inline object, not a JSON-encoded string") or a tolerant coerce-and-continue would save the retry. Filed #1781 with reproduction details.

PR #1790, opened 2026-08-31
feat: add focused native child sessions

stekman08:
## Problem

Fleet is useful for seeing native async children, but interacting with one currently means leaving the roster and using separate steering commands. That makes a live child feel like a log row rather than a Pi session, especially when a workflow has nested children.

## Vision

Keep Fleet as the lobby, then let an operator enter a focused, lightweight native child session without cloning Pi's TUI:

```text
Fleet
  workflow
    delegate
      worker                 Enter

╭────────────────────────────────────────────────────────────────╮
│ worker · 4d21a13f · child 0                                    │
│ Model: openai/gpt-5.4 · Thinking: high · Status: applied ...    │
├────────────────────────────────────────────────────────────────┤
│ ...live structured child transcript...                         │
├────────────────────────────────────────────────────────────────┤
│ Type a normal message to this child                            │
│ Esc Fleet · Ctrl+L model · Shift+Tab thinking · Enter submit   │
╰────────────────────────────────────────────────────────────────╯
```

## Bounded v1

- `Enter` on a live native async child opens a full-width focused transcript and Pi `CustomEditor`.
- Normal editor submission sends direct input through the existing acknowledged child-control inbox; no `s` prefix is required.
- `Esc` returns to Fleet and preserves the selected child.
- Pi's configured model-select and thinking-cycle keybindings open native `SelectList` pickers.
- Model/thinking changes are explicit control requests queued until the next safe turn boundary, with queued/applied/failed acknowledgements projected into status and the focused header.
- Child-side validation uses launch-resolved policy: model scope, registry availability, credentials, model thinking support, and the configured thinking ceiling. Fleet cannot widen those constraints.
- Nested native runs remain separate selectable Fleet rows, including workflow -> delegate -> sub-subagent trees.
- Missing focused-view APIs fail gracefully: Pi shows a notification and reopens Fleet at the same selection.

## Reuse and optionality

This builds on existing pieces rather than introducing a second TUI stack:

- Pi `CustomEditor`, `SelectList`, keybindings, and `keyHint`
- existing Fleet transcript parsing/rendering and trusted-path checks
- existing async status lifecycle and steer/follow-up control directory
- existing model-scope, model-registry, and thinking-ceiling helpers

The behavior is opt-in: nothing changes until an operator presses `Enter` on an eligible live native child. Existing Fleet controls and steering remain available.

## Tests and validation

- Focused unit set: **148 passed**
  - direct input routing and correlated delivery receipts
  - runtime request parsing and queued/applied/failed acknowledgements
  - safe settlement boundaries, including legacy Pi runtimes without `agent_settled`
  - model availability/credential, scope, thinking-support, and ceiling rejection
  - focus, submit, escape, Pi keybindings, ANSI layout, header refresh, API recovery, and selection preservation
  - native-only gating and nested-child selection
- `npm run typecheck`: passed
- `npm run test:integration`: **904 passed, 6 skipped, 1 pre-existing failure**
  - the failing timeout-acceptance assertion was reproduced unchanged at base commit `a9b17bb7`; it expects absent acceptance data while the current runner records a rejected timeout check
- `npm run test:e2e`: completed; real Pi-session coverage was skipped because the optional Pi runtime packages are unavailable in this test environment
- `git diff --check`: passed
- LSP: edited production files and newly added tests are clean; inferred-project warnings in an existing large test file predate this change
- Pi-Lens: no introduced findings; the project-wide scan reports existing baseline findings in previously large modules

The package has no `build` script; `typecheck` is its compile-time gate.

## Limitations

This intentionally excludes external CLI children, Herdr integration, full slash-command handling, attachments, shell mode, session-tree features, and resuming completed sessions. Runtime changes apply only at safe next-turn boundaries, not during an in-flight provider request.


nicobailon:
Thanks for the thoughtful PR, @stekman08. I am going to close this version rather than merge it as-is.

The user value is real: Fleet should make it easy to drop into a running native child and send focused guidance. The shape here is broader than I want to own in one PR, though. It adds a second in-Fleet focused TUI, a new child-control protocol, runtime model/thinking mutation, model-scope serialization through every async child/status record, and new ordering/state paths in the runner. That puts extra weight on hot async status/control paths even when the feature is not used.

Preferred shape: reuse the existing Herdr inspector as the focused child pane. Fleet stays the lobby; pressing into a child should open the child-specific Herdr inspector. That inspector can accept plain input as child guidance and keep explicit colon/command controls like status/stop. Runtime model/thinking changes should be a separate design if we decide to own them later.

I will preserve the useful idea and credit you in the smaller maintainer implementation.

stekman08:
@nicobailon

Thanks for picking this up so quickly. I scoped Herdr out too early and missed how much of the child-specific inspector path was already there, so I ended up building a much more elaborate solution than necessary. :)

I use and love Herdr, and given that the integration already existed, your solution is much more straightforward. I really like this direction. Thanks for preserving the idea and the credit!

PR #1793, opened 2026-09-01
fix: keep macOS subagent tasks out of argv

josephkallas:
## Summary

- keep subagent task content out of process arguments on macOS
- pass macOS tasks through Pi's existing `@file` path, regardless of size
- retain the existing 8 KB threshold on Linux and Windows
- add platform-explicit tests for Darwin, Linux, and Windows

## Motivation

On an enterprise-managed macOS host, child Pi processes were intermittently terminated before producing any model/tool activity. The correlated system log was:

```text
AppleSystemPolicy: ASP: Unable to apply provenance sandbox: ..., /opt/homebrew/Cellar/node/24.8.0/bin/node
```

The failures were reproducible when structured/code-like task content was present in the child argv and disappeared when the same task was delivered through the existing `@task.md` mechanism. Keeping prompt content out of argv also prevents shell metacharacters and potentially sensitive task text from being exposed to process-argument inspection.

This does not introduce a shell wrapper or disable endpoint security. Foreground and background launches continue to use argument-vector spawning with `shell: false` behavior.

## Validation

- `node --experimental-strip-types --test test/unit/pi-args.test.ts test/unit/subagent-startup-retry.test.ts`
  - 56 passed, 0 failed
- `npm test`
  - 1474 passed, 0 failed, 1 skipped
- live macOS subagent checks:
  - scout with inferred acceptance: completed
  - reviewer/high with inferred acceptance: completed

## Scope

The behavioral change is macOS-only. Linux and Windows continue to inline tasks up to the existing 8 KB threshold.

### DIFF
--- a/src/runs/shared/pi-args.ts
+++ b/src/runs/shared/pi-args.ts
@@ -16,6 +16,12 @@ import { PI_CODING_AGENT_PACKAGE_ROOT_ENV } from "../../shared/utils.ts";
 import { SUBAGENT_CAPABILITY_CEILING_ENV, decodeSubagentCapabilityCeiling, encodeSubagentCapabilityCeiling, intersectSubagentCapabilityCeilings, type ResolvedSubagentCapabilityCeiling, type SubagentCapabilityAudit } from "./capability-ceiling.ts";
 
 const TASK_ARG_LIMIT = 8000;
+
+// Keep code-like task content out of macOS exec arguments, which endpoint-security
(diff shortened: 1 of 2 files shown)


nicobailon:
Superseded by #1795, which landed the macOS task-file behavior on current main with CI/Greptile passing. Credit is preserved for @josephkallas in the changelog and squash commit body. Thank you for the original contribution.

PR #1836, opened 2026-09-03
Graceful workflow teardown: drain-and-flush all-terminal workflows on session replacement

### LINKED ISSUE #1833: Workflow controllers hard-abort on extension-session replacement even when every child run is already complete
**Repo:** pi-subagents `0.64.0` (host: pi-web-ui `0.62.1`, pi core `0.84.4`)
**Type:** Robustness / graceful degradation
**Verified against:** pi-subagents 0.64.0 — the abort loop below is still at `src/extension/index.ts:967`, byte-identical to 0.62.0.

## Summary

When the extension session is replaced or reloaded (e.g. the host disposes the conversation runtime on a project switch), `cleanup()` aborts all live workflow controllers:

```ts
// src/extension/index.ts (cleanup)
for (const controller of state.workflowControllers?.values() ?? []) {
    if (!controller.signal.aborted) {
        controller.abort(new Error("Workflow stopped because the extension session was replaced or reloaded."));
    }
}
```

The defensive intent is sound — a reloaded session must not launch children through a stale context. But the abort is unconditional, including the case where **all of a workflow's child runs have already finished** and the wrapper is merely in its assembly/post-processing phase. Aborting there destroys the wrapper's result after 100% of the expensive work is done.

## Observed incident

A 3-lane `workflowScript` (parallel `runs.all` fan-out, then assembly) was in its assembly phase when the host replaced the extension session. All three lanes had completed. The abort discarded the assembled result. Nothing *downstream* was lost thanks to existing durability:

- child outputs persisted at `~/.pi/agent/sessions/<project>/subagent-artifacts/<runId>_*_output.md`
- managed-worktree cleanup auto-ff'd `main` through completed children

but recovery was fully manual: run ids vanished from `subagent` status ("Async run not found"), and lane branches had to be identified and merged by hash.

## Proposal (any one of these would have saved the run)

1. **Drain-and-flush:** in `cleanup()`, if a workflow controller has no non-terminal child runs left, let its script run to completion (bounded, e.g. a short grace timeout) so it can finish assembly and emit its result/wake, then tear down. Only abort controllers that could still spawn children.
2. **Resumable abort:** before aborting, persist enough controller state that the existing receipt-based resume (`runs.run(key, { resume: { workflowRunId, key, latest: true } })`) can restart the workflow post-reload instead of reporting "Async run not found".
3. **Minimal marker:** if neither is wanted, have the abort path write a terminal status note into the run's artifacts ("aborted during assembly; N/N lanes complete") so the next session's recovery is deterministic instead of archaeological.

## Related

The host-side fix (retaining conversations with active subagent work until runs finish — proposed separately to pi-web-ui's `shouldRetainActive`) removes the most common trigger. This issue is the defense-in-depth layer for hosts and flows that dispose runtimes for legitimate reasons.

---

*Happy to provide the incident transcript and artifact layout if useful.*

redcomet168:
Fixes #1833

## Problem

`cleanup()` unconditionally aborts every live workflow controller when the extension session is replaced or reloaded. That is correct for controllers that could still spawn children, but it also kills workflows whose child runs are **all already terminal** and which are merely in their assembly/post-processing phase — 100% of the expensive work done, assembled result discarded. Recovery was manual (artifacts + git archaeology).

## Fix — drain-and-flush (issue option 1)

New `src/extension/workflow-teardown.ts`:

- Controllers with any **non-terminal** lane (`pending`/`running`/`partial`/`paused`, including detached) abort immediately — unchanged behavior, identical error message.
- Controllers whose lanes are **all terminal** get one bounded **10s grace** (single shared one-shot `unref`'d timer) to finish assembly and flush their result/wake. On expiry, controllers the workflow already removed (its settle `finally` deletes them — the "flushed" signal) are skipped; the rest are force-aborted with the same error.
- Missing/uninspectable job state → conservative immediate abort (old behavior).

The lane projection (`AsyncJobState.steps`) is updated synchronously in the child settle handler (`onTrace → updateTrace → persist`), so the predicate can only fail safe: worst case is an immediate abort, i.e. pre-PR behavior.

**Delivery during grace is covered by existing machinery, verified end-to-end:** a flush landing before the replacing runtime's scan is found via `primeExistingResults` (session-index candidates); one landing after is caught by the new runtime's `fs.watch` on the results dir. `claimPredecessor` preserves ownership across the reload, and the old runtime's epoch/ownership gates make double-delivery impossible.

### True shutdown vs reload

`cleanup()` passes `graceMs: 0` when `shuttingDownParentSession` is set — on a real process shutdown nobody can deliver the flush, so there is no reason to defer exit (the workflow worker thread pins the event loop for the grace window even though the timer itself is `unref`'d). The 10s window applies only to the reload path where a replacing runtime exists.

### Documented trade-off

The all-terminal reading is a snapshot: a sequential workflow between phases (`await run(a); await sleep(); await run(b)`) could launch one more lane through the stale context during the grace window. Damage is bounded by the expiry force-abort (`onAbort` → lanes stopped → `finish` terminates the worker); a launch-admission latch is a possible follow-up if maintainers want that window closed.

### Dropped wholesale map clears

The old `workflowControllers`/`workflowChildStops` `clear()` calls were removed: every workflow's settle `finally` already removes its own entries (including the early-return persist-failure path), the dying runtime's maps have no reachable consumer after cleanup (`currentSessionId` nulled, RPC handlers unsubscribed, delivery epoch-gated), and runIds are UUIDs so replacement collisions can't occur. QA independently attempted to construct a "neither aborted nor self-removed" leak case and could not.

## Tests

- New `test/unit/workflow-teardown.test.ts` (13 tests, fake timers): all-terminal survives grace + skips self-removed entries; live child aborts immediately with the exact message; grace expiry force-aborts; replaced-entry identity guard; mixed sets; uninspectable job; custom grace; `unref` asserted.
- `npm run typecheck` clean; oxlint: zero diagnostics on new files, `index.ts` findings all pre-existing; full `test:unit` 2893 pass / 0 fail (10 `herdr-inspector` cancels reproduce on pristine `main` — environment-dependent, untouched by this diff); scoped adjacent suites 207/207.

🤖 Generated with [pi](https://github.com/nicobailon/pi-subagents) orchestration — happy to split the follow-ups (launch-admission latch, timer clear on early settle) into separate PRs if preferred.

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -89,6 +89,7 @@ import {
 	SUBAGENT_CONTROL_MESSAGE_TYPE,
 	type SubagentControlMessageDetails,
 } from "./control-notices.ts";
+import { teardownWorkflowControllers } from "./workflow-teardown.ts";
 
 export { loadConfig, resolveAsyncByDefault } from "./config.ts";
 
@@ -962,12 +963,13 @@ export default function registerSubagentExtension(pi: ExtensionAPI): void {
 			runtimeCleaned = true;
 			const shuttingDownParentSession = parentSessionEnvValue;
 			// Workflow continuations retain their launch context; abort them before
-			// teardown so a reload cannot launch through a stale context.
-			for (const controller of state.workflowControllers?.values() ?? []) {
-				if (!controller.signal.aborted) controller.abort(new Error("Workflow stopped because the extension session was replaced or reloaded."));
-			}
-			state.workflowControllers?.clear();
-			state.workflowChildStops?.clear();
+			// teardown so a reload cannot launch through a stale context. Workflows
+			// whose child runs are all terminal are only assembling their result, so
+			// they get a bounded, unref'd grace window to flush it (#1833); a workflow
+			// that finishes in time delivers through the replacing runtime's normal
+			// result path, and expiry force-aborts with the same error. Workflows
+			// with any live child run still abort immediately, as before.
+			teardownWorkflowControllers(state);
 			clearRuntimeAgentsForPi(pi);
 			clearTimeout(resultIndexCleanupTimer);
 			clearTimeout(asyncRetentionTimer);
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -969,7 +969,10 @@ export default function registerSubagentExtension(pi: ExtensionAPI): void {
 			// that finishes in time delivers through the replacing runtime's normal
 			// result path, and expiry force-aborts with the same error. Workflows
 			// with any live child run still abort immediately, as before.
-			teardownWorkflowControllers(state);
+			// True parent-session shutdown skips the window (graceMs: 0): the process
+			// is exiting, nobody will deliver the flush, and the workflow worker
+			// thread is not unref'd, so the window would only delay exit.
+			teardownWorkflowControllers(state, shuttingDownParentSession ? { graceMs: 0 } : {});
 			clearRuntimeAgentsForPi(pi);
 			clearTimeout(resultIndexCleanupTimer);
 			clearTimeout(asyncRetentionTimer);
--- a/src/extension/workflow-teardown.ts
+++ b/src/extension/workflow-teardown.ts
@@ -1,0 +1,101 @@
+import type { AsyncJobState, AsyncJobStep, SubagentState } from "../shared/types.ts";
+
+/**
+ * Bounded grace window granted at teardown to workflows whose child runs are
+ * all terminal. Such workflows are usually only assembling their result, so
+ * aborting them outright discards completed child work (#1833). The window
+ * lets the workflow script finish assembly and flush its result and wake-up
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks for the detailed implementation and tests here.

I’m closing this as superseded by #1834, which already fixed and closed #1833 with a narrower current-main design.

I did a fresh exact-head review before closing. The main blocker is the grace-window shape here: it delays aborting an all-terminal workflow controller based on a child-step snapshot. That leaves the old workflow launch context usable during the grace window, so a sequential workflow between phases could still launch more work from stale context. The landed #1834 path instead aborts the controller immediately, aborts the child-facing signal, and only allows bounded pure result assembly once every launched child is settled.

Your diagnosis and reporter credit for #1833 are preserved in the #1834 changelog entry. If the parent-session-shutdown observation needs follow-up, it should be a separate narrow issue/PR against current `main`, not this stale teardown rewrite.

PR #1844, opened 2026-09-03
feat: run child agents as in-process pi sessions

nicobailon:
## Summary

Child agents are now pi sessions created with `createAgentSession` instead of spawned `pi` processes.

- **Foreground children** run inside the parent pi process. All foreground children of one parent share one model runtime, created on the first launch and released on session shutdown. The task is the session prompt, the system prompt is passed directly, structured output and tool diagnostics are captured in memory, and timeouts, interrupts, and stops abort the session. Steering goes through the session instead of a file inbox.
- **Background children** run inside the existing detached runner, which now hosts the child session itself instead of spawning a second pi and parsing its stdout. The runner consumes the parent's control inbox and delivers steer requests with `session.steer`. Every file the parent reads (`status.json`, `events.jsonl`, result files, terminal sidecars) keeps its schema, so the parent side of the background path is unchanged.
- **Child hooks** (prompt runtime, permission gate, tool budget, structured output, watchdog, wait tool, fanout, fast mode) take one typed `ChildRuntimeConfig` object instead of reading about fifty `PI_SUBAGENT_*` environment variables. The environment contract, the argv builder, task delivery through argv or temp files, the stdout protocol readers, the child-side steer inbox, and the startup retry loop are deleted.

## Behavior changes

- Foreground children never load the parent's ambient extensions. pi caches extension modules per process, so a second load would share module state with the parent. Agents that need MCP tools or models from a provider extension must run as background children, which load ambient extensions as before. The foreground preflight error says so.
- MCP servers from the runtime snapshot cannot be provided to an in-process child. Launches that would have needed them fail with a clear error.
- Background children require pi installed as the npm package. The runner imports pi's packages from the installed package directory; a standalone pi binary cannot run them.
- `PI_SUBAGENT_PI_BINARY` now only locates the `pi` binary for Herdr project panes and the profile model probe. `PI_SUBAGENT_TASK_DELIVERY` is removed.
- Detach is unchanged: a detached foreground child keeps running inside the parent and publishes the same receipt and completion notice.

## Measurements

From a spike on pi 0.84.4 with real model calls: per-child launch overhead drops from about 3.5 s of process time and about 240 MB RSS to 5 to 31 ms and about 0.1 MB heap. With two active children, one running a 26 s bash tool and one streaming a long reply, plus a child compaction, the parent's event-loop delay stayed at the 10 ms measurement floor and the parent's own stream did not stall. CPU per foreground launch is down about 60 percent; wall time is unchanged because model calls dominate it.

## Verification

- Typecheck clean.
- Unit and integration suites green (one pre-existing timing flake in the Orca progress test, present on main).
- Real foreground and background runs of the `scout` agent through the installed pi 0.84.4 answered correctly; the background run left no runner process behind.

## Line count

See the diff stat. Source is net about −1,200 lines; most of the total reduction is deleted tests of removed process behavior.

## Review fixes

Six defects found by adversarial review and a parallel reviewer run, fixed in the last six commits (net +29 source lines):

- Parallel children in one runner could observe each other's `MCP_DIRECT_TOOLS` and extension bindings; env application is now serialized through extension load.
- Detached foreground children were aborted on parent session shutdown; they now keep running and hold the shared runtime.
- A `bindExtensions` rejection leaked the created session; it is now disposed.
- The runner drained persistence before disposing child sessions with a 30 s bound; child sessions are disposed first and the drain is unbounded, and a stuck `session_shutdown` handler is bounded to 5 s.
- A workflow `runs.steer` issued before the foreground child's session existed failed immediately; it now polls until `ackTimeoutMs`.
- A shutdown-triggered abort reported a cold-start error; it now reports that the parent session shut down.

## Review fixes, round 2

Seven findings from the second parallel review, fixed in the last three commits:

- Every child session in one process shared each extension's module state, because pi clears its per-process extension factory cache only on a loader's second reload. This was the mechanism behind the env finding: two parallel background children in one runner evaluated an agent extension once, so an ambient adapter kept the first child's `MCP_DIRECT_TOOLS` and session state for both. Each child's loader is now marked as already loaded before its first reload, which makes pi clear the cache; the same pair now evaluates the extension twice. The serialized launch section runs through `bindExtensions`, so `session_start` handlers read their own child's values. Snapshot-and-restore was not adopted: an extension that reads the variables after `session_start` would then always see the runner's baseline instead of only a sibling's value in the overlap window. That residual window is documented; closing it belongs in the extension that does the late read.
- A synchronous `bindExtensions` throw now disposes the session, with a regression test. pi 0.84.4 declares the method `async`, so the path was not reachable, but the code no longer depends on that.
- `ChildSession.dispose()` returns the shutdown promise and both hosts report the run after it resolves, so the runner writes the terminal result after the child's `session_shutdown` work, as it did when the child was a process.
- The `pi-subagents/pi-args` subpath is renamed to `pi-subagents/child-tool-plan`, matching the module it re-exports.
- The stdout-protocol sentence in the packaged skill reference and the two source comments that cited it are rewritten.
- The nested-metadata log line dates from 2026-08-04 on main and is unchanged.
- The trailing blank line in `tool-budget.ts` is removed.

## Review fixes, round 3

- Background child watchdog status was dropped: the child hook wrote it to stdout and the runner never read it. The runner now passes a status sink in the child runtime config and folds the events into the child's event stream, as the foreground host did. The stdout default is removed. Found while validating the wording finding; the two existing background watchdog tests fail without the wiring now that the scripted fake reports through the sink.
- Stale child-process wording in docs, comments, the revive prompt, and test names now describes child sessions. The runner's per-child environment bounds are documented in `docs/agents.md`.
- The real-session e2e harness that installed a fake `pi` binary and parsed child stdout is removed with its `test:e2e` script and CI step; CI only ever skipped it against the package shim.
- Not changed: file-only output validation order is the same on main (empty terminal output is rejected before the file-only path is resolved), so the reported case fails identically there. Restoring the per-launch environment after each child was not adopted because pi-mcp-adapter re-reads `MCP_DIRECT_TOOLS` after `session_start`, and a restored baseline would make that read wrong for every child instead of only for an overlapping sibling.


nicobailon:
No-edit parallel review feedback for current head `28d7e81c147907ec9711b013b9a8d2b195eae64f`.

I would not merge this head yet. The review wave found three high-risk/blocking issues plus docs/API cleanup.

## Blocking / high-risk

### 1. Blocker: synchronous `bindExtensions()` failure still leaks the child session

Location: `src/runs/shared/child-session.ts:189-192`

The current code catches only a rejected promise from `session.bindExtensions(...).catch(...)`. If `bindExtensions()` throws synchronously before returning a promise, the `.catch` handler is never installed and the created session is not disposed.

Evidence: a no-edit inline repro with `bindExtensions() { throw new Error("sync bind failed"); }` printed `caught sync bind failed` and `disposed 0`.

Smallest fix: replace the promise `.catch(...)` with `try/catch` around `await session.bindExtensions(...)`, dispose/shutdown on both sync throws and rejected promises, and add a regression where `bindExtensions` is a non-async function that throws.

### 2. High: background result is committed before in-process child shutdown artifacts are drained

Locations: `src/runs/background/run-child-session.ts:497-554`, `src/runs/background/subagent-runner.ts:4663-4755`, `src/runs/background/subagent-runner.ts:4800-4802`

`runChildSession.finish()` calls `session?.dispose()` and immediately resolves using the pre-dispose messages/session state. `runSubagent` then writes the terminal result/status payload before `childSessions.dispose()` is awaited. Child shutdown work can therefore happen after result assembly.

Evidence: a no-edit inline proof used a child session whose `dispose()` writes a shutdown artifact after 75ms. The result resolved with `existsAtResolve:false`; the artifact appeared later with `existsAfterDelay:true`.

Smallest fix: make child disposal/shutdown awaitable and await it before `runSingleStep` builds `StepResult` / before `runSubagent` commits terminal result payloads. Do not rely only on runner teardown drain, because that happens after result assembly.

### 3. High: per-child extension/MCP env leaks through shared `process.env`

Locations: `src/runs/shared/child-session.ts:118-124`, `src/runs/shared/child-session.ts:164`, `src/runs/shared/child-launch.ts:146-152`, `src/runs/shared/child-launch.ts:292`

Runner-host child launches pass per-child `PI_SUBAGENT_EXTENSION_BINDINGS` and `MCP_DIRECT_TOOLS` through global `process.env`. `createDefaultChildSessionFactory` applies those values before `loader.reload()`, but never restores the previous values. In the in-process runner, live child extensions can observe another child's stale bindings/direct-MCP allowlist after creation.

Evidence: a no-edit inline proof created a child with `PI_SUBAGENT_EXTENSION_BINDINGS` and then printed `process.env.PI_SUBAGENT_EXTENSION_BINDINGS` after `factory.create(...)`; the child value was still present.

Smallest fix: snapshot and restore affected env vars around the full extension loading/binding window, or move this data to child-scoped runtime state. Add a two-child regression proving no cross-child env visibility.

## Medium cleanup

### 4. Legacy `./pi-args` package subpath is still published

Locations: `package.json:19`, `src/api/pi-args.ts:1`, `test/unit/package-manifest.test.ts:115`, `test/unit/package-manifest.test.ts:148`

The PR says child argv/env launch encoding was removed, but `pi-subagents/pi-args` is still exported and tested.

Fix: either remove the export/shim/test expectation for the intended hard cutover, or document this as a deliberate compatibility exception.

### 5. Packaged skill still documents removed stdout protocol behavior

Locations: `skills/pi-subagents/references/constraints-and-recipes.md:24`, `src/shared/utils.ts:416-419`

Stale references remain for `protocolError` and `protocol_output_limit`, even though the current in-process path no longer uses the stdout child protocol.

Fix: update the packaged skill/reference text and source comment to describe child-session events instead.

## Low / mechanical

### 6. Duplicate nested metadata stack traces

Location: `src/runs/foreground/subagent-executor.ts:6780`

Focused tests pass, but the nested metadata fallback prints duplicate `console.error` stack traces for a best-effort metadata failure.

Fix: suppress this best-effort stack or emit one bounded stackless diagnostic.

### 7. `git diff --check` fails

Location: `src/runs/shared/tool-budget.ts:67`

`git diff --check 358110f6d4f7d5981e045cdce5edec1b0ca00743..HEAD` reports `new blank line at EOF`.

## No finding from foreground control lane

The foreground steering/control lane found no issue after focused foreground, steering, stop, interrupt, status, Fleet, and session-transition test coverage.

nicobailon:
Validated each finding against the branch and fixed them in d6c9a7b16, a2c1f1c90, and 800fd06b1 (head 800fd06b1, CI green). Per finding:

**1. Synchronous `bindExtensions` throw.** Legit as a contract gap; not reachable with pi 0.84.4, which declares `bindExtensions` as `async`. Replaced the `.catch` with `try/catch` around the `await`, and added a regression with a non-async `bindExtensions` that throws.

**2. Result committed before child shutdown work.** Legit. `ChildSession.dispose()` now returns the shutdown promise, and both `runChildSession.settle()` and the foreground `finish()` resolve only after it, so the runner writes the terminal result after the child's `session_shutdown` handlers ran, as it did when the child was a process. Regression: a foreground run resolves only after a delayed dispose finished.

**3. Per-child env through `process.env`.** Legit, and the mechanism is wider than the finding describes. pi's `DefaultResourceLoader.reload()` uses a per-process extension factory cache that it clears only on a loader's second reload, so two loaders in one process for the same extension file evaluated its module once. Every child in one runner therefore shared each ambient extension's module state, including pi-mcp-adapter's load-time `MCP_DIRECT_TOOLS` capture and its session state. Fix: each child's loader is marked as already loaded before its first `reload()`, which makes pi clear the cache, and the serialized launch section now runs from env application through `bindExtensions`, so `session_start` handlers read their own child's values. Verified inside a real pi host (a scratch extension evaluated once per plain reload, again after the flagged reload) and end to end: two parallel background children in one runner process evaluated an agent extension once on the previous head and twice now.

Snapshot-and-restore was not adopted. pi-mcp-adapter re-reads `MCP_DIRECT_TOOLS` inside its initialization after `session_start` returns, and a restored baseline would make that read wrong for every child instead of only for a sibling in the overlap window. That residual window is documented in the changelog; closing it belongs in the adapter (use the load-time capture for the late read).

A deep import of `dist/core/extensions/loader.js` for `clearExtensionCache` was tried first and rejected: inside a real pi host it resolves to a second copy of the loader module, so it does not touch the host's cache.

**4. `./pi-args` subpath.** Legit. Renamed to `pi-subagents/child-tool-plan` with the file moved to `src/api/child-tool-plan.ts`; manifest test and changelog updated. No consumer of the old subpath was found in local extensions or repos.

**5. Stale stdout-protocol text.** Legit. The sentence is removed from the skill reference, and the two source comments that cited `MAX_CHILD_PENDING_LINE_BYTES` now describe the transcript and `events.jsonl` cost instead.

**6. Duplicate nested metadata stack traces.** Pre-existing on main: the line dates from commit 9d74cba959 (2026-08-04) and this PR does not touch it. Left as is.

**7. `git diff --check`.** Fixed.

Source delta for the round: +87 / −66 across 6 files; the rest is tests, docs, and the rename.

nicobailon:
Follow-up no-edit fix review for current head `800fd06b100489e92a6089100f93a5b0f1392e64`.

Several prior items are fixed: sync `bindExtensions()` throw/rejection disposal passes, `./pi-args` was removed/renamed to `child-tool-plan`, `git diff --check` is clean, and the duplicate nested metadata stack traces are confirmed pre-existing/out-of-scope.

I would still hold merge for the remaining issues below.

## Blocking / high-risk

### 1. Blocker: file-only output written during awaited child disposal is still reported as empty

Locations:
- `src/runs/foreground/execution.ts:782-783`, `1419-1432`, `1550-1569`
- `src/runs/background/subagent-runner.ts:1159-1188`, `1232-1238`, `1273-1281`
- `src/runs/background/run-child-session.ts:510-543`

The new shutdown await is real, but result assembly still validates empty terminal output before resolving the configured file-only output path.

Foreground proof: a no-edit inline fake `ChildSession.dispose()` waited briefly, wrote the configured file-only output path, and returned before `runSync` resolved. The output file existed and contained the expected content, but the result still returned `exitCode:1`, `Subagent produced no output...`, no `savedOutputPath`, and empty `finalOutput`.

Background proof: a no-edit inline async runner proof wrote the configured output path during delayed dispose. The terminal result/status was still failed as empty: `success:false`, child error `Subagent produced no output...`, `childOutput:''`, `savedOutputPath:undefined`, while `outputFileExists:true` and the file contained the expected shutdown artifact.

Smallest fix: after awaited child disposal, resolve/check `outputMode: "file-only"` output paths before empty-output validation, or suppress empty-output failure when the required file-only output exists/changed after the run. Then populate `savedOutputPath` / `outputReference` / `outputState` from that file.

### 2. High: runner child per-launch env still leaks through shared `process.env`

Locations:
- `src/runs/shared/child-session.ts:133`, `179`, `218`
- `src/runs/shared/child-launch.ts:145-153`, `292`

The loader/cache fix gives separate extension module instances during reload/bind, but `process.env` itself is still not restored after each child launch. Any loaded child extension that reads `process.env.PI_SUBAGENT_EXTENSION_BINDINGS` or `process.env.MCP_DIRECT_TOOLS` later can observe a later child’s values.

Proof: a no-edit inline probe launched two sessions with distinct env values, then invoked the first session’s prompt after the second launch. The first child’s late read observed the second child’s value: `{"firstPromptLateRead":"second-child","activeProcessEnv":"second-child"}`. A second probe confirmed reload/bind saw isolated values, but `process.env` remained set to each launched child value after `create()` returned.

Smallest fix: treat `launch.processEnv` as a scoped launch override. Snapshot the prior values for those keys, apply them through the serialized reload/create/bind section, and restore them in `finally` after success or failure. If MCP/extension adapters require late reads, move/capture that state per session instead of relying on global env. If any residual late-read behavior remains intentional, document the exact bounds in repo docs.

## Minor cleanup still left

### 3. Stale child-process/stdout wording remains

`protocolError` / `protocol_output_limit` are gone, but stale child-process/stdout phrasing remains in docs/comments/tests. Examples:

- `docs/agents.md:3`
- `docs/observability.md:63`
- `skills/pi-subagents/references/execution-controls.md:239-240`
- `src/extension/index.ts:910-914`
- `src/watchdog/register-child.ts:49`, `53-55`
- `src/runs/shared/abort-recovery.ts:104-106`
- `src/runs/background/async-resume.ts:626`
- `test/e2e/real-session-subagent.test.ts:4-8`
- `test/unit/streamed-progress-bounds.test.ts:67`

Smallest fix: replace native-child wording with child session / child host process / session event stream wording. For `fake-child-session`, clarify that raw stdout is only fixture compatibility, not production transport. Update the revive prompt regex and matching tests together if changing model-visible revive text.

nicobailon:
Validated the three follow-up findings against the branch. Two commits address them (48532f452, e92b462c0; head e92b462c0, CI green); one finding is not changed and one is not adopted, with reasons below.

**1. File-only output written during child disposal reported as empty.** Not a regression, so not changed. Both hosts validate empty terminal output before resolving the file-only output path, and main does the same (`execution.ts` on main: `missingOutput` at line 1560, the file-only branch at 1684). The proof wrote the file with no assistant text, which main rejects with the same "produced no output" error. Disposal timing does not enter: the check runs after `runSync` resolves, which is now after disposal. If file-only mode should accept a child that writes the file and says nothing, that is a separate change to the validation order on main, not part of this transport change.

**2. Per-launch env not restored.** Not adopted; bounds documented. A late reader wants its own child's values, and neither leaving nor restoring gives it that. Leaving the values is right except when a sibling launched since; restoring makes every late read see the runner baseline, which for pi-mcp-adapter's post-`session_start` read means `undefined`, and `undefined` there is "no override" rather than the child's `__none__`. So restoring is wrong for every child instead of only for an overlapping sibling. The bounds are now stated in `docs/agents.md` next to the ambient extension rules: values are held from application through the child's `session_start`, one launch at a time; reads at load or during `session_start` see the child's own values; later reads can see a sibling launched since. Closing that belongs in the adapter, which already captures the value at load and can use that capture for the late read.

**3. Stale child-process wording.** Fixed in e92b462c0: docs, comments, the revive prompt text (with the completion-guard regex and its three tests), the streamed-progress test name, and the fake's header. `docs/observability.md:63` had no such wording; the other cited locations did.

**Found while validating 3, fixed in 48532f452.** Background child watchdog status was dropped on the branch. `registerChildWatchdog` defaulted to writing status events to stdout, the runner never passed a sink, and nothing read the runner's stdout, so `status.json` and Fleet lost watchdog phases for background runs. The runner now passes a status sink in the child runtime config and folds the events into the child's event stream, as the foreground host already did. The stdout default is removed, and a configured watchdog without a sink throws at child load. The scripted fake now reports watchdog status through the sink the way a real child does; with that, the two existing background watchdog tests fail without the runner wiring and pass with it.

**Also removed in e92b462c0.** The real-session e2e harness (`test/e2e`, `test/support/real-session-runner.ts`, `test/support/real-session-child-cli.mjs`, the `test:e2e` script and CI step) installed a fake `pi` binary and parsed child stdout. Neither path exists now, and CI only ever skipped it against the package shim. Removed rather than left as dead code; a real-session e2e for the in-process route would be a new harness.

PR #1873, opened 2026-09-04
fix: alias pi-agent-core node export

### LINKED ISSUE #1871: Async runners on Pi 0.85 mangle pi-agent-core/node via JITI_ALIAS
## Summary

Detached/background runs fail before model execution with Pi 0.85.0 because the runner's `JITI_ALIAS` map aliases the package root but not the new `@earendil-works/pi-agent-core/node` export.

Jiti treats the root alias as a prefix and resolves the subpath as `dist/index.js/node` instead of the package export target `dist/node.js`:

```text
Cannot find module '.../@earendil-works/pi-agent-core/dist/index.js/node'
Require stack:
- .../@earendil-works/pi-coding-agent/dist/experimental/server.js
```

## Environment

- pi-subagents 0.65.0
- @earendil-works/pi-coding-agent 0.85.0
- @earendil-works/pi-agent-core 0.85.0
- jiti 2.7.0
- Node 24.19.0
- macOS

## Reproduction

Any detached child reaches the failure before model execution:

```text
subagent { agent: "read-only-reviewer", task: "Review the current diff" }
```

The Jiti behavior can also be reproduced directly:

```sh
JITI_ALIAS='{"@earendil-works/pi-agent-core":".../pi-agent-core/dist/index.js"}' \
  node jiti/lib/jiti-cli.mjs \
  pi-coding-agent/dist/experimental/server.js
```

## Root cause

`HOST_PEER_ALIASES` in `src/runs/background/runner-aliases.ts` contains:

```ts
{ specifier: "@earendil-works/pi-agent-core", pkg: "@earendil-works/pi-agent-core", subpath: "." }
```

but no exact alias for the `./node` export introduced into Pi's runtime graph. The existing runtime-import test only scans pi-subagents' own source graph, so it does not see runtime imports from the aliased Pi package.

## Proposed fix

Add the exact subpath alias:

```ts
{ specifier: "@earendil-works/pi-agent-core/node", pkg: "@earendil-works/pi-agent-core", subpath: "./node" }
```

and cover it with a focused alias-resolution assertion.

## Separate Pi 0.85 blocker

After adding the exact alias locally, loading advances to Pi 0.85's separately reported missing `@earendil-works/pi-server` dependency: earendil-works/pi#9132. That upstream packaging issue must also be fixed (or worked around by installing `pi-server`) before 0.85 detached runs can complete. This issue is specifically for the incorrect Jiti subpath rewrite owned by pi-subagents.

Related: #599, #334, #526.

nicobailon:
Fixes #1871.

Adds an exact detached-runner JITI alias for @earendil-works/pi-agent-core/node, so Jiti does not resolve the subpath as dist/index.js/node when Pi 0.85 imports the package node export.

Includes a focused alias-resolution regression test and changelog credit for @plopezlpz.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 - Lazy-load prompt audit optional Pi peers. Thanks [@VladimirGVP](https://github.com/VladimirGVP) for #1860.
 - Key subagent children launched under an explicit `sessionDir` by their run id so concurrent children resolve distinct session files instead of sharing `run-0`. Thanks to [@dat9uy](https://github.com/dat9uy) for #1859 and #1858.
+- Alias `@earendil-works/pi-agent-core/node` to Pi's exact `./node` export for detached runners. Thanks [@plopezlpz](https://github.com/plopezlpz) for #1871.
 
 ## [0.65.0] - 2026-09-04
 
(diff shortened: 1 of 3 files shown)


nicobailon:
Closing this duplicate maintainer PR in favor of contributor PR #1872 for the same #1871 fix. The pushed #1872 head contains the focused regression test and changelog credit for @plopezlpz.

PR #1875, opened 2026-09-04
fix: bound foreground workflow run id to prevent ENAMETOOLONG

jstillwa:
### Summary
In long conversations or when model providers emit serialized session tool-call IDs that exceed standard lengths (e.g. >255 characters), foreground workflow launches fail with `ENAMETOOLONG` when attempting to stat or inspect run status paths under `DIRS.async`.

While PR #1131 bounded tool-call IDs inside `.active-runs` and `result-index`, in `subagent-executor.ts` during foreground workflows (`requestParams.async === false`), `_id` was still passed directly into `workflowRunId` and `parentWorkflowRunId` without encoding. This resulted in paths like `/tmp/pi-subagents-uid-1000/async-subagent-runs/<raw_long_id>/status.json`, which exceeds Linux `NAME_MAX` (255 bytes).

### Fix
1. **Encode foreground workflow run IDs:** Use `encodeIndexSegment(_id)` in `subagent-executor.ts` to ensure foreground workflow run IDs and child path segments are deterministically capped using SHA-256 (`~sha256-<hash>`) when exceeding filesystem bounds.
2. **Defensive `readStatus` guard:**
   - In `src/shared/utils.ts`, treat `ENAMETOOLONG` as non-existent in `isNotFoundError`.
   - Bail out early returning `null` if any directory basename exceeds 255 bytes before calling `fs.statSync`.

### Validation
- Added unit test in `test/unit/index-segment.test.ts` verifying `readStatus` safely handles oversized directory paths without throwing `ENAMETOOLONG`.
- Ran unit tests across `index-segment.test.ts` (6/6 passed).

### DIFF
--- a/src/runs/foreground/subagent-executor.ts
+++ b/src/runs/foreground/subagent-executor.ts
@@ -49,6 +49,7 @@ import { updateActiveRunIndex } from "../background/active-run-index.ts";
 import { steeringReceipt } from "../background/steering.ts";
 import { acquireActiveAsyncCapacity, ActiveAsyncCapacityError, getActiveAsyncCapacitySnapshot, resolveAbandonedSlotReleaseAfterMs, resolveMaxActiveAsyncRunsPerSession, transferActiveAsyncCapacity, type ActiveAsyncCapacityHandle } from "../background/active-async-capacity.ts";
 import { isScheduledRunAction } from "../background/scheduled-runs.ts";
+import { encodeIndexSegment } from "../background/index-segment.ts";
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for the fix. I landed this through replacement PR #1877 so the branch could stay narrow on current `main` and include the review fixes Greptile found.

The merged replacement preserves your authored contribution and credits [@jstillwa](https://github.com/jstillwa) for #1875 in the Unreleased changelog.

PR #1880, opened 2026-09-04
fix: alias the host subpath imports the detached runner evaluates

pwguler:
## Problem

Follow-up to #1874. After the arbiter's eager peer imports were lazy-loaded, background runner crashes moved one layer down — into the **host's own dist**, evaluated by the runner's jiti:

```
Error: Cannot find module '.../pi-agent-core/dist/index.js/node'
  requireStack: [.../pi-coding-agent/dist/experimental/server.js]
    at server.js:6  →  import { NodeExecutionEnv } from "@earendil-works/pi-agent-core/node"
```

**Root cause:** `HOST_PEER_ALIASES` maps `@earendil-works/pi-agent-core` to its resolved *file* (`dist/index.js`). When the host's `server.js` imports the package **subpath** `./node`, jiti's alias resolution joins the subpath onto the aliased file path instead of re-resolving through the package `exports` map — producing the mangled `dist/index.js/node` request. The same applies to every other subpath the host's runtime graph imports.

## Fix

Enumerate the host subpath imports the runner actually evaluates, exactly the way `pi-ai/compat` and `typebox/compile` were already enumerated:

- `@earendil-works/pi-agent-core/node` (server.js:6 — the crash site)
- `@earendil-works/chord`, `@earendil-works/chord/context` (client-runtime.js:2)
- `@earendil-works/pi-server`, `@earendil-works/pi-server/unix` (server.js)
- `@earendil-works/pi-client/unix` (client-runtime.js:4)

Extend the test shim with minimal `chord`/`pi-server`/`pi-client` peer fixtures so `host-peer-runtime-imports.test.ts` covers the new rows, and add a dedicated regression test asserting `pi-agent-core/node` resolves to its exact export rather than the root alias plus a suffix.

## Verification

- `npx tsc --noEmit` clean.
- `npm test`: 2801 pass / 15 fail — the same 15 agent-management failures as clean main in this environment (environmental ENOENTs); no new failures. The new shim-backed assertions pass.
- Real install (pi-coding-agent@0.85.0, global npm): before this change every background dispatch died at `server.js:6` with the stack above; after it, the runner boots through full TS evaluation and background subagents dispatch successfully end to end.


nicobailon:
Thanks again for this fix, @pwguler.

I landed it through replacement PR #1883 so I could preserve your authorship while adding the required changelog credit and keeping the test fixtures minimal after review.

Landed commit: 4813c318e60c46f4afd3e8fbefd50db61dca1253
Merge commit on `main`: cb83adbc361c8f815c8d766f8766faf3d35ff9c0

Issue #1916, opened 2026-09-05
test: await background terminal evidence before acceptance-report teardown


nicobailon:
## Failure and current contract

In main Tests [33948394910, Windows job 101258498172](https://github.com/nicobailon/pi-subagents/actions/runs/33948394910/job/101258498172) at `34997e2f9b7149cfb5c6897d3093707d4f942196`, all 12 acceptance-file-report cases passed, then the owning test process failed with `ENOTEMPTY` from `test/support/isolated-temp-root.mjs:19`.

`test/integration/acceptance-file-report.test.ts` waits only for result-file existence (lines 113–120). Its `afterEach` immediately deletes each case directory, and `after` uninstalls the mock queue. Result publication does not prove background runner termination, so these teardown steps lack the required ordering guarantee.

This is distinct from #1906's missing terminal event. The missing guarantee is source-proved; the specific writer or Windows filesystem condition behind this ENOTEMPTY is not established. Local macOS controls passed and do not disprove the race.

## Scope

Await the owned background runners' existing terminal evidence before removing their test paths or mock queue, including assertion-failure paths. Preserve all acceptance/report assertions and fail if terminal evidence is unavailable. Keep this test-local; no production lifecycle changes, swallowed cleanup errors, extra timeout allowances, global retries or general cleanup framework.

Validate the existing file, relevant typecheck and Windows CI. Do not claim this resolves the other async timeouts.

nicobailon:
The requested teardown ordering is implemented in #1918 and validated on integrated main by runs 33979103482 and 33980355523, including Windows. afterEach awaits the owned-runner terminal barrier before deleting case paths; a failed barrier is retained so later hooks cannot retry cleanup/reset, and final teardown requires no unproven owners before uninstalling the mock queue. Existing acceptance assertions remain. Closing only this test-local ordering gap—not claiming a general Windows timeout fix. #1906 and the separately observed result-publication waits remain distinct.

Issue #1906, opened 2026-09-05
test: investigate recurring Windows timeout awaiting shared-cwd runner terminal event


nicobailon:
## Observed failure
Windows CI intermittently times out in the existing regression `does not use shared-cwd sibling tracked edits as parallel completion-guard proof` (`test/integration/async-execution.part-3.test.ts`). Its cleanup waits for `subagent.run.process_terminal` through `waitForAsyncEvent`; the helper's default deadline is 10 seconds.

Occurrences:
- Baseline main Tests run [33927944448](https://github.com/nicobailon/pi-subagents/actions/runs/33927944448), first attempt at `83be9c3de2cde1553c0269f383efc1eb1194dc8b`. The focused test passed locally; one failed-job rerun passed.
- PR #1905 Tests run [33942949176, Windows job 101243732019](https://github.com/nicobailon/pi-subagents/actions/runs/33942949176/job/101243732019) at `a8b7d5a599d51302bd3a19665f2c5f67f6e9f862`. Same missing terminal event, reported from the test's finally block. The PR does not change this test or the terminal-event observer. Ubuntu, Bun and clean-install checks passed.

## Required investigation
Trace result publication, runner shutdown and terminal-event observation on Windows. Establish whether the event is lost, delayed, or legitimately unavailable before changing the assertion. Preserve the cleanup safety predicate. Do not fix this with blind retries, sleeps, a weaker assertion or an arbitrary timeout increase.

The second occurrence makes this a recurring validation blocker rather than a resolved one-off flake. No root cause is claimed yet. The available CI artifacts contain smoke logs, not this runner's lifecycle artifacts.

nicobailon:
Diagnostic-only PR #1908 landed as 506ea96891f51c1541b7bf6f93d645d3b9454a91. Exact-head CI (33945187780) and post-merge main CI (33945577863) passed, including Windows. The original assertions, 10-second wait and cleanup predicate remain unchanged.

Keeping this issue open: passing runs do not establish the root cause or a fix. A future recurrence now emits bounded runner PID/spawn/exit/close observations and allowlisted terminal-sidecar/event evidence. The next action is to compare those diagnostics on the failing run before choosing an event-publication, observation-ordering or shutdown fix. No speculative production change or timeout increase is planned.

nicobailon:
[...]

PR #1918, opened 2026-09-05
test: wait for acceptance runners before fixture teardown

nicobailon:
Addresses #1916; main fix-forward candidate for the acceptance-test cleanup ordering failure.

Keeps teardown behind the existing cached observed-only runner barrier, including assertion failures. Adds bounded test-local disposal/exit markers and failure-only correlated status/candidate/proof/result/journal evidence plus a cached PID-presence observation. Markers and PID presence never substitute for terminal proof. The original 10-second deadline, 50ms cadence and acceptance assertions remain unchanged. No production/global-root cleanup suppression, retries or deadline extension. Failed proof retains the case cwd/mock; it does not guarantee retention of the separate runtime root.

Refreshed onto main379a. One 12/12 local acceptance run and typecheck passed; retained challenge and separate simplification found no changes. The initial simplification report-persistence failure was recovered through the same retained native child without changing source or repeating validation. Fresh exact-increment source review found no issues. Parent verified full rebase composition and the complete authorized delta. Original reviews keep their original SHA attribution.

Windows root cause is NOT established or claimed fixed. Prior main379a ENOTEMPTY and historical5cd terminal-tail failures remain evidence. New exact-head Ubuntu/Windows/Bun/pi085 and Greptile/full feedback are required, followed by required main CI; no unchanged failed-run retry is requested.

### DIFF
--- a/test/integration/acceptance-file-report.test.ts
+++ b/test/integration/acceptance-file-report.test.ts
@@ -9,8 +9,9 @@ import { after, afterEach, before, beforeEach, describe, it } from "node:test";
 import assert from "node:assert/strict";
 import * as fs from "node:fs";
 import * as path from "node:path";
+import { readProcessTerminal } from "../../src/runs/background/process-terminal.ts";
 import type { MockPi } from "../support/helpers.ts";
-import { createEventBus, createMockPi, createTempDir, events, makeAgent, makeAgentConfigs, makeMinimalCtx, removeTempDir, tryImport } from "../support/helpers.ts";
+import { createEventBus, createMockPi, createTempDir, events, makeAgent, makeAgentConfigs, makeMinimalCtx, tryImport } from "../support/helpers.ts";
 
 interface AcceptanceSummary {
 	status?: string;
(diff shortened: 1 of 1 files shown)


nicobailon:
The Windows gate still fails at this head: [run 33956197822](https://github.com/nicobailon/pi-subagents/actions/runs/33956197822).

The new diagnostics narrow the failure: the first background run finishes its logical work and publishes the final zero-writer candidate, but neither `exit` nor `close` is observed before the unchanged terminal deadline. The matching proof remains pending. The barrier correctly retains the fixture and mock queue; later hook failures are cascades.

This rules out failure to capture the process and an unfinished pre-candidate persistence drain at the snapshot. It does **not** establish OS liveness, actual exit, or a production root cause. Keeping this PR blocked rather than weakening cleanup, extending deadlines, or retrying the same head.

The next discriminating evidence would be disposal-entry/return and exit-phase markers for this acceptance launch, with a failure-only PID-presence observation. The earlier #1917 markers belong to another test and were not generated for this run. #1906 and the separate capability-audit correction #1925 retain their own evidence and scope.

PR #1919, opened 2026-09-05
fix: clear recovered assistant errors on tool-use completion

Jonathanm10:
## Summary

Clear an earlier `assistantError` when a native child produces an error-free `toolUse` response containing an actual tool call. Apply the same condition in the background and foreground drivers.

A recovered request can finish through terminating `structured_output` without another assistant text response. Currently, both drivers keep the earlier provider error latched in that case, returning a false failure; the background runner consequently skips reading the captured structured output.

```text
assistant error → retry → successful structured_output tool call → capture + terminate → settled
```

This is distinct from #894, which handles recovered tool errors during structured-output validation. This change fixes the assistant-error latch before that validation runs.

## Scope and safety

- Eight runtime lines, plus regressions using the existing scripted child-session integration fixture.
- Do not clear errors on retry-success notifications alone, missing tool calls, non-`toolUse` stops, or error-bearing responses.
- Existing structured-output/acceptance and timeout/cancellation/finalization checks remain unchanged.
- Does not diagnose or change the underlying transport failures.

## Validation

All runtime validation used a network-denied macOS sandbox with isolated HOME, agent, temporary, and model-exclusion storage. No live model/provider or production operations.

- New regression file: baseline **6 passed, 2 failed**; patched **8 passed, 0 failed, 0 skipped**. Both failure cases are recovered structured termination, one per driver.
- Existing affected integration selection: **8 passed**, covering ordinary-text recovery, empty-response failures, structured capture, and missing structured/acceptance evidence.
- Oxlint on the new test: passed (0 warnings/errors).
- `git diff --check`: passed.
- Independent native-agent source review: no blockers (OK with notes). Focused regressions, affected integration selection, typecheck, and new-test lint were rerun successfully after rebasing onto `90ff9369`; all three reviewed files remained byte-identical.

Commands:

```sh
node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/recovered-assistant-error.test.ts
npm run typecheck
./node_modules/.bin/oxlint test/integration/recovered-assistant-error.test.ts
```

Broader unit-suite attempt: **2,827 passed, 5 failed, 3 skipped**. Failures involved process-group/host-command checks (including `spawnSync ps EPERM` under the sandbox) and a worktree setup timeout. The worktree case passed when rerun alone. These failures were not baseline-compared, so the full unit suite is **not claimed green**. The full integration suite was not run.

### DIFF
--- a/src/runs/background/run-child-session.ts
+++ b/src/runs/background/run-child-session.ts
@@ -474,6 +474,10 @@ export function runChildSession(input: RunChildSessionInput): Promise<RunChildSe
 					}
 				}
 				if (event.message.errorMessage) assistantError = event.message.errorMessage;
+				else if (hasToolCall && event.message.stopReason === "toolUse") {
+					// A recovered request can finish via a terminating tool, without a text stop.
+					assistantError = undefined;
+				}
 				const eventUsage = event.message.usage;
 				if (eventUsage) {
 					usage.turns++;
--- a/src/runs/foreground/execution.ts
+++ b/src/runs/foreground/execution.ts
@@ -1105,6 +1105,10 @@ async function runSingleAttempt(
 						}
 					}
 					if (evt.message.errorMessage) assistantError = evt.message.errorMessage;
+					else if (hasToolCall && evt.message.stopReason === "toolUse") {
+						// A recovered request can finish via a terminating tool, without a text stop.
+						assistantError = undefined;
+					}
 					const assistantText = extractTextFromContent(evt.message.content);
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks @Jonathanm10! Your recovered-assistant-error fix and regression tests landed through #1927 in 063a5ad78a8f92ba46ed5c9cb37f50b07cc76514. Your original author, date and commit message were preserved in the implementation commit, with a separate changelog credit linking your profile and this PR. Closing this original as superseded by the merged replacement.

PR #1924, opened 2026-09-05
fix: keep fleet inspector active-run roster stable

### LINKED ISSUE #1923: Fleet inspector: active run rows jump around while monitoring - sort by start time
**Environment**: pi-subagents 0.65.0 (2026-09-04), @earendil-works/pi-coding-agent 0.84.4

**Phenomenon**: With two or more concurrent active subagents (e.g. parallel workers in a workflow), the fleet inspector roster reorders the *running* rows on almost every refresh. Example: two running `chorus-worker` runs (`4049c471`, `61e92e14`) plus one completed reviewer show as `4049c471, 61e92e14, reviewer` in one frame and `61e92e14, 4049c471, reviewer` in the next, purely as the workers emit heartbeats. Completed entries stay put; only running ones shuffle. Selection is preserved by key, so the selected row keeps its item but jumps between positions.

**Root cause**: `collectFleetSnapshot` / `orderFleetAsyncRuns` (`src/tui/fleet.ts`) sorts active runs by `lastUpdate` descending (`lastUpdate ?? endedAt ?? startedAt`), and `lastUpdate` is refreshed on essentially every activity heartbeat (`runs/background/async-execution.ts`, `runs/foreground/subagent-executor.ts`). Relative order of two active runs therefore flips whenever the "other" one emits anything. The same pattern also affects foreground controls (sorted by `updatedAt`) in the same function.

Note the compact fleet status widget (`src/tui/fleet-status.ts`, `SubagentFleetStatus`) already sorts entries by `startedAt` ascending — so the inspector is also inconsistent with the status widget.

**Suggestion**: make the active group order stable.

1. (Recommended) Sort active runs by `startedAt` (ascending, matching `fleet-status.ts`; or descending for newest-on-top) — minimal change; keep the active/terminal grouping and terminal group sorted by recency as today.
2. If "most recently active on top" is considered valuable, keep the stable order but surface activity via a pulse/activity marker or a last-activity column instead of moving rows.
3. If any activity-based ordering is kept for active runs, limit re-bubbling to state transitions (e.g. queued→running), not heartbeats; apply the same stable secondary key consistently to tracked jobs and foreground controls.

**Acceptance criteria**:

- With ≥2 concurrent active runs emitting at different rates, roster row order stays stable across refreshes.
- Active group still precedes terminal group; terminal group stays limited/ordered by recency; key-based selection preserved.
- A unit test covers ordering stability of `orderFleetAsyncRuns` / `collectFleetSnapshot`.

Related: #802 addressed terminal failures crowding the active context; this is the active-group counterpart (row reordering).

expoli:
## Problem

The fleet inspector roster reorders the *running* rows on almost every refresh when two or more active subagents run concurrently (e.g. parallel workers). Selection is preserved by key, but the selected row itself jumps between positions, making the roster hard to monitor. See #1923.

## Root cause

`orderFleetAsyncRuns` (`src/tui/fleet.ts`) sorted active runs by `lastUpdate` descending, and `lastUpdate` is refreshed on essentially every activity heartbeat. The relative order of two active runs therefore flipped whenever the "other" one emitted output. `collectFleetSnapshot` applied the same unstable rule to foreground controls (`updatedAt`).

## Fix

- Sort active (queued/running) async runs by `startedAt` ascending with an `id` tiebreak — the same stable, creation-order rule the compact fleet status widget already uses (`src/tui/fleet-status.ts`).
- Apply the same `startedAt` + `runId` rule to the foreground-control ordering in `collectFleetSnapshot`.
- Terminal runs keep recency ordering (`lastUpdate` descending, bounded window) so recently finished work stays on top; `recentForeground` history is unchanged.
- Remove the now-redundant pre-sort of tracked active jobs.

## Verification

- New unit test: active roster order is stable when workers emit at different rates, and later-started runs land after earlier ones (exercises the `listAsyncRuns` history path).
- Adjusted one existing foreground-order assertion to the new start-time rule.
- `npm run typecheck` and `npm run test:unit` pass (2833 passed, 0 failed).

Fixes #1923

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 ### Fixed
+- Keep the fleet inspector roster stable by ordering active runs by start time instead of last activity heartbeat (#1923). Thanks to [@expoli](https://github.com/expoli) for #1924.
 - Rescue confident read-only background completions misclassified as implementation before publishing missing-edit failures (#1911). Thanks to [@yanqianglu](https://github.com/yanqianglu).
 - Keep background streaming status updates batched during long-running and attention states while publishing child activity transitions immediately (#1901).
 - Honor distinct output paths on retained workflow follow-ups, preserving the original report and returning the saved output references (#1903).
--- a/src/tui/fleet.ts
+++ b/src/tui/fleet.ts
@@ -182,7 +182,8 @@ function asyncItems(run: AsyncRunSummary, description?: string): FleetItem[] {
 function orderFleetAsyncRuns(runs: AsyncRunSummary[], terminalLimit: number): AsyncRunSummary[] {
 	const updatedAt = (run: AsyncRunSummary) => run.lastUpdate ?? run.endedAt ?? run.startedAt;
 	const byNewest = (left: AsyncRunSummary, right: AsyncRunSummary) => updatedAt(right) - updatedAt(left);
-	const active = runs.filter((run) => run.state === "queued" || run.state === "running").sort(byNewest);
+	const byStartedAt = (left: AsyncRunSummary, right: AsyncRunSummary) => left.startedAt - right.startedAt || left.id.localeCompare(right.id);
+	const active = runs.filter((run) => run.state === "queued" || run.state === "running").sort(byStartedAt);
 	const terminal = runs.filter((run) => run.state !== "queued" && run.state !== "running").sort(byNewest);
 	return [...active, ...terminal.slice(0, terminalLimit)];
 }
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks, @expoli. I preserved your implementation, authorship and regression coverage in #1930, including the credit update. It adds one small correction for tracked jobs without a start timestamp, with a regression that fails on the original implementation. Your PR will stay open until the replacement lands; main CI is being checked after the separate #1929 test correction.

nicobailon:
Superseded by maintainer replacement PR #1930, which has landed with the original contribution preserved and credited. Thanks again for the fix.

PR #1937, opened 2026-09-05
fix(workflows): validate literal baseRef values in preflight

### LINKED ISSUE #1934: baseRef: static validation accepts unsupported full commit IDs; clarify errors and guidance
## Summary

In `pi-subagents` 0.65.0, `action: "validate"` accepts a workflow with a literal full commit ID in a child’s `baseRef`, but executing that workflow fails before child dispatch:

```text
runs.run baseRef must be a valid Git ref.
```

The restriction is intentional: #1850 explicitly says it rejects revision expressions and raw object IDs. This issue is about making that existing contract visible in static validation, tool guidance, and errors—not changing the policy to allow full commit IDs.

Related: #1842 and #1850.

## Reproduction

Validate this workflow (the literal below is a synthetic 40-character hex ID; the lexical result does not depend on whether it exists in a repository):

```js
subagent({
  action: "validate",
  workflowScript: "return runs.run('ref-probe', {agent:'scout', task:'Do not modify files.', worktree:true, baseRef:'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'});"
});
```

Static validation returns:

```json
{"ok":true,"errors":[]}
```

Executing a workflow with the same literal `baseRef` reaches `validateRunCall`, which rejects full 40- or 64-character hexadecimal IDs before dispatching the child or resolving the value with Git.

The observed session failure used a real full SHA equal to the checkout’s `HEAD`, not the synthetic value above. Git resolved that SHA successfully. The failed workflow receipt contained zero children. Omitting `baseRef` let the retry use the default `HEAD`; its worktree handoff recorded the same intended commit.

## Evidence

In the installed 0.65.0 source, the host workflow, embedded workflow worker, and managed-worktree validators all contain:

```ts
if (/^(?:[0-9a-f]{40}|[0-9a-f]{64})$/iu.test(ref)) return false;
```

Locations:

- `src/workflows/scripted-workflow.ts:15` — host predicate.
- `src/workflows/scripted-workflow.ts:62` — embedded worker predicate.
- `src/workflows/scripted-workflow.ts:647` — runtime `baseRef` rejection.
- `src/runs/shared/worktree.ts:352` — worktree predicate.
- `src/workflows/scripted-workflow.ts:1546` — `validateWorkflowScript` checks syntax and selected AST properties, but not a literal child `baseRef`.

The public guidance describes a “safe Git ref” that resolves to a commit, and the runtime error says only “valid Git ref.” Neither explains the full-object-ID exclusion. A model trying to pin an immutable baseline can therefore reasonably choose a full SHA and get an avoidable launch failure. Running static validation first does not catch it.

## Expected behavior

1. **Static diagnostics:** Apply the runtime lexical policy to directly inspectable literal `baseRef` values in child launch parameters, including `runs.run` and `runs.all`. A known-invalid literal should fail validation without starting a workflow. Computed values still require runtime checks; static validation should not execute arbitrary JavaScript or perform Git resolution.
2. **Actionable errors:** Explain that full commit IDs and revision expressions are unsupported, and suggest `HEAD` or a supported named ref. Do not silently substitute `HEAD`, since it may identify a different commit.
3. **Explicit guidance:** State the named-ref-only restriction in the tool description and `baseRef` documentation, including that omitted `baseRef` defaults to `HEAD` at worktree allocation time.
4. **Regression coverage:** Check static/runtime agreement for literal full 40/64-character IDs, invalid ref syntax, `HEAD`, and named refs. Preserve runtime validation and existing branch-name restrictions.

Full-SHA support, if desired, can be considered separately as a contract change.

## Workaround

When the intended baseline is the current checkout, omit `baseRef` or use `"HEAD"`, and verify the resulting handoff’s `baseCommit`. Do not blindly replace an arbitrary requested commit with `HEAD` or shorten a SHA merely to bypass the lexical restriction.

jeanduplessis:
## Summary

`action: "validate"` accepted literal full commit IDs that workflow execution rejected. The generic “valid Git ref” error did not explain the restriction introduced in #1850.

- Check statically known `baseRef` values in `runs.run` and `runs.all`, with source locations.
- Respect final property definitions and leave dynamic values, uncertain overwrites, and accessors to runtime.
- Explain unsupported full commit IDs and revision expressions in errors, tool guidance, and docs. Clarify that the default `HEAD` resolves at worktree allocation.

Runtime ref predicates, branch-name validation, Git resolution, and clean-checkout safeguards remain unchanged. No full-SHA support, new API, or dependency.

Closes #1934.

## Validation

- Full Node unit suite: 2,842 passed, 3 skipped, 0 failed.
- Full Node integration suite: 953 passed, 6 skipped, 0 failed.
- Five targeted Bun 1.4.0 tests passed.
- Typecheck and diff hygiene passed.
- Fresh read-only review found no material defects. An independent boundary probe verified worker and host rejection, including zero dispatch for an invalid `runs.all` batch.

Local validation used macOS/Node 24 and isolated temporary state. No real child model calls were made. Cross-platform CI remains pending.


nicobailon:
Superseded by maintainer replacement PR #1942, which has landed with the original contribution preserved and credited. Thanks again for the report and implementation.

PR #1947, opened 2026-09-05
test: distinguish acceptance runner exit boundaries

nicobailon:
Refs #1916. Diagnostic-only follow-up to #1918; this does not claim a Windows shutdown fix or close the investigation.

Add test-local exit-request/dispatch-return/throw markers, scoped native trace-exit, bounded runtime/cache-state metadata, and one failure-only Windows identity query. Original observed terminal-proof predicate, assertions, deadlines and cleanup ownership remain unchanged. No production/cache policy change, retry, or new dependency.

Validation: existing acceptance test passed12/12 on Darwin; final syntax and hygiene passed. Typecheck passed but excludes integration tests. Retained challenge and separate simplification were unchanged; fresh source review found no failures. Windows/CIM/native argument parsing and the shared two-second query budget remain unexercised locally and require CI evidence. Main remains red; this PR gathers discriminating evidence rather than declaring the cause resolved.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,8 @@ jobs:
         run: npm test
       - name: Integration tests
         timeout-minutes: 10
+        env:
+          NODE_DISABLE_COMPILE_CACHE: '1'
         run: npm run test:integration
 
   bun-workflow-script:
--- a/test/integration/acceptance-file-report.test.ts
+++ b/test/integration/acceptance-file-report.test.ts
@@ -7,7 +7,7 @@
  */
 import { after, afterEach, before, beforeEach, describe, it } from "node:test";
 import assert from "node:assert/strict";
-import { ChildProcess } from "node:child_process";
+import { ChildProcess, execFileSync } from "node:child_process";
 import { channel } from "node:diagnostics_channel";
 import * as fs from "node:fs";
 import * as path from "node:path";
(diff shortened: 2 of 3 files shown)


nicobailon:
The c315 Windows result confirmed completed JS exit dispatch but no native exit warning, with Node compile cache enabled. 82b0ad46 adds one test-local control: disable only Node compile cache for the first affected launch and restore the parent environment immediately. It preserves deadlines, assertions, ownership and the identity-query budget. One bounded post-proof measurement verifies treatment. Local12/12 passed with actual cache=false, native warning, exit/close and observed proof; this is not Windows evidence or a root-cause claim. This push supplies one Windows control attempt. Its outcome will be inspected before any further change; no automatic retry or production cache policy.

nicobailon:
The Windows control succeeded with verified cache=false and native exit/close/proof 41ms after dispatch. The immediately following untreated case reproduced the cache=true stall. 104b85ef confines cache isolation to the CI integration-test step (production and real clean-install smoke unchanged). It also replaces the separate auto-resume test’s bespoke two-second final-state loop with its existing logical-completion helper; strict state/phase/count assertions remain. The two-file +4/-5 final delta passed focused validation and fresh targeted review. This is the first landing candidate once all exact-head gates pass; required main CI follows before feature merges. No native-deadlock claim or CI waiver.

PR #1954, opened 2026-09-05
fix: make false-only tool schemas portable for Gemini converters

### LINKED ISSUE #1950: Gemini 400: boolean-literal enum [false] in subagent tool schema rejected by function_declarations (TYPE_STRING)
## Summary

When the `subagent` tool is registered and the active model routes to a **Google Gemini** endpoint (via an OpenAI-compat gateway that converts tool schemas to Gemini `function_declarations`), every request fails with a 400 before any generation happens:

```
400: {"message":"Invalid value at 'tools[0].function_declarations[4].***.properties[0].value.any_of[3].enum[0]' (TYPE_STRING), false
Invalid value at 'tools[0].function_declarations[4].***.properties[38].value.any_of[1].enum[0]' (TYPE_STRING), false
Invalid value at 'tools[0].function_declarations[4].***.properties[67].value.any_of[1].enum[0]' (TYPE_STRING), false","type":"upstream_error","code":400}
```

`function_declarations[4]` is the `subagent` tool. The three rejected paths are exactly the three schemas that use a **boolean-literal `enum`**.

## Environment

- **pi-subagents:** 0.65.1
- **pi:** current npm package
- **Model:** `gemini-3.7-flash` behind an OpenAI-completions gateway (sub2api) that translates OpenAI tool JSON Schema → Gemini `function_declarations`
- Models without strict schema validation (GLM, GPT via the same gateway) work fine — only Gemini rejects it.

## Root cause

Google's function-declaration schema requires `enum` entries to be **strings (TYPE_STRING)**. JSON Schema permits `enum: [false]` alongside `type: "boolean"`, but Gemini's validator rejects non-string enum members outright.

`src/extension/schemas.ts` has three boolean-literal enums:

| Line | Schema | Field |
| ------ | -------- | ------- |
| 92 | `AcceptanceOverride` | `{ type: "boolean", enum: [false] }` as `anyOf[3]` → matches `properties[0].value.any_of[3].enum[0]` |
| 260 | `MissionLaunchOverride` | `{ type: "boolean", enum: [false] }` as `anyOf[1]` → matches `properties[38].value.any_of[1].enum[0]` |
| 320 | `thinking` (ControlOverrides) | `{ type: "boolean", enum: [false] }` as `anyOf[1]` → matches `properties[67].value.any_of[1].enum[0]` |

Example (schemas.ts:258-263):

```ts
const MissionLaunchOverride = Type.Unsafe({
 anyOf: [
  { type: "object", additionalProperties: true },
  { type: "boolean", enum: [false] },
 ],
});
```

The `enum: [false]` is redundant anyway — `type: "boolean"` already restricts the value space to `true | false`, and the runtime treats any boolean as valid for these fields (only `false` is meaningful, but the enum isn't what enforces that).

## Suggested fix

Drop the three `enum: [false]` occurrences and keep `type: "boolean"`:

```diff
-  { type: "boolean", enum: [false] },
+  { type: "boolean" },
```

(same change at lines 92, 260, and inside the `thinking` `anyOf` at line 320)

I've applied this patch locally and Gemini requests now succeed. This is also consistent with the direction taken in #126 (schema compatibility across strict providers) — that fix removed parent-level `type` arrays but left these boolean enums in place.

## Repro

1. Configure a Gemini model behind an OpenAI-compat → Gemini translating gateway (or native Gemini API):

   ```json
   { "id": "gemini-3.7-flash", "name": "Gemini 3.7 Flash", "reasoning": true, "input": ["text", "image"], "contextWindow": 204800, "maxTokens": 65536 }
   ```

2. Load pi-subagents (any provider config that registers the `subagent` tool).
3. Send any message.
4. Observe the 400 above; switching the same session to a non-Gemini model works immediately.

nicobailon:
## Summary
- Remove three boolean-literal enums rejected by restricted Gemini function-declaration converters. Keep the false-only semantic contracts; this does not enable `true` as a supported option.
- Validate workflow acceptance defaults before mission/storage/script execution, including childless workflows.
- Credit [@Biaogo94](https://github.com/Biaogo94) for the report and local reproduction.

Closes #1950.

## Validation
- 109 focused schema/acceptance/mission/watchdog tests and typecheck passed on the original implementation.
- Public workflow-default regression reproduced the gap, then passed with the guard; three targeted integration cases, gate normalization and typecheck passed.
- Combined challenge/cleanup and independent review completed. The review's workflow-default finding was fixed and independently re-reviewed.
- Current-main refresh preserves the non-changelog patch byte-for-byte and contributor credit. New exact-head CI remains required.

Compatibility claim is limited to restricted function-declaration converters; native Gemini's `parametersJsonSchema` path is not claimed universally broken. No live provider run. Existing lint debt is unchanged: 124 baseline diagnostics, none added. Validation precedes executor launch effects, not trusted resource resolution or raw-file reads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Surface the published workflow receipt path in wait completions, notifications, and exact status/debug responses (#1938).
 - Surface structured output in completion notices when text is blank or only a closing think-tag (#1945). Thanks to [@npfedwards](https://github.com/npfedwards).
 - Validate literal workflow `baseRef` values before execution and clarify supported refs (#1934, #1937). Thanks to [@jeanduplessis](https://github.com/jeanduplessis).
+- Use portable boolean tool-schema branches for restricted Gemini function-declaration converters while preserving false-only runtime contracts (#1950). Thanks to [@Biaogo94](https://github.com/Biaogo94).
 - Preserve retained predecessor lineage in string-ID workflow continuations and receipts (#1920).
 - Clear recovered assistant errors on successful tool-use completion, including terminating structured output (#1919). Thanks to [@Jonathanm10](https://github.com/Jonathanm10).
 - Diagnose empty terminal text responses as empty-output failures rather than earlier exploratory tool errors (#1921).
--- a/src/extension/schemas.ts
+++ b/src/extension/schemas.ts
@@ -77,6 +77,8 @@ const AcceptanceEvidenceKinds = [
 	"manual-notes",
 ];
 
+// Provider boolean branches intentionally overapproximate false-only runtime inputs.
+// Restricted function-declaration converters only support string enum members.
 const AcceptanceOverride = Type.Unsafe({
 	anyOf: [
 		{ type: "string", enum: ["auto", "attested", "checked"] },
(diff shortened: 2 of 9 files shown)


nicobailon:
Fixed the unrelated setup-hook fixture failure in ce2b51f6: the generated test hook now drains stdin instead of racing the parent write with process exit. One test-fixture line changed; all three existing POSIX lifecycle cases pass. No production hook/error/timeout behavior changed. Original failed CI is retained, and new exact-head CI is required.

PR #1972, opened 2026-09-05
feat: advertise opt-in agents in parent prompt

### LINKED ISSUE #1971: Allow opt-in subagents to advertise themselves in the parent prompt
The parent currently discovers agent names and descriptions only by calling `subagent({ action: "list", capabilities: true })`. This creates a catch-22: it must first decide delegation is appropriate before discovering that a relevant specialist exists.

Add opt-in agent metadata, such as `advertise: true` or `promptHint`, that pi-subagents dynamically includes in the parent-facing prompt. The agent definition would own its routing guidance, while pi-subagents owns discovery and injection.

Only explicitly advertised agents should be included to limit prompt size and avoid exposing the full registry. This differs from #195: it requests dynamic metadata from discovered agents, not hardcoded builtin names or deterministic natural-language routing.

I'm asking for this because I have a YouTube transcript subagent that I want my agents to use to pull transcripts.  However, they don't ever see the description of that subagent so don't know they should use it unless I explicitly tell them or make some static routing rule in my `AGENTS.md`, but imo that's the wrong place for that.  The subagent definition should own whether it gets advertised to the parent.

nwalke:
Parent agents currently discover specialist names and descriptions only after calling `subagent({ action: "list", capabilities: true })`. That creates a routing gap: the parent must decide to delegate before knowing a relevant specialist exists.

 This adds opt-in parent-prompt discovery through `advertise: true`.

 - Supports file-defined, runtime-registered, and settings-overridden agents.
 - Injects advertised names and descriptions only when the subagent tool is active.
 - Excludes disabled agents and agents blocked by the active capability ceiling.
 - Keeps the catalog bounded, deterministically ordered, and XML-escaped.
 - Still requires capability listing before execution.
 - Exposes advertisement state through management and capability output.

 No agents are advertised by default, so the full registry is never injected automatically.

 Validation

 - npm run typecheck
 - Focused agent discovery, serialization, override, runtime registration, and prompt-injection tests
 - npm run test:integration
 
 Fixes #1971


nicobailon:
Thanks @nwalke for the opt-in agent discovery design and implementation. Replacement #1985 has merged, preserving your contributor commits, merge coauthor attribution, and linked changelog credit. It keeps file-defined advertisement with bounded catalogs and explicit refresh; broader runtime/settings advertisement remains deferred.

PR #1974, opened 2026-09-05
fix: tolerate unconfigured prompt runtime loads

### LINKED ISSUE #1973: Async in-process children can fail while loading prompt runtime extension
### Environment
- pi-subagents: **0.65.1** (latest published)
- Pi coding agent: **0.85.0**
- OS: macOS

### Symptom
Launching any async (`async: true`) subagent workflow that spawns native children fails at startup: every child dies **before running a single tool** (0 tool calls, 0 tokens consumed, pre-start). The run/board reports the child load error:

```
Failed to load extension ".../pi-subagents/src/runs/shared/subagent-prompt-runtime.ts":
Cannot read properties of undefined (reading 'runtimeAcknowledgements')
```

The very first workflow of the session was lost this way (all children failed, no code produced). A second attempt failed identically until I patched `subagent-prompt-runtime.ts` locally.

### Confirmed source facts
- `child-launch.ts` in `v0.65.1` already filters `isSubagentRuntimeExtensionPath()` from `additionalExtensionPaths` before constructing the child session.
- The same launch sets `ambientExtensions` for runner-hosted children when ambient extensions are enabled.
- `child-session.ts` passes `noExtensions: !launch.ambientExtensions`, the filtered `additionalExtensionPaths`, and the inline `extensionFactories` to Pi's `DefaultResourceLoader`.
- The runner sets `PI_SUBAGENT_CHILD=1`; `index.ts` is intended to skip parent registration when that value is present.

### Current diagnosis
The observed pre-start error is real and reproducible in the affected local run:

```
Failed to load extension ".../pi-subagents/src/runs/shared/subagent-prompt-runtime.ts":
Cannot read properties of undefined (reading 'runtimeAcknowledgements')
```

The original hypothesis that `child-launch.ts` directly leaked the prompt-runtime path is **ruled out for v0.65.1**: that path is explicitly filtered. The remaining likely compatibility issue is the runner's ambient-extension discovery/reload path: Pi may rediscover a `pi-subagents` entry point while constructing an in-process child, and the child-side guard or load ordering may not cover every entry path. This exact bypass point is not proven by the current source inspection, so it should be treated as a diagnosis request rather than a confirmed Pi-core defect.

The launch metadata's `effective` extension list is not sufficient proof of the leak: it projects `toolPlan.extensionArgs` before the runtime-path filter is applied.

### Reproduction
On Pi `0.85.0` + `pi-subagents` `0.65.1`, an async workflow that launches native children produced the error before any child tool call. The affected run had 0 tool calls and 0 tokens consumed. Foreground/non-async launches and the config-passing inline hook were unaffected in the observed case.

### Local workaround applied
Made `config` optional and returned early when absent in `subagent-prompt-runtime.ts`:

```ts
export default function registerSubagentPromptRuntime(pi: ExtensionAPI, config?: ChildRuntimeConfig): void {
  if (!config) return;
  registerRuntimeExtensionAcknowledgements(pi, config.runtimeAcknowledgements);
  // ...
}
```

This is defense-in-depth: an accidental path-based load no longer crashes the child, while the configured inline hook still registers normally.

### Suggested maintainer investigation
Please instrument or trace `DefaultResourceLoader`'s ambient discovery and reload sequence for runner-hosted in-process children. Confirm whether `pi-subagents/index.ts` or one of its declared extension entries is loaded in the child, whether `PI_SUBAGENT_CHILD` is present at that point, and whether the inline prompt-runtime factory is invoked before/after any discovered copy. Keep the optional-config guard as defense-in-depth; if a concrete duplicate path is found, prevent the runtime-owned modules from being discovered or loaded twice.

Happy to file a PR once the exact owner-side fix location is confirmed.

lertian:
## Summary

- Make the child prompt-runtime extension tolerate being loaded without a `ChildRuntimeConfig`.
- Add a regression test for an unconfigured path-based load.

## Context

An async in-process child on Pi 0.85.0 with pi-subagents 0.65.1 failed before its first tool call with:

```
Cannot read properties of undefined (reading 'runtimeAcknowledgements')
```

The runtime is normally installed through the configured inline child hook. However, if extension discovery or reload also invokes the module as a path-based extension, that invocation has no runtime config. Returning without registering hooks keeps the accidental invocation inert while preserving the normal configured path.

This is intentionally defense-in-depth. The exact ambient-extension discovery/reload path still needs confirmation; the existing `child-launch.ts` filtering of known runtime paths remains unchanged.

Closes #1973.

## Validation

- `npm run typecheck`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/subagent-prompt-runtime.test.ts` (50 passed)
- `git diff --check`

### DIFF
--- a/src/runs/shared/subagent-prompt-runtime.ts
+++ b/src/runs/shared/subagent-prompt-runtime.ts
@@ -441,7 +441,12 @@ function registerStructuredOutputTool(pi: ExtensionAPI, structured: NonNullable<
 }
 
 /** Register every child-side hook the prompt runtime owns for one child session. */
-export default function registerSubagentPromptRuntime(pi: ExtensionAPI, config: ChildRuntimeConfig): void {
+export default function registerSubagentPromptRuntime(pi: ExtensionAPI, config?: ChildRuntimeConfig): void {
+	// A path-based load has no ChildRuntimeConfig. This can happen if an
+	// ambient-extension discovery path finds the runtime module in addition to
+	// the configured inline factory. It must be inert rather than crashing the
+	// child before startup; the inline factory remains the real registration path.
+	if (!config) return;
 	registerRuntimeExtensionAcknowledgements(pi, config.runtimeAcknowledgements);
 	registerPermissionGate(pi, config.permissions, config.childWatchdog);
 	registerToolBudget(pi, config.toolBudget);
--- a/test/unit/subagent-prompt-runtime.test.ts
+++ b/test/unit/subagent-prompt-runtime.test.ts
@@ -71,6 +71,10 @@ const PROMPT_WITH_EXPLICIT_SKILL = [
 const CONFIGURED_SKILLS_SECTION = "\n\nThe following configured skills are available to this subagent.\nUse the read tool to load a skill's file when the task matches its description.\nWhen a skill file references a relative path, resolve it against the skill directory (parent of SKILL.md / dirname of the path) and use that absolute path in tool commands.\n\n<available_skills>\n  <skill>\n    <name>configured-skill</name>\n    <description>explicit agent skill</description>\n    <location>/tmp/configured-skill/SKILL.md</location>\n  </skill>\n</available_skills>";
 
 describe("subagent prompt runtime", () => {
+	it("ignores an unconfigured path-based load", () => {
+		assert.doesNotThrow(() => registerSubagentPromptRuntime({} as never));
+	});
+
 	it("registers no permission hook by default and routes ask only to the watchdog arbiter", async () => {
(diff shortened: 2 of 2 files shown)


nicobailon:
Thanks @lertian. Merged your fix and regression unchanged through #1984 (7fe9dee1bc186592e3f2b95c07d86c02f2edd57a), with your original commit retained in the preparation branch and explicit changelog credit. This fixes the unconfigured module-load boundary; we are not claiming to have proven the ambient rediscovery cause. Closing this PR as superseded by the landed replacement.

Issue #2006, opened 2026-09-06
Add agy-writer external CLI adapter for Antigravity CLI


Xernz:
## Summary

Add a code-owned `agy-writer` adapter and packaged agent for the Antigravity CLI (agy). The adapter follows the existing external-agent naming convention: the CLI identity plus the `-writer` suffix.

## Proposed contract

The adapter uses the existing external CLI runner with stdin prompt delivery and fixed arguments:

```text
agy --input-format text --output-format stream-json --mode accept-edits
```

It requires an installed agy CLI with existing local authentication. Its receipt and status metadata report `workspace-write` with permission mode `accept-edits`. The adapter owns its argv and does not pass permission-bypass, continuation, conversation, or model flags. User profiles cannot add argv.

The stream-json parser accepts JSONL progress events and requires a terminal `result` event with `status: "SUCCESS"` and a non-empty `response`. Malformed events, error results, missing terminal output, and events after the terminal state fail closed. The runner keeps its existing bounds, stop and timeout handling, and artifact capture.

Launch preflight checks `agy --version` and `agy --help`. agy prints help on stderr, so the shared preflight accepts stderr when stdout is empty while still rejecting probe errors and non-zero exits.

## Why there is no read-only adapter

There is deliberately no `agy` read-only adapter. Experiments could not establish a reliably enforced read-only contract for agy's `--mode plan`, including runs with restrictive permission settings. A write can occur, and unsuccessful quota or authentication runs do not prove that a write was blocked. Advertising `access: "read-only"` would conflict with pi-subagents' requirement that external agents declare only capabilities supported by evidence.

Advisory agy analysis can remain a user-local external-cli profile with its own settings and an explicit, local understanding of the risks. This change makes no read-only capability claim.

## Scope and validation

- Register and package only `agy-writer`.
- Persist its safety metadata through status and workflow receipts.
- Test stdin delivery, fixed argv ownership, stream-json terminal proof, failure handling, preflight validation, and receipt round-tripping.
- Provide an opt-in maintainer smoke that requires successful completion and `writeCanaryMatches: true`.
- Keep the smoke out of normal CI. It requires an authenticated CLI home and an existing disposable workspace trusted by agy.

nicobailon:
[...]

PR #2007, opened 2026-09-06
feat(runs): add agy-writer external CLI adapter

Xernz:
## Summary

Adds the code-owned `agy-writer` external CLI adapter for the Antigravity CLI.

- Launches agy with fixed stdin and stream-json arguments in `accept-edits` mode.
- Reports `workspace-write` and existing local authentication in status and workflow receipts.
- Parses bounded JSONL output and requires a successful terminal result with non-empty text.
- Validates agy version and help during launch preflight. The shared preflight now accepts help emitted on stderr when stdout is empty.
- Adds a packaged `agy-writer` profile, unit coverage, and an opt-in maintainer smoke.

There is no read-only `agy` adapter. Experiments could not establish a reliable read-only boundary for agy's model-guided `--mode plan`, including runs with restrictive permission settings. The adapter therefore makes no read-only claim and reports only the writer capability it is intended to use. See #2006.

## Validation

- `npm run test:unit`, 2,846 passed, 4 skipped
- `npm run typecheck`
- `git diff --check`
- Opt-in smoke documented in `docs/agents.md`; it requires an authenticated agy home and an existing disposable trusted workspace.

## Scope

The commit changes 15 files. `package-lock.json` is unchanged.


nicobailon:
Thanks @Xernz for the contribution and for carefully investigating the CLI’s permission boundaries. I’m not interested in taking on built-in Antigravity CLI support at the moment. With demand currently coming from one person, I don’t want to add the ongoing adapter maintenance burden. I’m closing this and #2006 for now, but I’m open to reconsidering if more users express a need for it. Thanks again for the work!

PR #2012, opened 2026-09-06
feat(context): fork from seed session files with fork:<path> context values

amemifra:
## Summary

Extends subagent launch-context resolution so that `context` (tool input), agent frontmatter `defaultContext`, and config `defaultSubagentContext` accept a third form, `fork:<seed-session-path>`, in addition to `fresh` and `fork`.

A seed-backed fork branches from the named session file's current leaf (manager-reported leaf, or the last entry id as fallback) instead of the parent session. It reuses the existing branched-session creation, Anthropic thinking-block sanitization, and pruned-fork plumbing unchanged. Paths support `~` expansion.

## Motivation

Fork context currently always branches from the parent conversation. That makes it impossible to give an agent a durable, curated starting context: every launch either starts fresh (no context) or inherits whatever happens to be in the current parent session.

With seed sessions, a project can maintain one or more curated conversations (e.g. a role-specific expert briefing, a project post-mortem, an architecture primer) and have every child of that role start from that context — without re-injecting briefing content into every task prompt.

```md
---
name: trace-runner
defaultContext: fork:/path/to/expert-briefing-session.jsonl
---
```

## Semantics

- Parent-seeded `fork` behavior is unchanged, including implicit-fork fallback to `fresh` and fail-fast validation when no per-index seed sources are configured.
- Seed forks are **strict**: a missing seed file fails the launch with a clear error instead of silently falling back to `fresh` (silently starting fresh would be indistinguishable from success while losing the entire point of the seed).
- `fork:` with an empty path is not a fork value (rejects cleanly in config validation).
- Per-agent/per-index sources are supported: in a multi-child launch, each child resolves its seed from its own agent's `defaultContext`.

## Implementation notes

- `createForkContextResolver` gains an optional `forkSourceForIndex(index)`; the executor records each task's agent-resolved seed before forking, with a selected-agent fallback for calls that race ahead of bookkeeping.
- Parent-session existence/leaf checks become lazy only when per-index seed sources are configured; otherwise fail-fast behavior is preserved exactly.

## Testing

- New unit tests cover: `fork:<path>` parsing, branching from the seed file, last-entry-id leaf fallback, missing-seed error, no-parent-session-required when all indexes use seeds, and mixed seed/parent fail-fast.
- `npm run typecheck` clean; `npm run test:unit`: 2968 tests, 0 failures.
- Docs updated (`docs/agents.md`, `docs/configuration.md`) + CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Added
+- `fork:<seed-session-path>` context values: `context` (tool input), agent frontmatter `defaultContext`, and config `defaultSubagentContext` may now name a seed session file that the child branches from instead of the parent session. Seed forks are strict (a missing seed file fails the launch) and support `~` expansion. Parent-seeded `fork` behavior is unchanged.
+
 ## [0.66.0] - 2026-09-06
 
 ### Highlights
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -251,6 +251,7 @@ Use these fields when an agent should see more:
 | `inheritGlobalContext: true` | Also keep the operator's global context file from the Pi config agent directory (such as `~/.pi/agent/AGENTS.md`). Defaults to `false`. |
 | `inheritSkills: true` | Let the child see Pi's discovered skills catalog. |
 | `defaultContext: fork` | Prefer forked session context when a launch omits `context`; if the parent has no persisted session file or current leaf yet, the implicit default falls back to `fresh` without a failed first attempt. Explicit `context: "fork"` remains strict, and explicit `context: "fresh"` still wins. |
+| `defaultContext: fork:<seed-session-path>` | Always fork from the named seed session file instead of the parent session. The path may be absolute or `~`-expanded. Seed forks are strict: a missing seed file fails the launch with a clear error instead of silently falling back to `fresh`. Useful for role expert-context seeds — maintain a curated conversation and have every child of that role start from it. |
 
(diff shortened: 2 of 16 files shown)


nicobailon:
Thanks, @amemifra, for the proposal and implementation. I am closing this as not planned in its current form. Durable role briefings are useful, but we want to keep them in existing agent prompts, defaultReads and skills rather than add seed-session ancestry and ownership across global defaults, agent configuration and public launch APIs. This proposal does not yet establish the session-tree requirement those existing mechanisms cannot meet.

The exact-head review also found concrete blockers: the existing integration test “uses global defaultSubagentContext fork for a fresh-default agent” fails; disk agent frontmatter silently drops the advertised seed default; and the public tool schema and structured delegation validator reject the new context values. Seed source placement and the host opener potentially migrating the source file also need an explicit ownership contract. The changelog conflict itself is not the reason for closure.

I do not want to ask you to repair all those seams before agreeing that we should own this broader API. If a concrete use case truly requires preserved conversation-tree state rather than reusable briefing content, an issue describing that requirement would be a better starting point for a much narrower design.

amemifra:
Thanks @nicobailon for the thoughtful review and clear rationale — completely fair call.

I agree the durable-briefing need is better served by the existing mechanisms (agent prompts, `defaultReads`, skills) for anything that's reusable content, and the review correctly caught real seams I hadn't covered (global-vs-agent default precedence, the public schema boundary, the frontmatter drop, and the session-ownership question for the host opener).

For context: my immediate use case was role-scoped expert briefings for subagents on a single-user machine, which is indeed reusable content — so existing mechanisms should cover it, and I'll go that route. If we ever hit a case that genuinely requires preserved conversation-tree state (rather than static briefing content), I'll open an issue describing that requirement first, as you suggested.

Thanks again for the time spent on a thorough review of an outside PR.

PR #2025, opened 2026-09-06
fix: cancel final drain when child work resumes

harche:
## Summary

Cancel a child’s previous final-drain window when `turn_start`, `agent_start`, or `auto_retry_start` signals resumed work. Clear the corresponding completion flags in both native hosts so a late watchdog status cannot rearm the obsolete drain.

## Problem

A terminal assistant message does not necessarily mean the child session has finished. Queued steering or follow-up input can immediately start another turn:

1. The child emits a successful assistant message with `stopReason: "stop"`.
2. The host arms the 1,000 ms final-drain timer.
3. Pi consumes queued input and emits `turn_start` for the continuation.
4. The old timer is not cancelled and calls `session.abort()` while that continuation is running.

This was observed as `OpenAI Responses stream ended before a terminal response event`, with an aborted, empty assistant message about 1,000 ms after a successful report. The error is a consequence of the host aborting continued work, not evidence that the preceding report failed to stream.

The shared lifecycle projection previously returned no action for resumed turns; agent/retry starts only reset the compaction flag. Returning `cancel-drain` disarms the existing final-drain and watchdog-tail timers. Resetting both completion flags also prevents a subsequent idle watchdog notification from treating the previous answer as the end of the current work.

The existing final-drain grace period, explicit stop/timeout handling, and provider retry policy are unchanged. No dependency or public API changes are included.

## Tests

Added lifecycle unit tests and scripted integration regressions for both foreground and background native children. The continuation takes 1,400 ms, longer than the old final-drain grace; tests require its output rather than accepting the earlier report. A background watchdog regression covers stale completion flags.

Verified on Node v26.8.1 / macOS against base `e65cbb24835de2c729fa0efe8945959b016bf389`:

- New targeted tests before the production fix: **11 failed, 1 passed, 0 skipped**. Failures were assertion failures: obsolete output or `none` instead of `cancel-drain`.
- Same targeted tests after the fix: **12 passed, 0 failed, 0 skipped**.
- Full unit suite: **2,979 passed, 0 failed, 13 skipped**.
- `test/integration/in-process-child.test.ts` and `test/integration/async-execution.part-4.test.ts`: **57 passed, 0 failed, 0 skipped**, including existing stop/timeout, stuck-session drain, and watchdog-tail cases.
- `git diff --check`: passed; package manifests and lockfile unchanged.

Dependencies were installed with `npm ci --ignore-scripts`. Tests used isolated home/temp directories and scripted child sessions; no live-model test or full integration-suite run is claimed. CI covers the repository’s supported platforms.

## AI assistance

Pi assisted with diagnosis, implementation, regression tests, and an independent review. The submitting author remains responsible for the changes.

### DIFF
--- a/src/runs/background/run-child-session.ts
+++ b/src/runs/background/run-child-session.ts
@@ -334,6 +334,8 @@ export function runChildSession(input: RunChildSessionInput): Promise<RunChildSe
 		}
 		const applyChildLifecycle = (action: ChildLifecycleAction): void => {
 			if (action === "cancel-drain") {
+				cleanTerminalAssistantStopReceived = false;
+				agentSettledReceived = false;
 				clearFinalDrainTimers();
 				clearWatchdogTailTimer();
 				return;
--- a/src/runs/foreground/execution.ts
+++ b/src/runs/foreground/execution.ts
@@ -746,6 +746,8 @@ async function runSingleAttempt(
 		}
 		const applyChildLifecycle = (action: ChildLifecycleAction): void => {
 			if (action === "cancel-drain") {
+				cleanTerminalAssistantStopReceived = false;
+				agentSettledReceived = false;
 				clearFinalDrainTimers();
 				clearWatchdogTailTimer();
 				return;
--- a/src/runs/shared/child-lifecycle.ts
+++ b/src/runs/shared/child-lifecycle.ts
@@ -14,8 +14,9 @@ export function projectChildLifecycle(event: { type?: string; willRetry?: unknow
 		if (state) state.compactionRetryActive = event.willRetry === true;
 		return event.willRetry === true ? "cancel-drain" : "none";
 	}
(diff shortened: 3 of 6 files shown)


nicobailon:
Thanks, @harche. Your fix landed in #2028 (45a405a389d56a65895ba6866cdcd41dd1d5683c), with the original code and tests unchanged, your co-authorship preserved, and an Unreleased changelog credit added. Exact-head CI and Greptile passed; closing this original as superseded by the landed replacement.

harche:
> Thanks, @harche. Your fix landed in #2028 ([45a405a](https://github.com/nicobailon/pi-subagents/commit/45a405a389d56a65895ba6866cdcd41dd1d5683c)), with the original code and tests unchanged, your co-authorship preserved, and an Unreleased changelog credit added. Exact-head CI and Greptile passed; closing this original as superseded by the landed replacement.

Thanks.

Issue #2020, opened 2026-09-06
0.66.0: Cannot find module '@earendil-works/pi-tui' — imported but no longer a dependency


kroediger:
# pi-subagents 0.66.0: `Cannot find module '@earendil-works/pi-tui'` — Dependency fehlt, Import bleibt

**Betrifft:** pi-subagents **0.66.0** (via `pi update npm:pi-subagents`), pi-coding-agent 0.84.4, macOS.

## Symptom

Jeder Start eines Subagenten (worker/reviewer) stirbt im Runner sofort:

```
Error: Cannot find module '@earendil-works/pi-tui'
Require stack:
- .../pi-subagents/src/watchdog/register-main.ts
    at .../pi-subagents/src/watchdog/register-main.ts:2:20
    at import (.../pi-subagents/src/watchdog/tool-actions.ts:5:27)
    ...
```

## Ursache

`src/watchdog/register-main.ts` Zeile 2 importiert zur Laufzeit:

```ts
import { Text } from "@earendil-works/pi-tui";
```

Aber **0.66.0 deklariert `@earendil-works/pi-tui` nicht mehr in `dependencies`** (in 0.65.1 war es noch dabei). Nach `pi update npm:pi-subagents` wird das Paket deshalb nicht installiert, und der Import bricht — bis das Paket manuell in der Agent-npm-Wurzel nachgezogen wird.

## Workaround (bis zum Fix)

```
npm install @earendil-works/pi-tui@0.81.0   # in ~/.pi/agent/npm
```

Danach starten Subagenten wieder.

## Erwartung

Entweder `@earendil-works/pi-tui` wieder in `dependencies` aufnehmen, oder den Import aus `register-main.ts` entfernen, wenn das Paket bewusst kein Laufzeit-Abhängiger mehr sein soll.

## Reproduktion

1. `pi update npm:pi-subagents` (auf 0.66.0)
2. Subagent starten (z. B. `worker` mit trivialem Task)
3. Runner stirbt mit dem obigen `MODULE_NOT_FOUND`

_Hinweis:_ Neben diesem Fehler gab es zuvor einen zweiten (Inkompatibilität der Inline-Hook-Factory-Signatur, `runtimeAcknowledgements` undefined), der mit 0.66.0 behoben zu sein scheint — der verbleibende Blocker ist ausschließlich die fehlende `pi-tui`-Dependency.

nicobailon:
[...]

PR #2029, opened 2026-09-06
fix: select chord aliases for supported Pi hosts

### LINKED ISSUE #2026: Background preflight rejects Pi 0.84.3 because chord aliases are mandatory
With Pi **0.84.3** and `pi-subagents` **0.66.0**, native background children fail before startup:

```text
<pi-package-root> does not provide @earendil-works/chord,
@earendil-works/chord/context, so the async runner cannot
create child sessions.
```

**Environment:** Nix-installed Pi 0.84.3, pi-subagents 0.66.0, Node 24.19.0, Linux.

An async workflow attempting `delegate` and `scout` children fails for both. Calling `resolveHostPeerAliases(piPackageRoot)` directly reproduces the problem: all 11 other aliases resolve; only those two are missing.

Pi 0.84.3 does not declare or ship `chord`. The installation contains the Pi package, SDK exports, and runtime dependencies—it is not a standalone binary, despite the error’s suggestion.

`src/runs/background/runner-aliases.ts` puts both `chord` aliases in the unconditional `HOST_PEER_ALIASES` list, and `spawnRunner` rejects any missing entry.

This differs from https://github.com/nicobailon/pi-subagents/issues/1944, which addressed experimental server/client aliases on Pi 0.85.1.

Proposed solution: make required aliases match the supported host version, with regression coverage for Pi 0.84.3 without chord. If that version is intentionally unsupported, report the minimum supported Pi version instead.

nicobailon:
## Summary
Require chord aliases only for chord-era hosts; public Pi 0.84.3/0.84.4 no longer fail background preflight solely because chord is absent. Preserve required host TUI/SDK identity and the existing exact-0.85.0 server exception. Unknown/prerelease hosts remain fail-closed.

Closes #2026.

## Validation
Public package manifests establish chord's introduction at Pi 0.85.0. Base/current resolver comparison reproduced then fixed the two missing aliases on real 0.84.3/0.84.4 installs. Eight focused tests, typecheck and hygiene pass. Packaged clean-install smokes passed on 0.84.3, 0.84.4, 0.85.0 and 0.85.1, including watchdog/TUI imports and local-provider session lifecycle (no external model calls).

Retained-writer deletion challenge and fresh independent semantic review passed at `7a67ada5f1520e7d99c7a1251a885f5d4d3142b4`. No new dependency or second SDK instance. #2020 remains installation-evidence-blocked; this PR does not claim that separate report fixed.

Thanks to [@samuela](https://github.com/samuela) for #2026; changelog credit included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ### Fixed
 - Prevent stale final-drain timers from aborting resumed native foreground and background child work. Thanks to [@harche](https://github.com/harche) for #2025.
+- Avoid requiring chord aliases on pre-0.85 Pi hosts while keeping required host runtime aliases fail-closed (#2026). Thanks to [@samuela](https://github.com/samuela).
 - Show exact recorded async child IDs in workflow status and actionable child steering guidance when an owned workflow has no foreground route, without treating queued messages as consumed (#2011).
 - Bound transcript previews by individual line size and total rendered body size, preserving recent context and full artifact references. Thanks to [@rtbe](https://github.com/rtbe) for #2015.
 - Await a bounded, offline model registry refresh before opening `/subagents` model and thinking pickers, and warn on refresh failures. Thanks to [@ianbmacdonald](https://github.com/ianbmacdonald) for #2008.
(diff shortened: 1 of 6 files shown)


samuela:
thanks @nicobailon !

PR #2033, opened 2026-09-06
fix(windows): avoid Windows Terminal Worktrunk collision

Zethu5:
## Summary
- invoke Worktrunk through `git wt` on Windows, which dispatches to its `git-wt` executable
- prevent `worktreeProvider: "auto"` from opening Windows Terminal through its conflicting `wt.exe` alias
- enable the existing Worktrunk probe and provisioning coverage on Windows
- document the platform-specific command

## Root cause
The Worktrunk capability probe spawned bare `wt`. Windows resolves that name to Windows Terminal by default, which opens a Help dialog during worktree-isolated subagent launches. Worktrunk's Windows package instead installs `git-wt`.

Using Git's external-command dispatch avoids the alias without adding a shell boundary. When Worktrunk is absent, the existing native Git fallback remains unchanged.

## Validation
- `npm run typecheck`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/worktree.test.ts`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Fixed
 - Prevent stale final-drain timers from aborting resumed native foreground and background child work. Thanks to [@harche](https://github.com/harche) for #2025.
 - Avoid requiring chord aliases on pre-0.85 Pi hosts while keeping required host runtime aliases fail-closed (#2026). Thanks to [@samuela](https://github.com/samuela).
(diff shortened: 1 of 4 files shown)


Zethu5:
Addressed the missing changelog attribution in `70fea195`.

The extensionless `git-wt` shim is intentional. Production now exercises Git's external-command dispatch (`git wt`), and Git for Windows dispatches extensionless `git-*` commands through its bundled shell. Worktrunk also requires Git for Windows. A `.cmd` shim cannot run through the production `shell: false` boundary, so it would not test this path. The enabled test passed on `windows-latest` before this documentation-only follow-up.

PR #2041, opened 2026-09-07
fix: send x-opencode-session on internal arbiter/watchdog/prompt-audit calls

### LINKED ISSUE #2040: Internal helper calls miss `x-opencode-session` on OpenCode/OpenCode Go
OpenCode emailed users that requests without an `x-opencode-session` header may start erroring on 09/06. Email included at the bottom.

Pi itself sends this header on every normal session request since 0.75.5. But pi-subagents' internal helper calls don't. These are calls the user never sees but the provider counts:
- task-mutation arbiter (the read-only vs implementation check after a delegated task, including the background runner's completion-guard rescue)
- watchdog review
- watchdog permission decisions
- prompt-audit rewrite (prompt redo)

All four build their own bare Agent and stream through raw `streamSimple`, which bypasses pi's provider-attribution header merge. So the helper requests hit OpenCode/OpenCode Go without the session header while every other request carries it.

### Root cause
The header merge lives in pi's session path only (earendil-works/pi#4847). Bare agents sidestep it, so even though these helpers reuse the session's OpenCode model, the header never gets attached.

### Reproduction
Pi >= 0.75.5 with an OpenCode Go model + pi-subagents. Delegate a task, then check provider logs or a proxy: normal session requests have the header, helper requests don't.

IdrisGit:
Background in earendil-works/pi#4847: pi added this header to its session path; OpenCode enforces it from Sep 6.

## What's happening
Pi emits `x-opencode-session` only inside its own session streamFn. Our four internal call sites (task-mutation arbiter, watchdog review, watchdog permission arbiter, prompt-audit rewrite) build bare Agents over raw streamSimple and skip that path entirely, so helper requests go to OpenCode/OpenCode Go unheadered while normal session traffic carries it.

## The fix
One small helper, `src/shared/opencode-session-headers.ts`: returns the two
headers for OpenCode models only (`opencode` / `opencode-go` providers or an `opencode.ai` baseUrl — detection copied from pi's provider-attribution), `undefined` for everything else.

The four call sites merge it in their existing auth wrappers, ahead of request/auth headers, so configured headers still win — same precedence pi uses. Session id comes from the session whose model services we borrow: the parent's for watchdog review, prompt audit, and the foreground arbiter; the child's for the permission arbiter and the detached runner (captured as a plain string next to the existing `completionIntentContext`, no live session held).

## Scope
Headers only. No prompt/tool/timeout/auth changes, no sessionId in agent state, so request bodies for other providers are untouched. Non-OpenCode requests are byte-identical to before.

fixes #2040

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixed
 - Run the child prompt filter before ambient extensions inspect the system prompt. This keeps provider bridges aligned with the final child-visible context while preserving intentional global-context and parent-only skill exclusions. Thanks to [@leftytennis](https://github.com/leftytennis) for #2043.
+- Send OpenCode session-routing headers on internal helper model calls so task arbitration, watchdog, and prompt audit requests keep the same provider session as ordinary Pi traffic. Thanks to [@IdrisGit](https://github.com/IdrisGit) for #2041.
 - Intersect agent frontmatter `tools:` with host-available builtins before spawning children, so hosts with restricted tool menus (e.g. Prime Agent's `ipython`-only set) reject unavailable tools at launch instead of after spawn. Thanks to [@BioInfo](https://github.com/BioInfo) for #2034.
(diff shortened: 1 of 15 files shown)


nicobailon:
@greptileai please review current head b54587dd5c4c11314b1bf647cb9fdd5ca1d709ed. The contributor-credit finding is resolved; the current summary predates this head. Fresh independent source review and all four required CI checks passed.

PR #2051, opened 2026-09-08
fix: reconcile nested work before child completion

shaharmor:
## Summary

Nested reviewers now retain ownership of descendant work through completion or cancellation. A successful early-final recovery delivers terminal results and runs a reconciliation turn instead of accepting a premature "still running" response.

The existing shared child runtime already registers `bg_wait` and drains at `agent_end`. This change reuses that runtime rather than loading the root extension or registering a second wait tool. Reviewer-specific skill/profile changes are separate configuration work, not part of this PR.

## Why the existing drain was insufficient

| Gap | Result |
| --- | --- |
| Wait responses contained status and output-free metadata | A reviewer could not reconcile its descendants' findings |
| Nested native runs used a different storage namespace | Exact-session waits could miss live persona reviewers |
| Native hosts started a one-second final-stop grace before the drain finished | An otherwise valid long drain could be aborted |
| Draining preserved the early final without another model turn | Keeping work alive did not produce a merged report |

## Lifecycle decisions

- Deliver bounded terminal output from result files, watcher records, replay, and archives. Preserve exact artifact references when truncating: 32 KiB per run and 50 KiB per wait/continuation.
- Search ordinary workflow storage and the existing nested route namespace with exact-session filtering. Track fast-terminal receipts as well as currently active work.
- Protect legitimate drains in both native hosts, then require a reconciliation continuation. Missing result evidence and paused work cannot authorize success. Explicitly consumed failed results remain failed but allow a degraded report. Early `structured_output` cannot bypass reconciliation.
- Cascade parent interrupt, stop, and timeout through existing controls. Interrupt pauses supported native descendants; unsupported workflow/external pause boundaries stop with a diagnostic. Explicit foreground stop remains routable after detach. Cancelling only `bg_wait` does not cancel the work.
- Bound cancellation teardown separately from the normal wait window. Unconfirmed cancellation retains durable control/recovery records and reports unresolved run identities.

No model retries are added. Explicit `runs.run({ async: true })` detached-receipt semantics are unchanged; the reviewer regression uses omitted/default-awaited child calls inside an async outer workflow.

## Related

Related: #1844, #2028, #2038.

## Validation

| Check | Result |
| --- | --- |
| Typecheck and `git diff --check` | Passed |
| Focused runtime regressions | 150 passed |
| Full unit suite | 3,002 passed, 13 skipped, **1 dependency-related failure** |
| Full integration suite | 1,004 passed, 14 skipped |
| Native nested lifecycle smoke on Pi 0.84.4 | 8 passed; independently rerun |

The native smoke mocks only the model provider. It exercises actual AgentSessions, the fanout tool, workflow execution, detached native runners, result storage, and control delivery. The main case is root -> reviewer -> two model arms -> persona reviewers: reviewer and arms attempt premature finals, then consume persona evidence and return reconciled finals. Both ordinary and nested storage are checked for remaining owned active runs.

Other native cases cover stop, interrupt, workflow cancellation, timeout, persona failure, and an unresponsive descendant. Unit regressions also cover session isolation, fast completion, replay, truncation references, consumed failures, paused receipts, and early structured output.

### Validation limits

- The local package mirror could not supply `pi-server`/telemetry 0.85 dependencies, and the public registry returned `ENOTCONN`. Tests used exact declared available dependencies plus the existing coding-agent test shim; manifests and lockfiles are unchanged. The sole full-unit failure is `async-spawn-preload.test.ts` importing unavailable `pi-server`.
- Actual native execution was tested with Pi **0.84.4**, not 0.85. Exact 0.85 integration remains unverified locally.
- Windows native execution and real model-provider behavior were not tested locally. CI platform coverage is still needed; the native smoke is opt-in.

To run the native topology smoke against an installed Pi SDK:

```sh
PI_SUBAGENTS_NATIVE_PI_ROOT=/path/to/pi-coding-agent \
node --experimental-strip-types \
  --import ./test/support/isolated-temp-root.mjs \
  --import ./test/support/native-peer-loader.mjs \
  --test test/integration/nested-async-reviewer.test.ts
```

### DIFF
--- a/src/extension/fanout-child.ts
+++ b/src/extension/fanout-child.ts
@@ -6,6 +6,8 @@ import { discoverAgents } from "../agents/agents.ts";
 import { getArtifactsDir } from "../shared/artifacts.ts";
 import { createSubagentExecutor, type SubagentParamsLike } from "../runs/foreground/subagent-executor.ts";
 import { resolveWaitToolConfig } from "../runs/background/wait-config.ts";
+import { cancelOutstandingWork } from "../runs/background/auto-drain.ts";
+import { waitForSubagents } from "../runs/background/subagent-wait.ts";
 import type { ChildRuntimeConfig } from "../runs/shared/child-runtime-config.ts";
 import { readNestedControlRequests, resolveInheritedNestedRoute, type NestedRoute, writeNestedControlResult } from "../runs/shared/nested-events.ts";
 import { deliverSubagentIntercomMessageEvent } from "../intercom/result-intercom.ts";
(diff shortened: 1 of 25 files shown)


nicobailon:
Thanks @shaharmor for the nested-completion work. The narrowed replacement #2069 has merged as 505afdceabc5c6ac9330426690aa7bffb401a309, preserving the descendant-settlement behavior and visible changelog credit. Closing this superseded PR.

Issue #2027, opened 2026-09-06
Notify the parent as individual async workflow children finish


nicobailon:
## Problem

An async `workflowScript` updates child status as each child finishes, but the parent normally receives its completion wake only after the entire workflow finishes. Independent reports can therefore sit ready while the parent remains idle, unable to advance their next actions without polling.

This occurred during a five-child backlog review on main `e65cbb24835de2c729fa0efe8945959b016bf389`:

| Child | Completion elapsed time |
| --- | ---: |
| Auditor review | 89 seconds |
| Packaging investigation | 91 seconds |
| Fork review | 158 seconds |
| Recovery investigation | 161 seconds |
| Drain review | 182 seconds |

The aggregate completion arrived at approximately 182 seconds. The first actionable report waited approximately 93 seconds. The children had distinct isolated worktrees and independent deliverables.

## Current source evidence

- `src/workflows/scripted-workflow.ts:2254–2273` records each normalized child result and terminal trace independently.
- `src/runs/foreground/subagent-executor.ts:5409–5509` projects those transitions into workflow status.
- The async workflow publishes its parent-facing aggregate result after `await runWorkflowScript(...)` returns, around `subagent-executor.ts:5649–5679`.
- `src/runs/background/completion-batcher.ts` defaults to 150 ms debounce and a 1,000 ms maximum wait. That short batching window is not the all-siblings barrier.

These line references describe the exact main commit above. The gap is incremental completion delivery, not missing progress tracking or slow notification batching.

## Desired behavior

When an individual child of an async workflow settles, deliver a compact parent-facing notification with:

- workflow run ID;
- child key and exact child run/attempt ID;
- truthful outcome (completed, failed, paused, or stopped as applicable);
- finalized output/report reference or bounded diagnostic;
- an explicit indication that the enclosing workflow remains running when applicable.

Wake the owning parent so it can consume that result and advance independent work without waiting for siblings or polling status.

Keep `runs.all` semantics unchanged: it still resolves to the ordered result array after all children settle. Preserve the terminal aggregate workflow result/receipt for reconciliation.

## Implementation direction and boundaries

Prefer the default behavior for async workflows, reusing the existing completion ownership, delivery acknowledgement, deduplication, and brief batching infrastructure. Do not add a polling loop, require child-authored progress updates, or introduce a new orchestration API merely to deliver an existing result.

- Notify only after the child's final result and advertised report reference are available. A transient progress status marked completed is not sufficient proof.
[...]

PR #2055, opened 2026-09-08
feat: opt-in quiet completion for durable schedules

pablontiv:
## Summary

Adds an opt-in `quiet: true` flag to `schedule.create`. A quiet schedule's **successful** runs still post their visible `subagent-notify` completion notice, but that notice no longer triggers a parent turn. Failed, stopped, or paused runs wake the session exactly as before. Default behavior is unchanged: nothing is stored or emitted unless `quiet: true` is passed.

Motivation: a recurring schedule (nightly sweep, periodic review) that succeeds routinely currently forces a parent turn on every fire. The operator wants the evidence on screen, not an agent turn spent acknowledging it.

## How it composes

No new primitive. It rides the existing plumbing:

- `ScheduleRecord.quiet?: boolean` persisted next to `sessionOnly` (omitted when false), validated on read.
- `executionParams` forwards it as `scheduleOrigin.quiet: true`, which already flows through `AsyncJobState` and the workflow result payload into `notifier.deliver`.
- `notify.ts` sets `triggerTurn: false` only when `scheduleOrigin.quiet === true` **and** the computed status is `completed`. `display` was already `true` for any scheduled notice, so the output stays visible. An explicit `triggerTurn: false` from the result watcher still wins.

## Changes

- `src/shared/types.ts`: `ScheduleOrigin.quiet`.
- `src/runs/background/scheduled-runs.ts`: record field, parse validation, create validation, `Quiet:` line in create/show output, origin propagation.
- `src/runs/background/notify.ts`: `scheduledCompletionTriggersTurn(origin, outcome)` holds the rule; `deliver` applies it to the final completion.
- `src/runs/foreground/subagent-executor.ts`: the incremental child notice from #2027 applies the same rule, otherwise a quiet schedule's workflow children would still force a turn on every settle.
- `src/extension/schemas.ts`, `src/extension/tool-description.ts`, `src/runs/foreground/subagent-executor.ts`: public param (schema follows the `sessionOnly` precedent with no description to stay under the 18k schema budget; the tool description carries the explanation).
- `docs/missions.md`, `CHANGELOG.md`.

## Tests

- `test/unit/scheduled-runs.test.ts`: persists and reads back `quiet` (details, on-disk JSON, `schedule.show`, fresh manager restore), launch params carry `scheduleOrigin.quiet`, default and explicit `false` store nothing, non-boolean rejected at create and on a corrupted record.
- `test/unit/notify.test.ts`: default scheduled success triggers a turn; quiet success is `display: true` + `triggerTurn: false` with attribution intact; quiet failed/stopped still trigger; a batch of only quiet successes stays silent while a mixed batch wakes; explicit `triggerTurn: false` is not overridden; the shared rule is table-tested for every outcome. The executor send site for incremental child notices has no existing test harness upstream, so it is covered through the shared helper only.

Verified locally with `npm run typecheck`, `npm test`, and `npm run test:integration` on top of current `main`. The same change was first built and verified against `v0.64.0` (the version deployed on the requesting machine); the two conflicts on rebase were CHANGELOG and tool-description text only.

## Out of scope

No new timers, no runtime patching, no change to intercom relay, no change to non-scheduled completions.


nicobailon:
Replaced by #2068, which landed at `ad8d329a`. Thanks to [@pablontiv](https://github.com/pablontiv) for the original contribution.

PR #2056, opened 2026-09-08
feat: distinct colors per agent in fleet status view

savinofiore:
Fleet status renders every agent name in the same muted gray (`theme.fg("muted", agent)`), so in a multi-agent run it's hard to tell which row belongs to which agent without reading the label closely.

This hashes the agent/child name into a fixed palette of 8 ANSI-256 colors, deliberately distinct from the theme's semantic `success`/`error`/`accent` hues so agent identity never collides with state color. Applied at both render sites:
- top-level roster row (`renderEntry`)
- nested descendant rows (`renderNestedRow`)

Hash key is the base agent/name (without the `(modelThinking)` suffix), so the same agent keeps the same color across the roster and its nested rows.

No new deps, no config, no behavior change beyond color.


nicobailon:
Replaced by #2066, which landed at `44982312`. Thanks to [@savinofiore](https://github.com/savinofiore) for the original contribution.

PR #2060, opened 2026-09-08
fix: preserve read-only tasks after host clamping

stekman08:
## Summary

Host capability clamping can remove the mutation tools a task requested before launch. This change allows explicitly read-only review/audit tasks to proceed with the remaining tools, while keeping implementation intent and writer intent conservative when mutation tools are unavailable.

The shared intent handling now recognizes common read-only wording and still gives later implementation imperatives precedence. Acceptance inference retains checked review requirements for sensitive keywords even when the task is read-only.

## Verification

- `node --experimental-strip-types --test test/unit/completion-guard.test.ts test/unit/task-intent.test.ts test/unit/acceptance.test.ts` - 125 passed
- `npm run typecheck` - passed
- `npm run test:integration` - 1003 passed, 6 skipped
- `npm test` - 2989 passed, 13 skipped, 1 unrelated `nested-control` timing failure; the isolated nested-control test passes
- Baseline reproduction blocked `Read-only audit of the current implementation; report findings.` after mutation tools were clamped; the fixed contract allows it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Document task-derived behavior labels for workflow launches in the built-in pi-subagents skill, including reviews and retained follow-ups.
 
 ### Fixed
+- Preserve explicit read-only task intent after host capability clamping, while still blocking implementation tasks that lack mutation tools. Thanks to [@stekman08](https://github.com/stekman08) for #2060.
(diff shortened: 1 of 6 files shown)


nicobailon:
Replaced by #2065, which landed at `ebfc5c55`. Thanks to [@stekman08](https://github.com/stekman08) for the original contribution.

Issue #2045, opened 2026-09-07
Async follow-up rejects the fast field written to its recovery descriptor


isty2e:
## Summary

On pi-subagents 0.66.0 with npm-installed Pi 0.85.1, a native async child launched with `fast: true` completes successfully, but its follow-up fails before the next child starts:

```text
Invalid async recovery descriptor '.../recovery-descriptor.json': unknown field 'fast'.
```

## Reproduction

Run this through `workflowScript` with `async: true`, concurrency 1, and a spawn budget of 2:

```js
return await runs.lanes([{
  key: "fast-resume",
  stages: [
    {
      key: "initial",
      agent: "delegate",
      model: "openai-codex/gpt-5.6-luna:xhigh",
      fast: true,
      task: "Remember K7. Do not use tools. Reply OK."
    },
    {
      key: "follow-up",
      resume: "previous",
      task: "Do not use tools. Reply with the remembered token."
    }
  ]
}]);
```

The initial stage succeeds. The persisted descriptor contains `"fast": true`; the follow-up rejects it.

## Cause

The mismatch is also present on `main` at `56f247ac86169cfffa1af9a4c07997c33804bf9a`:

- [`async-execution.ts`](https://github.com/nicobailon/pi-subagents/blob/56f247ac86169cfffa1af9a4c07997c33804bf9a/src/runs/background/async-execution.ts#L1848) writes the effective `fast` value when enabled.
- [`readAsyncRecoveryDescriptor()`](https://github.com/nicobailon/pi-subagents/blob/56f247ac86169cfffa1af9a4c07997c33804bf9a/src/runs/background/async-resume.ts#L320-L329) omits `fast` from its strict field allowlist.

## Expected

Descriptors written by the extension should round-trip through resume, preserving the effective fast setting. Accept and validate the boolean field, and cover initial execution → follow-up for enabled, explicitly disabled, and omitted settings.

Related: #1417 introduced native fast mode; #654 fixed the same writer/reader mismatch for a different field. Neither reports this `fast` rejection.

PR #2068, opened 2026-09-09
feat: opt-in quiet completion for recurring schedules

nicobailon:
## Summary

Replaces #2055 with a maintainer branch based on current `main`. This preserves the contribution reviewed at exact original head `0585a64cdabd434f0e5a8c98baeaa76a8f5922be` and the approved quiet-schedule behavior at `7038858242a505513d02766a0104515e3899fb3f`.

Adds opt-in `quiet: true` for durable/recurring schedules. A quiet schedule's successful automatic runs still post their visible completion notice, but that notice no longer triggers a parent turn. Failed, stopped, or paused outcomes still wake the session. Default behavior stays noisy.

Quiet is persisted only on recurring `every` schedules. `executionParams()` attaches it for automatic timer and `schedule.run-due` fires, and for `schedule.run` only when that launch passes `quiet: true`. One-shot `at` schedules stay noisy.

RPC `manage` now forwards an explicit `quiet: true` on `schedule.run` through reconstruction into `runManual`. Omitted RPC quiet stays noisy even when the stored schedule is quiet. The flag is not persisted from a manual RPC call.

Thanks to [@pablontiv](https://github.com/pablontiv) for the original contribution in #2055.

## Contract

- Opt-in quiet completion for durable/recurring schedules only
- Default stays noisy
- Quiet does not suppress failure, stop, or pause parent delivery
- Manual runs and one-shot `at` schedules stay noisy unless the caller explicitly asks for quiet on that launch

Does not close #2055.

<div><a href="https://cursor.com/agents/bc-ecbf3003-a300-4da6-9e1c-06ba155d6799?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ecbf3003-a300-4da6-9e1c-06ba155d6799&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

PR #2072, opened 2026-09-09
fix: ignore filename verbs in task-intent classification

### LINKED ISSUE #2039: Task-intent classifier treats verbs inside artifact filenames as implementation requests
## Summary

In pi-subagents 0.65.0, implementation-keyword matching also examines filenames embedded in ordinary task prose. An artifact-render task can therefore be classified as implementation because its output filename contains `-update-`, even though the task does not request a source change.

In the observed background worker run, the requested artifact was produced and verified, but completion was reported as:

```
Subagent completed without making edits for an implementation task.
```

This report is deliberately narrower than whether `worker` is the best role for rendering. Changing only an artifact filename should not change the task's mutation intent. Rendering is not necessarily read-only: it legitimately writes an artifact.

## Reproduction at the classifier boundary

Public, synthetic examples below; no original user prompt or paths are included. In a checkout of the affected version, call the exported function from `src/runs/shared/task-intent.ts`:

```ts
import { classifyTaskMutationIntent } from "./src/runs/shared/task-intent.ts";

const brief = "Render an audio briefing with the existing renderer to artifacts/daily-briefing.mp3.";
const renamed = "Render an audio briefing with the existing renderer to artifacts/daily-update.mp3.";

console.log(classifyTaskMutationIntent("worker", brief));
console.log(classifyTaskMutationIntent("worker", renamed));
```

The second input matches `WORKER_IMPLEMENTATION_PATTERNS` because `update` has word boundaries inside `daily-update.mp3`. This is the same matcher behavior isolated in the original investigation. The synthetic pair is supplied as a minimal reproducer derived from that source analysis, not claimed as a separate end-to-end model run.

Relevant source in 0.65.0:

- `src/runs/shared/task-intent.ts`, `WORKER_IMPLEMENTATION_PATTERNS`: bare `update|add|remove|replace|create` matching.
- `stripFrameworkInstructions` strips some framework-owned output lines, but not filenames occurring in ordinary prose.
- `classifyTaskMutationIntent` passes that remaining prose to the implementation matcher.

## Expected

An artifact filename should not itself create an implementation obligation. Both examples should retain the same intent unless the actual instruction changes. A completed render should not need a gratuitous source edit or a false blanket no-edit statement to satisfy the classifier.

A focused fix could exclude filename/path tokens from verb-intent matching while retaining genuine instructions such as `Update the source file ...`. I have not patched the installed package.

## Related issues and scope

I checked #395, #648 and #1765, all closed. Those fixes concern prose such as “suggested fix”, scoped product/config prohibitions, and escaped read-only task wording respectively. This case is the remaining filename-token match in 0.65.0, without relying on a blanket prohibition.

Generated acceptance text was also relevant in the original investigation, but I am not presenting a separate reproduction for that here. Likewise, background/foreground intent-arbiter differences are not a second issue in this report. No claim is made that the latest upstream release has been live-tested or that every no-edits failure has this cause.

Local mitigation is truthful role/acceptance selection and verification of the requested artifact, not retrying successful work. This issue is only the reproducible classifier boundary problem.

nicobailon:
## Summary

Fixes #2039. The worker implementation matcher treated verbs inside artifact filenames (for example `daily-update.mp3`) as implementation requests because hyphens and slashes are word boundaries.

Path-like tokens are neutralized before implementation matching only when they contain an implementation verb. That keeps `Render … daily-update.mp3` from becoming implementation, while a real imperative whose object is a filename (`Fix package.json`, `Update package.json`) still wins. `taskMayMutate` / read-only acceptance is unchanged.

Rebased onto current `main` (includes #2073). Source and tests for this fix are unchanged from approved heads `b13ff1545b5d1e020b2916199e93768736f8211d` and `ef9e340b8fff626793c4300bd59b945dee09d0a3`.

Thanks to [@SiebertLanhove](https://github.com/SiebertLanhove) for #2039.

## Test plan

- [x] #2039 pair: `daily-briefing.mp3` and `daily-update.mp3` both classify as `unknown`
- [x] Later imperative after a filename still classifies as implementation
- [x] Filename-as-object cases: `Fix package.json`, `Update daily-update.mp3`, delegate `Update package.json`, `Review only; fix the package.json`
- [x] Changelog keeps current main Fixed lines (including #2073 / #2045) plus the #2039 entry once

<div><a href="https://cursor.com/agents/bc-a36bd68e-cee5-4a82-988e-f6a2f9c4d341?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-a36bd68e-cee5-4a82-988e-f6a2f9c4d341&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Document task-derived behavior labels for workflow launches in the built-in pi-subagents skill, including reviews and retained follow-ups.
 
 ### Fixed
+- Ignore verbs inside filenames and path-like tokens when classifying implementation intent, so artifact names such as `daily-update.mp3` do not create an implementation obligation. Thanks to [@SiebertLanhove](https://github.com/SiebertLanhove) for #2039.
 - Accept the boolean `fast` field on async recovery descriptors so a setting the writer persists can round-trip through follow-up. Thanks to [@isty2e](https://github.com/isty2e) for #2045.
 - Override `PI_PACKAGE_DIR` in detached runners with the detected npm Pi package root so npm runtime assets resolve from that package instead of an inherited host path. Thanks to [@alvarosevilla95](https://github.com/alvarosevilla95) for #2050.
 - Preserve explicit read-only task intent after host capability clamping, while still blocking implementation tasks that lack mutation tools. Thanks to [@stekman08](https://github.com/stekman08) for #2060.
--- a/src/runs/shared/task-intent.ts
+++ b/src/runs/shared/task-intent.ts
@@ -97,11 +97,6 @@ function stripSeverityCompounds(task: string): string {
 }
 export { stripSeverityCompounds };
 
-// Hyphens and slashes are word boundaries, so a token like daily-update.mp3
-// or src/add.ts would otherwise look like the verbs update/add. Neutralize
-// only tokens that contain those verbs; leave objects such as package.json
-// so "Fix package.json" stays an implementation obligation. Acceptance still
-// uses the raw write-verb vocabulary.
 const PATH_LIKE_TOKEN_PATTERN = /[^\s]+[/\\][^\s]+|[^\s/\\]+\.[A-Za-z][A-Za-z0-9]{0,9}\b/g;
 const PATH_INTERNAL_IMPLEMENTATION_VERB = /\b(?:implement|edit|modify|refactor|delete|update|add|remove|replace|create)\b/i;
 
--- a/src/runs/shared/task-intent.ts
+++ b/src/runs/shared/task-intent.ts
@@ -7,6 +7,7 @@
  * - `classifyTaskMutationIntent` / `expectsImplementationMutation`: does the
  *   task REQUIRE file changes? Consumed by the completion mutation guard,
  *   which blocks completion, so its vocabulary is deliberately narrow.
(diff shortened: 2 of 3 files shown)


nicobailon:
Challenge is running on exact head `dcbe89542a0dec7e89aa3e57ed8a85f0f6d24f25`. Do not start a second writer for #2039.

nicobailon:
Fresh review approved exact head `ef9e340b8fff626793c4300bd59b945dee09d0a3`. Queued behind #2068. Do not land this SHA until that land finishes and this head is still current. Credit stays with @SiebertLanhove.

PR #2089, opened 2026-09-09
fix: preserve active theme across subagent rendering

kubahasek:
## Summary
- preserve the parent’s active theme when foreground children share its process
- initialize themes normally in detached/headless runner processes
- rerender slash results after light/dark appearance changes

## What this fixes
Starting a foreground subagent could switch parent tool cards to the wrong adaptive theme. Skipping initialization entirely then caused headless child extensions to fail with `Theme not initialized`. The child factory now initializes only when the current process has no shared theme.

## Validation
- focused in-process child tests: 24 passed
- typecheck passed
- `git diff --check` passed

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -336,14 +336,18 @@ function createSlashResultComponent(
 	theme: ExtensionContext["ui"]["theme"],
 	rendererConfig?: MainWindowRendererConfig,
 	foregroundDetachShortcut?: string,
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks, [@kubahasek](https://github.com/kubahasek). Your theme fix landed through #2091 with your original commits/authorship retained and explicit changelog credit. The replacement integrated current main and passed the required exact-head checks; post-merge main CI is now green. Closing this original PR as superseded by that landed integration. Thank you for fixing foreground theme preservation and the detached initialization path.

PR #2101, opened 2026-09-09
fix(workflow): reject static fanout budget mismatch

ton77v:
## Summary

- statically count unconditional literal `runs.run` and `runs.all` launch keys, including resumed children
- reject a provable `maxSubagentSpawnsPerRun` mismatch before agent discovery, artifact creation, or child launch
- return advisory warnings for dynamic or conditional launch counts while preserving runtime fan-out enforcement
- apply the same effective request/config budget to offline `action: "validate"`

## Tracking

Related: https://github.com/ton77v/dotfiles/issues/271

The reported workflow declared five conceptual lanes but contained six concrete child launches. Runtime enforcement caught the sixth only after the first five had completed. This change reports every statically proven key plus `minimum required: 6; configured: 5` before any expensive work begins.

## Validation

- `npm run typecheck`
- focused unit tests: 3 passed
- focused integration tests: 4 passed
- static validation benchmark: 5,000 representative six-launch scripts averaged 0.2662 ms each
- full unit and integration suites were also run locally on Windows; existing package-agent precedence and async fixture/watchdog failures remain. The directly affected legacy fan-out assertion was updated to the new fail-fast contract and passes.
- post-push touched-file suites: `scripted-workflow.test.ts` 120 passed; `single-execution.part-1.test.ts` 93 passed, 7 platform skips.

<signature>🦀 **sent by [Agent System v0.20.2](https://antoniov.online/agentic-systems): pi | gpt-5.6-sol**</signature>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Document task-derived behavior labels for workflow launches in the built-in pi-subagents skill, including reviews and retained follow-ups.
 
 ### Fixed
+- Reject workflow scripts whose statically provable child launches exceed `maxSubagentSpawnsPerRun` before discovery, artifact creation, or child launch; dynamic launch counts remain advisory and retain runtime enforcement.
 - Restore background SDK sessions for the official Pi 0.85.1 Linux x64 standalone release, without a separate SDK install; npm runners keep their package-root and lifecycle behavior. Thanks to [@xz-dev](https://github.com/xz-dev) for #2049.
 - Preserve the parent's active theme when foreground children start, while initializing themes in detached runners and refreshing slash-result rendering. Thanks to [@kubahasek](https://github.com/kubahasek) for #2089.
 - Advance live workflow and lane elapsed times from their starts, and nest loaded workflow children in the async widget instead of repeating lane rows and sibling cards. Partial fix for #2085; thanks to [@expoli](https://github.com/expoli).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@
 - Document task-derived behavior labels for workflow launches in the built-in pi-subagents skill, including reviews and retained follow-ups.
 
 ### Fixed
-- Reject workflow scripts whose statically provable child launches exceed `maxSubagentSpawnsPerRun` before discovery, artifact creation, or child launch; dynamic launch counts remain advisory and retain runtime enforcement.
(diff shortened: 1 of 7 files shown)


nicobailon (changes requested):
Thanks @ton77v. Focused tests/typecheck pass, but the analyzer has false hard rejections. Reviewed 20dc9c573 and verified the analyzer/executor/public admission files are unchanged in your main-merge head 04a68e464.

At budget 1, these valid scripts return done with zero actual runWorkflowScript launches, while validation claims minimum 2:
```js
if (true) return "done";
await runs.run("a", {}); await runs.run("b", {});
```
```js
try { return "done"; } catch (e) {
  await runs.run("a", {}); await runs.run("b", {});
}
```
Uninstantiated class instance fields and optional-chain arguments reproduce the same problem. Restrict proven traversal to supported syntax; uncertain/deferred flow should remain advisory rather than become a general control-flow analyzer.

Additional reproduced cases:
- A block-local `const runs = {run(key){return key;}}` is mistaken for the workflow API. Shadowing/reassignment must invalidate proof.
- Later spreads overwriting literal runs.all keys can collapse two apparent keys into one actual reused key. Unknown later properties must invalidate the earlier literal key proof.
- Public `action:"validate"` with `maxSubagentSpawnsPerRun` is rejected by public-execution.ts before the new override handling. The existing guard needs narrow integration for the advertised public request/config consistency, with a public-boundary regression.

Runtime enforcement remains intact, including dynamic batch over-budget admission. Please preserve that safety net and make the static minimum conservative. Keeping your branch untouched for corrections.

nicobailon (approved):
Thanks @ton77v. Verified all four correction classes on 2eaa1bd3: the exact prior worker/public counterexamples now behave conservatively, effective offline request budgets work, runtime fanout enforcement remains intact, and contributor credit is present. Focused unit/public/integration checks and typecheck pass. Required exact-head CI and restored main health still gate merge.

ton77v:
Thanks for the concrete repros. Addressed in `2eaa1bd3`:

- restricted hard proof to reachable top-level direct `runs.run` / `runs.all` forms;
- made returns, throws, mutually exclusive control flow, deferred class fields, optional-chain arguments, `runs` shadowing/reassignment, and unknown object properties/spreads advisory;
- allowed `action: "validate"` to accept `maxSubagentSpawnsPerRun` at the public boundary, with unit and public integration coverage;
- added the required contributor/PR changelog credit.

Runtime fan-out enforcement is unchanged. Post-fix verification passed typecheck, 132 touched unit tests, and the full touched integration file (93 passed, 7 Windows platform skips). Greptile's rerun is 5/5. Please re-review when convenient.

<signature>🦀 **sent by [Agent System v0.20.2](https://antoniov.online/agentic-systems): pi | gpt-5.6-sol**</signature>

PR #2119, opened 2026-09-09
fix: recover transiently excluded models

VladimirGVP:
## Summary
- re-probe one configured model when every candidate is excluded only by transient transport/provider failures
- bound recovery with an exactly-one cross-process append-only claim and targeted exclusion clearing
- keep auth, quota, rate-limit, request-shape, explicit-model, configuration, and model-scope errors fail-closed

## Validation
- `npm run typecheck`
- model exclusion/fallback unit tests: 145 passed
- `single-execution.part-2`: 198 passed
- full suite: 3075 passed; the same 20 failures as pristine base, with no patch-only failures
- independent reviewer: OK with notes
- final techlead: ACCEPT-WITH-WARNINGS
- live configured launch after installation/reload reached `openai-codex/gpt-5.6-sol` through the new runtime

## Notes
The remaining warnings are non-blocking: bounded synchronous mutation-lock contention and two additional coverage opportunities.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Document task-derived behavior labels for workflow launches in the built-in pi-subagents skill, including reviews and retained follow-ups.
 
 ### Fixed
+- Re-probe one cached model after a transient transport/provider outage only when every launch candidate is excluded for a recoverable availability reason. Foreground and detached runners use bounded cross-process ownership, clear only a successful probe's matching exclusion, and keep auth, quota, disabled, unknown-model, request, explicit-model, and model-scope failures fail-closed.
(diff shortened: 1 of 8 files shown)


nicobailon:
Initial exact-head review on `6274036e`: VISION **fits**, but disposition is **replace** — the diff is large (+672/−66) and the persistence/claim framework exceeds the narrow transient-recovery invariant. Also CONFLICTING with current `main`.

A smaller maintainer replacement from current `main` is being prepared. This PR stays open until that replacement lands.

Thanks to [@VladimirGVP](https://github.com/VladimirGVP) for #2119.

nicobailon:
Closing without merge.

Thanks to [@VladimirGVP](https://github.com/VladimirGVP) for the report and the PR. We're pausing new feature intake for a bit, and this diff is larger than we want to take on right now (+672/−66 with a heavy claim/persistence layer).

A smaller recovery path may still land separately via #2120; this PR itself will not.

Issue #2117, opened 2026-09-09
Final-stop drain can still abort a renewed steer turn when Pi reaches turn_start after the grace period


nicobailon:
## Summary

The final-stop grace drain in both native child hosts can still abort a renewed turn that Pi starts to consume a queued steer or follow-up. #2028 cancels the drain on `turn_start`, which fixes the common case, but Pi emits `turn_start` only after `turn_end` hooks, the steering drain, and `prepareNextTurn` have completed. If that gap exceeds `FINAL_STOP_GRACE_MS` (1000 ms), for example because of slow child hooks or compaction ahead of the renewed turn, the timer fires, sets `forcedTermination`, and calls `session.abort()`. The renewed turn then ends with `stopReason: "aborted"` and `errorMessage: "Request aborted"`, and the step fails despite committed final output.

Split out of #2057, where this timer was the abort source on pre-#2028 builds.

## Where

- `src/runs/background/run-child-session.ts`: `startFinalDrain` and the `FINAL_STOP_GRACE_MS` timer.
- `src/runs/foreground/execution.ts`: the equivalent drain in the foreground host.
- Pi 0.85.1 agent loop: on the renewed turn it awaits `prepareNextTurn` and then emits `turn_start`, then appends the queued user message, then issues the provider request.

## Expected

The drain never aborts a session that still has queued steering or follow-up input, regardless of how long Pi takes to reach `turn_start`.

## Suggested direction

- In the drain timer callback, check `session.agent.hasQueuedMessages()` (already used in `src/runs/shared/readonly-session-evidence.ts`) and re-arm the timer instead of aborting when input is queued. Optionally skip arming altogether when a queued message is present at the terminal stop.
- Keep the existing stop, timeout, and interrupt behavior unchanged; only the post-final-stop cleanup abort is affected.
- Add a regression in both hosts where a steer arrives after the scripted final message and `turn_start` is delayed past 1000 ms, asserting the session is not aborted and the continuation output is kept. This needs the fake child session to simulate Pi's boundary drain rather than only recording steers.

nicobailon:
Fixed by https://github.com/nicobailon/pi-subagents/pull/2118 (`808387a2`). Credit [@yanqianglu](https://github.com/yanqianglu) for the original #2057 report and direction that led to this slice.

Issue #2057, opened 2026-09-08
Async steer racing child completion marks result-bearing run failed


yanqianglu:
## Summary

A default `steer` request delivered while an async child is finalizing can leave the run classified as `failed`, even after the child emitted its final answer, settled, and the runner exited successfully.

This makes terminal status unreliable for orchestrators: `failed` can mean either “no usable result” or “a control-plane abort raced committed output.”

## Expected

Once final output has been committed, a steering-induced abort should not overwrite the terminal result as an ordinary failure.

I am not prescribing one exact lifecycle state. Any of these would be understandable:

- preserve `complete` and record the steering as delivered/late;
- report the steering as `missed` without changing terminal status;
- use a distinct terminal state such as `cancelled_after_result`;
- or consistently discard the output and report a non-failure cancellation state.

The important invariant is that committed output and terminal status agree.

## Actual

The steering request is accepted and delivered. The child produces output and settles. The run is then recorded as:

```text
status: failed
error: Request aborted
```

The runner process exits `0`, and the run’s output log contains the full final response and acceptance report.

Observed event sequence:

```text
subagent.steer.delivered
agent_end
agent_settled
subagent.step.failed  # Request aborted
subagent.run.completed # status=failed
```

The timestamps showed `agent_settled` followed by `subagent.step.failed` about 128ms later.

## Environment

- `pi-subagents`: 0.66.0
- Pi: 0.84.4
- Node: 22.23.2
- OS: macOS

## Why this matters

Async orchestration relies on terminal state to decide whether to consume output, resume, retry, or alert a user. A result-bearing run marked `failed` causes misleading notifications and can trigger unnecessary duplicate work.

## Workaround

Use `mode: "follow_up"` for non-urgent guidance and `resume` after completion. That avoids the race in practice, but does not remove the terminal-accounting hazard for the documented default `steer` behavior.

## Suggested regression test

Simulate an async child that emits final output while a default steering request is in flight. Assert that committed output and terminal status remain consistent, and that a steering-induced abort cannot turn a result-bearing settled run into `failed`.

nicobailon:
Thanks @yanqianglu. I traced the settlement path and compared steering in Pi 0.84.4 and current 0.85.1. Native steering queues input rather than directly aborting it. A bounded native-core probe at assistant completion and agent completion produced no abort, and 19 related integration tests pass; this does not reproduce your full async setup or rule out the reported race.

[...]

PR #2121, opened 2026-09-09
fix: report follow-up and steer consumption at run end

### LINKED ISSUE #2116: Follow-up steers are always reported failed at run end even when the child consumed them
## Summary

Follow-up mode steers on async runs are never marked delivered, so every one of them ends the run with `subagent.steer.failed` and the reason `run ended before queued follow-up delivery`, even when Pi consumed the follow-up and the child answered it. Steer mode has the mirror inaccuracy: `delivered` means Pi accepted the message into its queue, not that a turn consumed it.

Split out of #2057, where the reporter showed the session file contradicting the steer failure reason.

## Current behavior

- `registerSteer` in `src/runs/background/run-child-session.ts` returns `{ state: "queued" }` for `follow_up` and `{ state: "delivered" }` for `steer` immediately after `created.followUp()` / `created.steer()` resolve.
- Nothing later transitions a `queued` target. `applySteerDelivery` is the only writer and it runs once per request.
- At run end, `src/runs/background/subagent-runner.ts` marks every target still in `queued` state as failed with `run ended before queued follow-up delivery`.
- Pi 0.85.1 drains the steering queue and then the follow-up queue at the end of any turn without tool calls, emits the queued item as a user `message_start` / `message_end`, and runs a renewed turn before `agent_end`. That consumption is observable from the child event stream but is not observed today.
- The existing test `delivers inbox steer requests to the background child session` in `test/integration/async-execution.part-2.test.ts` only asserts the mid-run `subagent.steer.queued` event and never checks the run-end event.

## Expected

- A follow-up that Pi appended as a user message and processed reports `delivered` (or a distinct `consumed` state) rather than `failed`.
- A steer or follow-up that was accepted but never consumed before the child settled reports `failed` with a precise reason, for example `child completed before consuming steering`.
- The run-end `subagent.steer.failed` reason stays accurate in both cases.

## Suggested direction

- In `run-child-session.ts`, match user `message_end` events against outstanding accepted steers by comparing text with `formatSteerMessage(request)` and report consumption to the runner. The `StepSteerHandler` contract returns a single promise, so this needs either a second callback or matching in the runner's `onChildEvent` path.
- At settle time, use `session.agent.hasQueuedMessages()` (already used in `src/runs/shared/readonly-session-evidence.ts`) to fail unconsumed steers with a specific reason.
- Extend `test/support/fake-child-session.ts` so a steer or follow-up received after the scripted final message emits the user message and a continuation step, mirroring Pi's boundary drain. Add a regression that asserts the run-end steering state and events for both modes.

nicobailon:
Follow-up (and steer) consumption is now reported from the child event stream at run end. API acceptance is not treated as consumption.

Closes #2116.

## Invariant

Consumed steering requests report `delivered` because the child emitted a matching user `message_end`. Accepted-but-unconsumed requests fail at settlement with a **request-specific** reason (`child completed before consuming steering` vs `child completed before consuming follow-up`). `hasQueuedMessages()` is not used to pick or prove a particular request's reason.

## What changed

- Track each accepted steer/follow-up per live child with its `formatSteerMessage` text.
- Match user `message_end` content to the oldest equal accepted request and transition that target exactly once to `delivered`.
- At child settlement, fail only still-unconsumed requests, each with a mode-specific reason. Aggregate queue state is not copied onto every target.
- Reuse the #2118 fake-child-session boundary drain; emit the consumed user `message_end` so tests can assert consumption.
- Cover both modes, equal-text duplicates, consumed and unconsumed endings, mixed modes while queued input remains, and terminal events.
- Parent `steerAsyncRun` recovery treats `queued` or `delivered` as child acceptance on one shared abort path (pre-recovery and in-claim). A late API accept no longer pauses a live single child. Recovery still runs only when the child never acked (stays routed).

## Out of scope

- Public API and persistence-schema changes
- Final-stop drain latch work from #2117 / #2118
- Broader foreground-host work beyond the queued-acceptance recovery gate

Thanks to [@yanqianglu](https://github.com/yanqianglu) for #2057 → #2116.

<div><a href="https://cursor.com/agents/bc-67b30968-755c-4a24-8efd-41869477ef23?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-67b30968-755c-4a24-8efd-41869477ef23&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


nicobailon:
Maintainer finishing pass at 2efaa72d3e1c0a3664f56b8c89132157060e595b: fresh independent review found no production defect or reason for a rewrite. Fixed the new held-queue test fixture's missing live handle, aligned async/foreground receipt docs, and added changelog credit for [@yanqianglu](https://github.com/yanqianglu). Focused steering integration 3/3, unit tests 27/27, typecheck and diff hygiene pass. Production source is unchanged from the reviewed candidate after verified mechanical integration with main. Exact-head CI and Greptile remain the final landing gates.

nicobailon:
Fixed the transport-acceptance finding in 1f61ec4d7bd36d84af343cf705e57576ad565732 by removing the premature queued callback. Request correlation remains registered before the transport call; rejected transport now returns its error without a successful queued acknowledgement. Focused steering units 28/28, integration 3/3, typecheck and diff hygiene pass. The source fix is one deleted line. No broader rewrite or API change.

PR #2137, opened 2026-09-10
fix: detect host builtins by name so extension overrides count

### LINKED ISSUE #2135: 0.67.0: host builtin-tool discovery breaks when an extension re-registers builtin tools (e.g. pi-tool-display) — all child tools pruned to  [contact_supervisor]
Summary

 The new host tool-availability intersection introduced in 0.67.0 (getHostBuiltinToolNames + hostAvailableBuiltins filtering in
 src/runs/shared/child-tool-plan.ts) only accepts tools whose sourceInfo.source === "builtin". When another extension wraps/re-registers the builtin
 tools — pi-tool-display (https://www.npmjs.com/package/pi-tool-display) does exactly this, so read/bash/edit/etc. come back from pi.getAllTools() with
 source: "npm:pi-tool-display" — the computed host builtin set ends up empty (or nearly empty), and every subagent launch has its entire tool allowlist
 pruned down to [contact_supervisor].

 Environment

 - pi-subagents: 0.67.0 (works fine on 0.66.0)
 - pi: 0.85.1
 - pi-tool-display: 0.5.0 (re-registers builtin tools to customize TUI rendering)
 - macOS

 Actual behavior

 Every child launch fails or is degraded. Review/scout lanes fail closed:

 Agent 'reviewer': tool contract could not be satisfied; host runtime does not
 provide permitted required repository tools [read, grep, find, ls].
 Requested tool names: [read, grep, find, ls, contact_supervisor];
 effective tool allowlist: [contact_supervisor].

 Worker launches fail with:

 Agent 'worker' was given an implementation task, but its tool allowlist has
 no mutation-capable tools.

 Children that do launch report having only contact_supervisor.

 Evidence

 Dumping pi.getAllTools() in the host session shows the builtin tools are present but carry the wrapping extension's source:

  json
 { "name": "read",  "source": "npm:pi-tool-display" },
 { "name": "bash",  "source": "npm:pi-tool-display" },
 { "name": "edit",  "source": "npm:pi-tool-display" },
 { "name": "powershell", "source": "builtin" }   // not wrapped; only survivor

 getHostBuiltinToolNames() (child-tool-plan.ts) keeps only source === "builtin" || (source === "auto" && PI_BUILTIN_TOOL_NAMES.has(name)), so
 read/bash/edit/write/grep/find/ls are all dropped, hostAvailableBuiltins collapses, and the intersection in resolvePiLaunchToolPlan removes every
 declared tool except NATIVE_COORDINATION_TOOL_NAMES.

 Note that pi core intentionally lets extension tools with the same name override builtin entries in _toolDefinitions (agent-session.js), so "a
 builtin-named tool with an extension source" is a legitimate, supported state — the discovery logic just doesn't account for it.

 Expected behavior

 Host builtin-tool discovery should recognize host-provided tools regardless of whether an extension has wrapped them, e.g. treat a tool as
 host-available when PI_BUILTIN_TOOL_NAMES.has(tool.name) regardless of sourceInfo.source, or union with pi.getActiveTools()/a name-based check.

 Suggested fix

  ts
 // getHostBuiltinToolNames: accept builtin *names* from any source,
 // since extensions may legitimately re-register builtin tools.
 .filter((tool) => {
     const source = tool.sourceInfo?.source;
     return source === "builtin"
         || source === "auto"
         || PI_BUILTIN_TOOL_NAMES.has(tool.name);
 })

 (Failing safe the other way — skip the intersection when the computed set contains none of PI_BUILTIN_TOOL_NAMES — would also prevent this class of
 outage.)

 Workaround

 Pin pi-subagents@0.66.0, or uninstall the tool-wrapping extension.

carlesba:
Every child launch lost its entire tool allowlist down to `contact_supervisor` on any host where an extension re-registers Pi's builtin tools, so review and scout lanes failed closed and agents given implementation wording failed for want of mutation-capable tools.

Host discovery treated `sourceInfo.source === "builtin"` as proof that the host provides a tool, but overriding a builtin by registering a tool under the same name is documented Pi behaviour, and pi core replaces the registry entry so a wrapped `read` reports its wrapper as the source. Because `powershell` is not usually wrapped it survived as the lone builtin-sourced entry, which made discovery look successful and defeated the zero-results fail-safe that would otherwise have skipped the intersection entirely.

I deliberately did not take the wider `source === "auto"` form suggested in the issue, because an existing test pins non-builtin `auto` names out of the builtin-name set and the name check alone subsumes the previous `auto` case.

Fixes #2135

### DIFF
--- a/src/runs/shared/child-tool-plan.ts
+++ b/src/runs/shared/child-tool-plan.ts
@@ -341,6 +341,11 @@ export function resolvePermissionSystemExtension(): string | undefined {
  * `hostAvailableBuiltins` to `resolvePiLaunchToolPlan` so child tool plans
  * intersect declared agent tools with what the host actually supports.
  *
+ * Extensions may override a builtin by registering a tool under the same name
+ * (pi's docs/extensions.md), which replaces the registry entry and its
+ * `builtin` provenance, so a builtin name counts as host-provided whatever
+ * source reports it.
+ *
  * Returns `undefined` when builtin tool discovery fails or yields nothing,
  * so callers skip the intersection (fail-safe to allowing all declared tools).
  * This handles test mocks without proper tool registration and hosts whose
@@ -352,7 +357,7 @@ export function getHostBuiltinToolNames(pi: Pick<ExtensionAPI, "getAllTools">): 
 			.getAllTools()
 			.filter((tool) => {
 				const source = (tool.sourceInfo as { source?: string } | undefined)?.source;
-				return source === "builtin" || (source === "auto" && PI_BUILTIN_TOOL_NAMES.has(tool.name));
+				return source === "builtin" || PI_BUILTIN_TOOL_NAMES.has(tool.name);
 			})
 			.map((tool) => tool.name);
 		return builtins.length > 0 ? builtins : undefined;
--- a/test/unit/child-tool-plan.test.ts
+++ b/test/unit/child-tool-plan.test.ts
@@ -217,6 +217,45 @@ describe("production launch path supplies hostAvailableBuiltins", () => {
 		assert.deepEqual(builtins, ["read", "bash"]);
 	});
 
+	it("getHostBuiltinToolNames keeps builtin names an extension override re-registered", () => {
+		const wrappedPi = {
+			getAllTools: () => [
+				...["read", "bash", "edit", "write", "grep", "find", "ls"].map((name) => ({ name, sourceInfo: { source: "npm:pi-tool-display" } })),
+				{ name: "powershell", sourceInfo: { source: "builtin" } },
+				{ name: "custom-tool", sourceInfo: { source: "npm:pi-tool-display" } },
+			],
+		};
+		const builtins = getHostBuiltinToolNames(wrappedPi);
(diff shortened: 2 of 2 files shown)


IdrisGit:
Hit this too on Linux with the same versions. Tried swapping the extension order—didn’t help. The tools were there and working, but discovery only picked up `powershell`.

Tested the same change locally and it fixes the tool plan without dropping exclusions or capability checks. Haven’t tested a full child run yet, but this matches what I found.

clementprevot:
We kind of opened the PR at the same time @carlesba: https://github.com/nicobailon/pi-subagents/pull/2138

I've taken a slightly larger approach in mine where you can also expose extension provided tools to the subagents instead of keeping only core one.
You can still configure the plan to filter out non core tools though.

carlesba:
feel free to close this one if the other PRs merge. I just wanted a quick fix to update my deps. It's annoying right now :)

nicobailon:
Thanks, @carlesba. PR #2143 landed the minimal host-tool fix as d9864f8288152e83270f62090d7d66eb5ff729bb and covers this regression while preserving your contribution in the commit co-author trailers and profile-linked CHANGELOG credit. Closing this PR as superseded by #2143.

carlesba:
Thank @nicobailon, when should we expect the patch release?

Issue #2136, opened 2026-09-10
0.67.0 host-tool intersection (#2034) strips extension-overridden builtins (bash) from child allowlists; review/scout lanes now fail closed


reevonr:
# 0.67.0 host-tool intersection (#2034) strips extension-overridden builtins (bash) from child allowlists; review/scout lanes now fail closed

## Environment

- pi-subagents **0.67.0** (regression — worked on 0.66.0)
- pi 0.85.1, macOS
- Conflicting extension: `pi-patty-bg-tasks@1.1.6`, which re-registers the builtin shell tool via `pi.registerTool({ name: "bash", ... })` (documented as "bash (override): The built-in bash tool, with a survival instinct")

## Summary

Since 0.67.0, every child launch drops `bash` from the agent's declared tools when any installed extension **overrides** a builtin tool by re-registering it under the same name. Plain agents (worker, architect, …) silently lose bash — children truthfully report "there is no `bash` tool available to me" while the parent still sees bash in its own tool list. Review/scout lane launches now fail closed at startup:

```
Agent 'scout': tool contract could not be satisfied; host runtime does not provide
permitted required repository tools [bash]. Requested tool names: [read, grep, find, ls,
bash, write, ...]; effective tool allowlist: [read, grep, find, ls, write,
contact_supervisor, ...]. This is a lane infrastructure failure, not a completed
review/scout result.
```

## Root cause

`getHostBuiltinToolNames()` (`src/runs/shared/child-tool-plan.ts`) classifies host tools as available only when:

```ts
const source = (tool.sourceInfo as { source?: string } | undefined)?.source;
return source === "builtin" || (source === "auto" && PI_BUILTIN_TOOL_NAMES.has(tool.name));
```

An extension that overrides a builtin (the patty pattern, and the pattern #2059 already acknowledges for workflow children: *"extensions wrap Pi built-ins"*) produces a tool whose `sourceInfo.source` is `"extension"`. The new host intersection from #2034 then treats `bash` as host-unavailable and strips it from the child plan — even though the child would have no problem dispatching it, because **children inherit ambient extensions by default** and therefore load the same override.

So the source check is stricter than what actually matters (dispatchability in the child): the `PI_BUILTIN_TOOL_NAMES` name-level test already exists for `source === "auto"`, but the `"extension"` source never gets that chance.

## Reproduction

1. Install an extension that re-registers `bash` (e.g. `pi-patty-bg-tasks`).
[...]

Issue #2034, opened 2026-09-06
Child spawn fails on Prime Agent: agent `tools:` is passed through without checking the child host


BioInfo:
## What happens

`pi-subagents` loads and registers fine under Prime Agent (PrimeIntellect's pi fork). Then every shipped agent dies at child spawn:

```
Error: Unknown built-in tool(s): read, grep, find, ls.
Available built-in tools: ipython
```

2 of 2 attempts on `oracle`, `async: false`, failing before the child model is called. The artifact reads `Subagent run failed before producing output.`

## Why

Prime Agent swapped pi's built-in tool menu for one persistent IPython kernel. Its CLI parser holds a `REMOVED_BUILTIN_TOOL_NAMES` set - `read`, `write`, `grep`, `find`, `ls` - and errors when any of them shows up on `--tools`.

`resolveChildToolPlan` in `src/runs/shared/child-tool-plan.ts` (0.66.0, L313-327) builds `declaredBuiltinTools` from the agent frontmatter and filters it against `excludedToolSet` only. Nothing intersects it with what the child runtime provides, so `agents/oracle.md`'s `tools: read, grep, find, ls, bash` goes straight onto the child's `--tools` line. Same for `scout`, `worker`, `reviewer`, `delegate` and `researcher`.

## Two blockers sit in front of this one

Both fixable, and each looks like the whole problem until you clear it.

**1. Peer dep missing.** Prime Agent bundles `pi-agent-core`, `pi-ai` and `pi-tui`, not `pi-coding-agent`. `src/extension/index.ts` imports `keyText` from it as a value, so the factory throws `MODULE_NOT_FOUND`. Prime Agent swallows that and starts normally - exit 0, no warning, no tool. Install the package and `subagent`, `subagent_wait` and `subagent_supervisor` all register.

**2. Wrong child binary.** `pi-spawn.ts` resolves the child from the `pi-coding-agent` package root, so fixing (1) hands it a real pi to launch instead of the host agent. `PI_SUBAGENT_PI_BINARY=$(command -v prime-agent)` redirects it, and that turned a seven-minute silent hang into the error above.

## The ask

Intersect an agent's declared `tools:` with what the child runtime provides. Prime Agent's set is `["ipython"]`. `capabilityCeiling.allowedTools` already does that intersection for another purpose, so the only missing piece is the host's own list.

[...]

PR #2138, opened 2026-09-10
fix: keep host-discovered tool names that extensions provide

### LINKED ISSUE #2132: 0.67.0: npm-installed builtin tool wrappers are dropped by host discovery, blocking scout launches
## Summary

On Pi 0.85.1 with pi-subagents 0.67.0 and npm-installed pi-tool-display 0.5.0, native `scout` launches fail before execution because standard repository tools are classified as unavailable, although the parent can call them.

The installed `getHostBuiltinToolNames()` implementation accepts `builtin` and `auto` sources, but not package sources such as `npm:pi-tool-display`. This appears to be an uncovered package-install case of #2061 / #2059, exposed by the fail-closed check in #2114 / #2058.

## Environment

- Windows
- Pi Coding Agent: 0.85.1
- pi-subagents: 0.67.0 (matches npm latest at reporting time)
- pi-tool-display: 0.5.0
- Both extensions installed through npm packages in Pi settings
- No pi-tool-display config file; default tool overrides apply

Relevant settings package entries:

```json
{
  "packages": [
    "npm:pi-tool-display",
    "npm:pi-subagents"
  ]
}
```

These are the relevant entries from an existing installation, not a claim that a clean two-package installation has been tested.

## Observed failure

A read-only smoke launch:

```js
subagent({
  agent: "scout",
  task: "Use ls to list the current directory. Do not modify files.",
  async: true
})
```

Repeatedly returns:

```text
Agent 'scout': tool contract could not be satisfied; host runtime does not provide permitted required repository tools [read, grep, find, ls, bash]. Requested tool names: [read, grep, find, ls, bash, write, contact_supervisor]; effective tool allowlist: [contact_supervisor]. This is a lane infrastructure failure, not a completed review/scout result.
```

With a previous explicit agent tool list using `powershell` instead of `bash`, the effective allowlist was `[powershell, contact_supervisor]`; read/grep/find/ls were still missing. Removing both `defaultTools` and agent-specific `tools` overrides did not resolve the failure. The error changed to the default bash-based list above, confirming the updated agent configuration was read.

The parent successfully called `read`. Doctor reported async support and filesystem checks as available/OK. The errors are recorded as parent-session tool results; the child never starts.

## Source trace

Installed pi-subagents `src/runs/shared/child-tool-plan.ts:349–358`:

```ts
export function getHostBuiltinToolNames(pi: Pick<ExtensionAPI, "getAllTools">): string[] | undefined {
  try {
    const builtins = pi
      .getAllTools()
      .filter((tool) => {
        const source = (tool.sourceInfo as { source?: string } | undefined)?.source;
        return source === "builtin" || (source === "auto" && PI_BUILTIN_TOOL_NAMES.has(tool.name));
      })
      .map((tool) => tool.name);
    return builtins.length > 0 ? builtins : undefined;
  } catch {
    return undefined;
  }
}
```

Relevant installed-source behavior:

1. pi-tool-display `src/types.ts:70–78` defaults read/grep/find/ls/bash/edit/write overrides to true.
2. `src/tool-overrides.ts` registers same-name tools through `pi.registerTool`, delegating execution to built-in implementations.
3. Pi `dist/core/package-manager.js:989` uses `{ source: sourceStr, scope, origin: "package" }` for package resources. For this package, `sourceStr` is `npm:pi-tool-display`.
4. Pi `dist/core/resource-loader.js:567–576` assigns extension source metadata to registered tools.
5. Pi `dist/core/agent-session.js:2128–2133` replaces same-name built-in definitions and source metadata with registered tool definitions. `getAllTools()` returns the resulting source metadata.
6. pi-subagents filters out these package-sourced standard names, then intersects requested tools against the reduced host set (`child-tool-plan.ts:407–412`). The scout required-tool check rejects the launch.

The `auto` branch added for auto-discovered wrappers does not cover npm-installed wrappers.

## Function-level reproduction (verified)

I extracted the actual installed `getHostBuiltinToolNames()` function and ran it with synthetic `getAllTools()` entries, changing only `sourceInfo.source`. Inputs were the seven standard names plus a builtin-sourced `powershell` entry:

| Source of seven standard tools | Detected tools |
| --- | --- |
| `builtin` | read, grep, find, ls, bash, edit, write, powershell |
| `auto` | read, grep, find, ls, bash, edit, write, powershell |
| `npm:pi-tool-display` | powershell |

The assertions passed using the installed function, not a rewritten approximation. The surviving builtin entry matters: it produces a nonempty host set, so the function does not take its `undefined` fallback.

## Expected behavior

Standard builtin slots wrapped by an npm-installed extension should not be classified as absent solely because their source is a package identifier. Preserve capability ceilings and explicit exclusions; no request to disable required-tool validation or grant arbitrary extension tools.

Please cover package-sourced builtin wrappers alongside the existing `auto` case, ideally with a regression that exercises discovery and the scout launch contract together.

## Verification limits

- Repeated real scout failures and successful parent read are observed.
- Installed source chain and function-level filtering reproduction were checked.
- Current live parent tool metadata was not directly dumped.
- A clean two-package installation and an end-to-end launch after disabling pi-tool-display have not yet been tested.
- No full session logs, credentials, private configuration, or user-specific filesystem paths are included.

### LINKED ISSUE #2133: v0.67.0 blocks reviewer/scout when host getAllTools() reports a partial builtin menu
## Summary

   I hit what looks like a false positive in the v0.67.0 reviewer/scout preflight.

   If another Pi extension wraps one of Pi’s built-in file tools by re-registering the same name, pi-subagents no longer considers that tool available. As a result, `reviewer` and `scout` fail before the child starts. The only reason I came across this is that I'm developing my own extension which wraps the core tools to run strict security and privacy gates when reading files and paths.

   ## Environment

   - Pi `0.85.1`
   - pi-subagents `0.67.0`
   - Plain `pi` invocation
   - No `defaultTools` setting
   - No capability ceiling reported

   Pinning pi-subagents to `v0.66.0` allows the reviewer to start again.

   ## Reproduction

   A minimal example is an extension that wraps `read` while retaining the original behavior:

   ```ts
   import { createReadTool, type ExtensionAPI } from "@earendil-works/pi-coding-agent";

   export default function (pi: ExtensionAPI) {
     const read = createReadTool(process.cwd());

     pi.registerTool({
       ...read,
       async execute(...args) {
         return read.execute(...args);
       },
     });
   }
 ```

 With that extension loaded, launching the builtin reviewer fails with:

 ```text
   Agent 'reviewer': tool contract could not be satisfied; host runtime does not provide permitted required repository tools [read, grep, find, ls].
   Requested tool names: [read, grep, find, ls, contact_supervisor]; effective tool allowlist: [contact_supervisor].
   This is a lane infrastructure failure, not a completed review/scout result.
 ```

 The new host-tool check appears to count only tools whose sourceInfo.source is builtin or auto. A supported extension wrapper changes the provenance to extension-sourced, so the wrapped builtin is treated as absent.

 Why this seems worth distinguishing

 I understand why pi-subagents should not assume that an arbitrary extension tool named read is a safe replacement for Pi’s builtin read.

 However, parent-side provenance and child tool availability are different questions. A policy extension may wrap a builtin intentionally, while only Pi or the host runtime can say what the child will actually receive.

 I do not think a name-based fallback would be safe. An explicit core-level “overrides builtin” identity, host capability attestation, or documented opt-in override would preserve the safety boundary while avoiding this false positive.

### LINKED ISSUE #2134: v0.67.0 silently strips extension tools (pi-web-access) from researcher/evidence-auditor allowlists — run completes without web tools
## Summary

On v0.67.0, agents whose `tools:` allowlist names extension-registered tools — notably the builtin `researcher` and `evidence-auditor`, which declare `web_search`, `fetch_content`, `get_search_content`, `source_check` (pi-web-access) — launch successfully but **silently lose those tools**. The run completes (exit 0) with a degraded result instead of failing, because the new host-tool intersection (#2034) prunes any tool whose `sourceInfo.source` is not `builtin`/known-`auto`, and package extensions register their tools with source `npm:<package>`.

## Environment

- Linux x64
- Pi Coding Agent 0.85.1 (npm install, Node 24.15.0)
- pi-subagents 0.67.0
- pi-web-access 0.28.0 (ambient, loaded via settings `packages`)
- Models from a provider-extension package
- settings `packages`: `["npm:@aliou/pi-neuralwatt", "npm:pi-mcp-adapter", "npm:pi-subagents", "npm:pi-web-access"]`

## Observed failure

A standard async launch `subagent({ agent: "researcher", task: "..." })` completes, but the child runs without web tools. Child reply:

> I can't run this smoke test as specified: no `web_search` tool is available to me in this session (only file read/write and supervisor contact).

Runner stderr (`async-subagent-runs/<id>/runner.stderr.log`):

```
[pi-subagents] Agent 'researcher': host runtime tool availability omitted [web_search, fetch_content, get_search_content, source_check]. Requested tool names: [read, write, web_search, fetch_content, get_search_content, source_check, contact_supervisor]; effective tool allowlist: [read, write, contact_supervisor]. This is a non-fatal tool-plan diagnostic, not verification of the child's runtime tool menu.
```

## Control experiments

1. A custom agent with **no** tools allowlist, launched through the same detached runner, receives the full ambient toolset (`web_search`, `source_check`, `fetch_content`, `get_search_content`, MCP tools, …). Ambient extension loading and providers work; naming the tools in an agent allowlist is what triggers the drop.
2. A live host probe (`pi.getAllTools()` from a `session_start` hook) reports:

```
TOOL web_search | source=npm:pi-web-access origin=package
TOOL fetch_content | source=npm:pi-web-access origin=package
TOOL get_search_content | source=npm:pi-web-access origin=package
TOOL source_check | source=npm:pi-web-access origin=package
```

`getHostBuiltinToolNames()` (`src/runs/shared/child-tool-plan.ts`, v0.67.0) keeps only `source === "builtin"` or `source === "auto"` with a known Pi core name, so extension-registered tools are classified as unavailable host builtins and pruned from `effectiveToolAllowlist` — even though the detached runner then loads the ambient extensions that actually provide those tools.

## Relation to other reports

Same root family as #2132 and #2133, but a distinct and arguably worse variant:

- Those hit the reviewer/scout **fail-closed** check → the lane fails loudly.
- `researcher`/`evidence-auditor` are not on the fail-closed path (agent names don't match the review/scout pattern), so the tools are dropped with only a non-fatal warning. The run **succeeds** and returns a research brief produced without any web access, which can mislead users who don't check the runner stderr.

## Mitigation

Pinning 0.66.0 (`npm:pi-subagents@0.66.0` in settings packages) restores the full allowlist. Note the pin requires restarting pi so the loaded host extension and the on-disk runner match — the host-side runner spawn args changed between 0.66.0 and 0.67.0 (`runner-peer-preload.mjs` exists only in 0.67.0, so a 0.67.0 host with a 0.66.0 package on disk fails with `ERR_MODULE_NOT_FOUND`).

## Suggested directions

- Count tools registered by loaded package extensions as host-available (accept package-provenanced sources, or intersect against the ambient set the detached runner actually loads rather than the builtin-only set), or
- Route agents whose declared tools were dropped to the same fail-closed treatment as reviewer/scout, so a researcher without its research tools never completes "successfully".

### LINKED ISSUE #2135: 0.67.0: host builtin-tool discovery breaks when an extension re-registers builtin tools (e.g. pi-tool-display) — all child tools pruned to  [contact_supervisor]

clementprevot:
Fixes #2132, fixes #2133, fixes #2134, fixes #2135.

## Summary

In 0.67.0, `getHostBuiltinToolNames()` only trusted tools whose `sourceInfo.source` is `builtin` or `auto`. Extensions that re-register a core tool under the same name (e.g. "pi-tool-display" wraps bash/grep/find/ls/edit/write, "pi-hashline-edit-pro" wraps read) report the extension as the source, so the computed host set lost those names. The child tool plan then pruned every requested tool:

- reviewer and scout launches were refused outright ("tool contract could not be satisfied", #2132, #2133, #2135)
- other agents launched silently with a degraded allowlist, so children genuinely never got `read` (#2133)
- extension tool names declared in agent `tools:` lists (`mcp__*`, `web_search`, ...) were always dropped, because the intersection filtered every non-builtin name (#2134)
- the completion guard then rejected implementation-shaped tasks with "no mutation-capable tools", and tool-less children could answer without reading anything, inventing tool results

## Fix

Discovery now returns every registered tool name regardless of source, and the tool plan applies a two-tier rule in `resolvePiLaunchToolPlan`:

- core tool slots (`read`, `bash`, ...) resolve by name whatever their source. pi core ships them, so the name stays resolvable even when a launch denies extensions and the wrapper is absent
- extension tool names are only kept when the child loads the same ambient extension set. For example, `denyExtensions` or an explicit `extensions` list on the launch, keeps pruning them, so those launches still fais as expected.

## Testing

- new plan-level tests: shadowed core names are kept, extension names are kept when the child inherits ambient extensions and pruned under `denyExtensions` or `extensions: []`, and names absent from the host are still pruned
- new discovery tests: names are reported whatever their source (`builtin`, `auto`, package, `mcp`), and `undefined` is still returned on failure or when nothing is registered
- existing tests updated to the new discovery contract; the fail-closed paths (absent names, review-lane contract, `requireReadTool`) still hold

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Keep host-discovered tool names that extensions provide: core tool slots resolve by name whatever their source (extensions that re-register a core tool shadowed it with an extension source, so child tool plans pruned every requested tool and foreground scout/reviewer launches failed), and extension tool names survive on launches that load ambient extensions in the parent's project (runner launches). Foreground launches, cross-project launches, capability ceilings, and explicit extension lists still fail closed. (#2132, #2133, #2135)
+
 ## [0.67.0] - 2026-09-10
 
 ### Highlights
--- a/src/api/preflight.ts
+++ b/src/api/preflight.ts
@@ -88,7 +88,7 @@ export interface SubagentLaunchContractInput {
 	capabilityCeiling?: ResolvedSubagentCapabilityCeiling;
 	inheritedCapabilityCeiling?: ResolvedSubagentCapabilityCeiling;
 	/** Builtin tool names the host runtime provides; used to intersect agent-declared tools. */
-	hostAvailableBuiltins?: readonly string[];
+	hostToolNames?: readonly string[];
 	/** Per-launch bridge config; replaces the global `intercomBridge` config exactly as the tool and delegation overrides do. */
 	intercomBridge?: IntercomBridgeConfig;
 	/**
@@ -415,7 +415,7 @@ export async function resolveSubagentLaunchContract(input: SubagentLaunchContrac
 			capabilityCeiling: effectiveCapabilityCeiling,
 			agentName: agent.name,
 			permissionRules,
-			hostAvailableBuiltins: input.hostAvailableBuiltins,
+			hostToolNames: input.hostToolNames,
 		});
 	} catch (error) {
 		const message = error instanceof Error ? error.message : String(error);
--- a/src/api/preflight.ts
+++ b/src/api/preflight.ts
@@ -89,6 +89,8 @@ export interface SubagentLaunchContractInput {
 	inheritedCapabilityCeiling?: ResolvedSubagentCapabilityCeiling;
 	/** Builtin tool names the host runtime provides; used to intersect agent-declared tools. */
 	hostToolNames?: readonly string[];
+	/** Whether the launch loads ambient extensions (foreground parent-hosted launches do not). */
+	ambientExtensions?: boolean;
 	/** Per-launch bridge config; replaces the global `intercomBridge` config exactly as the tool and delegation overrides do. */
 	intercomBridge?: IntercomBridgeConfig;
 	/**
@@ -416,6 +418,7 @@ export async function resolveSubagentLaunchContract(input: SubagentLaunchContrac
 			agentName: agent.name,
 			permissionRules,
 			hostToolNames: input.hostToolNames,
+			ambientExtensions: input.ambientExtensions,
 		});
 	} catch (error) {
 		const message = error instanceof Error ? error.message : String(error);
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -14,7 +14,7 @@ import { createAtomicJsonWriter, writePrivateAtomicJson } from "../../shared/ato
 import { buildEffectiveSystemPrompt } from "../shared/effective-system-prompt.ts";
 import { currentCompletionOwnerId } from "../../shared/completion-owner.ts";
 import { planChildLaunch, resolveStepBehavior, suppressProgressForReadOnlyTask, type ResolvedStepBehavior } from "../shared/child-launch-plan.ts";
-import { applyThinkingSuffix, getHostBuiltinToolNames, projectLaunchResolvedChildExtensions, resolvePiLaunchToolPlan } from "../shared/child-tool-plan.ts";
+import { applyThinkingSuffix, getHostToolNames, projectLaunchResolvedChildExtensions, resolvePiLaunchToolPlan } from "../shared/child-tool-plan.ts";
 import { injectSingleOutputInstruction, normalizeSingleOutputOverride, resolveSingleOutputPath, validateFileOnlyOutputMode } from "../shared/single-output.ts";
 import { applyWatchdogLaunchRules, sendRuleViolationWarning } from "../../watchdog/rules.ts";
 import { buildChainInstructions, isDynamicParallelStep, isParallelStep, resolveExistingReadInstructionPaths, resolveExistingReadPaths, writeInitialProgressFile, type ChainStep, type SequentialStep, type StepOverrides } from "../../shared/settings.ts";
@@ -963,7 +963,7 @@ export function buildAsyncRunnerSteps(id: string, params: AsyncRunnerStepBuildPa
 		const launchRuleError = applyWatchdogLaunchRules({ cwd: stepCwd, agent: a.name, model: modelCandidates[0] ?? model, warn: (violation) => sendRuleViolationWarning(ctx.pi, violation) });
 		if (launchRuleError) throw new AsyncStartValidationError(launchRuleError);
 		const fast = s.fast ?? params.fast ?? a.fast;
-		const hostAvailableBuiltins = getHostBuiltinToolNames(ctx.pi);
+		const hostToolNames = getHostToolNames(ctx.pi);
 		const toolPlan = resolvePiLaunchToolPlan({
 			tools: a.tools,
 			excludeTools: a.excludeTools,
@@ -982,7 +982,7 @@ export function buildAsyncRunnerSteps(id: string, params: AsyncRunnerStepBuildPa
 			agentName: a.name,
 			permissionRules,
 			runtimeSnapshotHost: ctx.pi,
-			hostAvailableBuiltins,
+			hostToolNames,
 		});
 		const launchResolvedExtensions = externalRunner ? undefined : projectLaunchResolvedChildExtensions(toolPlan);
 		if (externalRunner && permissionRules) {
@@ -1379,7 +1379,7 @@ export function executeAsyncChain(
 				deadlineAt,
 				globalConcurrencyLimit: params.globalConcurrencyLimit,
 				runFanoutBudget,
-				hostAvailableBuiltins: getHostBuiltinToolNames(ctx.pi),
+				hostToolNames: getHostToolNames(ctx.pi),
 				workflowGraph,
 				...(params.parentWorkflowRunId ? { parentWorkflowRunId: params.parentWorkflowRunId } : {}),
 				...(params.workflowKey ? { workflowKey: params.workflowKey } : {}),
@@ -1752,7 +1752,7 @@ export function executeAsyncSingle(
 			return formatAsyncStartError("single", error instanceof Error ? error.message : String(error));
 		}
 	}
-	const hostAvailableBuiltins = getHostBuiltinToolNames(ctx.pi);
+	const hostToolNames = getHostToolNames(ctx.pi);
 	const toolPlan = resolvePiLaunchToolPlan({
 		tools: agentConfig.tools,
 		excludeTools: agentConfig.excludeTools,
@@ -1771,7 +1771,7 @@ export function executeAsyncSingle(
 		agentName: agentConfig.name,
 		permissionRules: resolvePermissionRules(ctx.permissions, agentConfig.permissions),
 		runtimeSnapshotHost: ctx.pi,
-		hostAvailableBuiltins,
+		hostToolNames,
 	});
 	const launchResolvedExtensions = externalRunner ? undefined : projectLaunchResolvedChildExtensions(toolPlan);
 	if (!externalRunner) {
@@ -1976,7 +1976,7 @@ export function executeAsyncSingle(
 				launchContractDigest,
 				launchResolvedExtensions,
 				runFanoutBudget,
-				hostAvailableBuiltins,
+				hostToolNames,
 				...(params.parentWorkflowRunId ? { parentWorkflowRunId: params.parentWorkflowRunId } : {}),
 				...(params.workflowKey ? { workflowKey: params.workflowKey } : {}),
 				...(lane ? { lane } : {}),
--- a/src/runs/background/async-execution.ts
+++ b/src/runs/background/async-execution.ts
@@ -785,6 +785,10 @@ export function buildAsyncRunnerSteps(id: string, params: AsyncRunnerStepBuildPa
 	const chainSkills = params.chainSkills ?? [];
 	const availableModels = params.availableModels;
 	const runnerCwd = resolveChildCwd(ctx.cwd, cwd);
+	// Ambient discovery in the runner happens in the runner cwd: only trust
+	// the parent registry snapshot for ambient names when the child stays in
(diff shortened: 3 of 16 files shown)


danieldekay:
Confirmed the regression scope from a fresh macOS reproduction (Pi 0.85.1, pi-subagents 0.67.0), and this fix targets the exact mechanism.

**Reproduction (macOS):** async workflow spawning the builtin `scout` and `reviewer` both fail at spawn:

```
Agent 'scout': tool contract could not be satisfied; host runtime does not provide permitted required repository tools [read, grep, find, ls, bash]. Requested tool names: [read, grep, find, ls, bash, write, contact_supervisor]; effective tool allowlist: [contact_supervisor].
```

Failures are pre-execution (63 ms / 29 ms), confirming it's the tool-plan intersection, not a child runtime issue.

**Additional data point beyond #2132/#2133/#2135/#2136:** the named wrapper extensions (pi-tool-display, pi-patty-bg-tasks, pi-hashline-edit-pro) are **not installed** in this environment — I searched all `pi.registerTool` call sites across `~/.pi/agent`. The regression therefore is not limited to those packages: any host whose `getAllTools()` reports a non-`builtin`/`auto` source for a core tool name (package-sourced wrappers, other override extensions) trips the fail-closed check in `#2114`.

**On the proposed fix:** accepting core slot names regardless of declared source matches reality — core names resolve by name in the child even when an extension layers over them (`#2134`). Worth keeping the two-tier behavior for extension names (`denyExtensions` / explicit `extensions` should still prune them) exactly as PR #2136's suggested rule and this PR describe; that preserves the #2034 safety intent for genuinely restricted hosts while restoring wrapper compatibility.

**Verification plan locally:** pinned `npm:pi-subagents@0.66.0` as an immediate workaround; will re-run the identical scout/reviewer smoke launch after config takes effect and can post whether 0.66.0 passes.

:+1: from this environment — looking forward to the release.

clementprevot:
OK it was becoming absurd, the more I fixed automated review comments, the more I had
I returned back to the basis of the fix and I think all upcoming automated review comment are for a follow-up

nicobailon:
Thanks, @clementprevot. PR #2143 landed the narrower host-tool invariant as d9864f8288152e83270f62090d7d66eb5ff729bb: core tool slots are intersected with host availability while declared non-core extension tools remain for child startup validation. It covers the reported regressions without the broader provider inference, and preserves your contribution in the commit co-author trailers and profile-linked CHANGELOG credit. Closing this PR as superseded by #2143.

PR #2141, opened 2026-09-10
feat: checkpointBeforeDeadlineMs — runner-issued checkpoint steer before the async deadline

freezscholte:
## Why

`timeoutMs` is a hard kill. `docs/tool-reference.md` already says an elapsed timeout is not a mutation-safe boundary and asks the orchestrator to `steer` a checkpoint before the deadline by hand. In practice that rarely happens: a workflow script has no `setTimeout`, and a parent in an interactive session is not awake at the right moment. The runner is the only place that knows the deadline, so let it issue the steer.

Measured on a Luna writer told to append 5 lines + `sleep 6` per tool call, `timeoutMs: 180000`:

| | without option | `checkpointBeforeDeadlineMs: 60000` |
|---|---|---|
| steer | none | requested/routed +118.6 s, queued mid-tool, delivered +126.4 s |
| ending | killed at 180 s, `timedOut`, 70 lines, no handoff | child replied `HANDOFF 60 60`, run completed +129.2 s |
| receipt | — | `status.json steering.recent[0].source = "deadline-checkpoint"`, target `delivered` |

## What

- New opt-in `checkpointBeforeDeadlineMs` (tool param, integer ≥ 1) with a global config default. **No default value**: when absent nothing changes.
- `runSubagent` arms one timer next to the existing run deadline timer and, if steps are still running, calls the existing `deliverSteerRequest` with an untargeted steer (`source: "deadline-checkpoint"`). Routing to running steps, `subagent.steer.*` events, `status.json` lifecycle and the receipt all come from the existing steering code; no new bookkeeping.
- Disarmed when the deadline would leave < 1 s before the checkpoint.
- Wired through `executeAsyncSingle` only (the docs say "async single-agent runs"); foreground runs ignore it.
- Pure helpers in `src/runs/background/deadline-checkpoint.ts` with unit tests; docs and CHANGELOG entry.

It is a request, not a guarantee: a child that ignores the steer is still killed at the deadline, but the receipt now shows delivered-but-not-complied instead of silence.

## Validation

- `npm run typecheck` clean.
- `npm test`: 3125 pass, 13 skipped, 1 fail (`owned process tree … kill EPERM`), which fails identically on a clean `main` checkout here.
- Live run above with `pi -ne -e <this checkout>`.


nicobailon:
Superseded by #2142, which preserves the checkpoint-handoff contribution with the required validation and contributor credit and has now landed as 23a2b9570740f950ced4d571d20df29a3bbc94e8. Thank you, @freezscholte, for the original implementation.

PR #2144, opened 2026-09-10
fix: re-drive after compaction only when it interrupted the parent run

mxp7064:
## Problem

`session_compact` sends the `subagent-compaction-resume` message (`triggerTurn: true`, "Resume the parent task now") after **every** compaction while async work is active. That matched pi's behaviour when the re-drive was added (#f54af731, 2026-08-04): compaction aborted the parent, so something had to restart it.

Two of the three compaction reasons no longer interrupt the parent:

- **threshold** — since pi 0.84.4 ("Pi now compacts between tool execution and the next assistant response in the same run", earendil-works/pi#6879) the run simply continues. The extra message then lands as a follow-up turn after the run has finished: a spurious "resume the parent task" prompt with nothing to resume. Measured in a long interactive session: 1 in 4 threshold compactions produced one.
- **overflow** — pi retries the aborted turn itself (`willRetry: true`).

Only **manual** compaction (`/compact`, `ctx.compact()`) aborts first and, per the `AgentSession.compact()` doc, "never retries or continues the interrupted agent turn" — and even then only when a run was actually in flight.

## Fix

- `agent_end` records whether the last assistant message ended with `stopReason: "aborted"`.
- `session_before_compact` with `reason === "manual"` arms the re-drive from that flag.
- `session_compact` consumes the flag; threshold and overflow never re-drive. Async-work and UI checks unchanged.

No wall-clock, no new config. `event.reason` has been on both compaction events since pi 0.6x.

## Tests

`test/unit/compaction-resume.test.ts` now covers the four cases: inline threshold (no re-drive), manual after an aborted run (re-drive), manual at idle (no re-drive), overflow with `willRetry` (no re-drive). `npm run test:unit`: 3126 pass / 0 fail; `tsc --noEmit` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Reject invalid global `checkpointBeforeDeadlineMs` values during config loading instead of silently disabling the requested checkpoint.
+- Send the `subagent-compaction-resume` re-drive only when a manual compaction actually cut a parent run. Threshold compaction runs inline inside the same run since pi 0.84.4 and overflow compaction retries the turn itself, so re-driving them started a spurious "resume the parent task" turn after runs that had already finished.
 
 ### Added
 
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -823,7 +823,8 @@ export default function registerSubagentExtension(pi: ExtensionAPI): void {
 
 	registerWaitTool(pi, state, waitToolConfig.enabled, waitSubscriptionManager, waitToolConfig.defaultTimeoutMs);
 
-	pi.on("agent_end", async (_event, ctx) => {
+	pi.on("agent_end", async (event, ctx) => {
+		parentRunAborted = lastAssistantStopReason(event?.messages) === "aborted";
 		if (!ctx.hasUI) await drainOutstandingWork({ state, events: pi.events });
 		const ownerSessionId = state.currentSessionId;
 		if (!ownerSessionId) return;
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks, @mxp7064. The automatic-compaction part of this change is the right direction, but review found one remaining case that prevents landing this head.

After a user independently aborts a parent run and it settles, a later idle `/compact` still sees the latched `aborted` stop reason and injects `subagent-compaction-resume`. That restarts work the operator had stopped. Clearing the latch on `agent_settled` cannot fix it because genuine manual compaction also aborts and settles before Pi emits `session_before_compact`.

Pi 0.85.1 currently exposes identical extension-observable histories for those two cases. We need a documented causal host signal—such as `interruptedActiveRun` captured when manual compaction begins, or a correlated pre-abort compaction event—before this can be implemented without a timing/private-field heuristic. I’m keeping the PR open with that concrete revisit trigger. When the host contract exists, we should preserve your contribution and add profile-linked CHANGELOG credit for #2144.

nicobailon:
Thank you @mxp7064 for identifying the automatic-compaction re-drive problem and for the work in this PR. I landed the narrower no-core-change fix in #2150: Pi retains its native threshold continuation and overflow retry, while the extension no longer sends duplicate automatic-compaction resumes. Your contribution is profile-linked in the changelog. The broader manual-causality distinction remains unavailable through Pi 0.85.1’s public extension API, so the replacement preserves the existing manual behavior rather than adding an unsafe latch.

nicobailon:
Closing now that the credited replacement #2150 has landed.

PR #2147, opened 2026-09-10
fix(reviewers): provide bounded diff evidence

tiratatp:
## Summary
- inject the existing bounded `watchdog_diff` tool for reviewer agents instead of giving them shell access
- block direct and symlinked Git metadata reads while leaving scout and non-reviewer capabilities unchanged
- preserve a run-scoped repository baseline across foreground, background, workflow, and resume paths without leaking it across repositories
- add focused regressions for capability ceilings, exclusions, symlink access, cross-repository runs, and nested resume

## Validation
- `npm run typecheck`
- focused affected unit tests: 96 passed
- `npm run test:integration`: 1033 passed, 6 skipped
- `git diff --check`
- `npm test`: 3127 passed, 13 skipped; one unrelated macOS environment failure remains in `test/unit/owned-process-tree.test.ts:71` (`kill EPERM`)

### DIFF
--- a/agents/reviewer.md
+++ b/agents/reviewer.md
@@ -50,7 +50,8 @@ Review a PR or issue by understanding the context, then verifying:
 - Tests and docs are updated as needed.
 
 ## Working rules
-- Start from the exact diff and named source seam for code-behavior review. Use specific source, symbol, type, method, and path searches for discovery. Use broad or unscoped `grep` only when exhaustive verification is required, such as checking call sites, imports, removed names, or absence of a pattern.
+- Start from `watchdog_diff` for bounded current-change evidence when it is available, then inspect changed files with `read`, `grep`, `find`, and `ls`. Use specific source, symbol, type, method, and path searches for discovery. Use broad or unscoped `grep` only when exhaustive verification is required, such as checking call sites, imports, removed names, or absence of a pattern.
+- Never read `.git` paths. If `watchdog_diff` is unavailable, fail closed: ask the parent for the exact diff or changed-file list instead of probing Git internals.
 - Read the relevant files first. Read plan and progress when the task supplies them.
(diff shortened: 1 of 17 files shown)


nicobailon:
Thanks for the thoughtful work here, @tiratatp. I reviewed the unchanged head and then ran independent scope, execution-seam, and security analyses. I'm closing this as not planned rather than asking you to revise it.\n\nThe safe same-run case is already owned by the existing child watchdog lifecycle: it captures the pre-mutation baseline at session start and gives its private read-only watchdog reviewer bounded `watchdog_diff` access without shell. General reviewer launches also need to keep shell because a session-start `HEAD` baseline cannot expose an already-committed PR/base diff.\n\nThe proposed generic path has no authoritative signal that a specific reviewer is causally reviewing an earlier writer. Inferring that from agent names, task text, lane metadata, ordering, or baseline presence would create an incorrect authority boundary. The reviewed head also has concrete linked-worktree common-gitdir, management hot-path, and provider-availability issues. Adding a new explicit workflow entitlement would be a separate product/security decision rather than a repair to this patch.\n\nI appreciate the clear motivation and extensive tests; they materially helped identify the boundary we need to preserve.

tiratatp:
Thanks—your authority-boundary explanation makes sense, and I agree the generic reviewer entitlement was the wrong approach.

One concrete symptom behind the PR: a fresh reviewer without exact diff evidence called `find("*")` and received 300 results dominated by `.git/objects`, refs, and worktree metadata before useful source files. The goal was reducing that context noise, not preventing legitimate inspection of adjacent code.

I’ll use the existing child watchdog for same-run reviews and pass an exact diff/changed-file manifest for generic reviews. Would you view excluding `.git/**` from Pi core’s default `find` behavior as the smaller upstream fix? That preserves shell-based `git diff` for committed PR reviews and avoids reviewer-role inference.

PR #2151, opened 2026-09-10
feat(workflows): add audited research orchestration

Muskos:
**What**

Add a new opt-in `/audited-research` workflow:

- Frame the research question, decision, and 2–3 distinct research angles.
- Run fresh-context research lanes in parallel using `researcher`, with `scout` only when local repository context is relevant.
- Pass successful research outputs and failed or missing lane context to one fresh `evidence-auditor`.
- Keep final synthesis owned by the parent.
- Preserve contradictions, uncertainty, and partial-failure information.
- Use existing managed output references for research and audit handoffs.
- Add focused workflow contract tests and concise documentation.
- Keep `/parallel-research` unchanged as the lightweight research workflow.

**Why**

`researcher` now produces stronger evidence, and `evidence-auditor` can independently validate decision-critical claims, but using them together still requires manual orchestration.

This PR composes the existing `workflowScript`, `runs.all`, `runs.run`, and artifact handoff patterns into one explicit workflow. It adds no new runtime primitives or persistence mechanisms.

**Motivation**

This is the third (and last) small step toward a deeper research flow on top of pi-subagents:

`parallel research → evidence audit → parent synthesis`

`/audited-research` makes that flow available as an explicit opt-in workflow while keeping `/parallel-research` lightweight.

This PR does not add `/deep-research`, iterative research, retries, automatic contradiction resolution, or a synthesis subagent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Add `/audited-research` for opt-in parallel research followed by independent evidence validation and parent-owned synthesis. Thanks to [@Muskos](https://github.com/Muskos).
+
 ### Fixed
 
 - Allow an explicit model request when a cached unavailable-model exclusion is contradicted by the current model registry, while preserving live health, auth, quota, and rate-limit exclusions. Thanks to [@xz-dev](https://github.com/xz-dev) for identifying the stale explicit-request cache symptom in #2145.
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks @Muskos — this is a cool feature and the composition is thoughtful. I’m going to pass because it adds too much code and maintenance surface for a niche workflow that many users won’t want. The existing primitives already let interested users assemble this flow without making it a built-in project surface.

Muskos:
Makes sense, thanks @nicobailon for the review. I agree that the existing primitives are enough to compose this as a user-level workflow, so keeping it out of the built-in surface is reasonable. I’ll keep the orchestration on the user side.

PR #2155, opened 2026-09-10
fix: avoid caching transient empty model responses

### LINKED ISSUE #2154: A single transient "produced no output" failure bans a model globally for 24h across all projects
I ran into this issue with a global model ban in pi-subagents for 24 hours. I had an agent debug and compile the report below.

## Summary

One transient empty-response failure (12.5s, zero tokens, possible cold start) against `openai-codex/gpt-5.6-terra` caused `recordRetryableModelFailure` to persist a 24-hour exclusion to the per-user store. Every new pi process on the machine — including those in **other, unrelated projects** — then refused to launch any subagent configured with that model, failing with:

```
No usable subagent models remain after registry, scope, and cached-exclusion filtering.
```

The model was healthy and present in the registry the whole time; the reviewer role on a sibling model of the same provider kept working. The only remedy was manually deleting `model-exclusions.json`.

Observed with pi-subagents **0.61.0** (bundled with patchmill 0.20.0); relevant code re-checked against **0.66.0**.

## What happened (concrete incident)

1. 08:44 UTC — a `worker` subagent on `gpt-5.6-terra:high` in project A failed after 12.5s with 0 tokens: "Subagent produced no output (possible model cold-start or empty response)."
2. `isRetryableModelFailureAttempt` special-cases that exact error as retryable (`src/runs/shared/model-fallback.ts`), so `recordModelFailure` wrote `{provider: "openai-codex", modelId: "gpt-5.6-terra"}` to `$TMPDIR/pi-subagents-<scope>/model-exclusions.json` with `DEFAULT_MODEL_EXCLUSION_TTL_MS = 24h`.
3. 16:39 UTC — project B (same machine, same user) tried to launch a `worker` with that model as its single candidate (no `fallbackModels`). `buildModelCandidates` → `filterFallbackCandidates` removed the only candidate → launch failed with the error above. The issue was auto-blocked pending human intervention.

### 1. One transient hiccup = 24h global ban

Cold-start/empty-response/transport failures are common provider-side blips. A single occurrence quarantines the model for 24 hours. There is no escalation (N failures within a window) and no graduated backoff.

### 2. The exclusion store is global per user → cross-project contamination

The store lives at `path.join(os.tmpdir(), "pi-subagents-<scope>", "model-exclusions.json")` (`src/shared/types.ts` TEMP_ROOT_DIR; `src/runs/shared/model-exclusions.ts` getExclusionsFilePath). Every repo and every pi session for the user shares it. A failure in project A silently blocks project B. `PI_MODEL_EXCLUSIONS_PATH` / `PI_SUBAGENTS_TEMP_ROOT` exist as overrides but are undocumented in the failure path, and nothing in the error message points at the store.

### 3. Load-once semantics make behavior inconsistent and hard to diagnose

`ensureLoaded()` reads the file once per process and caches it in memory. Long-running sessions never see exclusions recorded (or cleared) by other processes, while every new session sees all of them. In the incident, project B's _earlier_ run kept launching the "excluded" model all day; only the rerun — a fresh process — failed. This makes the failure look non-deterministic.

### 4. The stale-exclusion escape hatch does not cover "produced no output"

In 0.66.0, `ignoreStaleModelUnavailableExclusion` ignores exclusions whose reason matches `model not found / unknown model / model unavailable / model disabled` when the model is currently in the registry. A "produced no output (cold-start)" reason does not match, so even current versions honor a stale, transient exclusion for the full TTL. (In 0.61.0 there is no escape hatch at all.)

### 5. Diagnosability

The error names three filters (registry, scope, cached-exclusion) but does not say which one fired or where the store lives. 0.66.0 appends exclusion evidence when candidates were filtered, which helps; 0.61.0 shows the bare message. Consider always including the excluded candidates, the reason, expiry, and the store path, plus a hint on how to clear it.

## Suggested changes

1. **Graduated exclusions**: require N consecutive retryable failures (e.g. 2–3 within a short window) before persisting an exclusion, and/or use a much shorter TTL (minutes) for cold-start/empty-response/transport reasons, growing only on repeated failures.
2. **Scope the store to the project** (or make per-project the default and global opt-in), and print the store path in the error message.
3. **Extend the stale-exclusion escape hatch** to cover `produced no output` / cold-start reasons when the model is present in the active registry.
4. **Re-read or stat-watch the store** (cheap mtime check on use) so external clears take effect without a process restart and behavior is consistent across processes.
5. **Ship an operator affordance** to inspect and clear exclusions (slash command or `pi subagents exclusions list/clear`), and mention it in the error text.

## Reproduction

1. Configure an agent with a single model (no `fallbackModels`), e.g. `worker` → `openai-codex/gpt-5.6-terra`.
2. Simulate one transient failure for that model in any project (or write the entry directly):
   ```json
   { "version": 1, "exclusions": [
     { "modelId": "gpt-5.6-terra", "provider": "openai-codex",
       "reason": "Subagent produced no output (possible model cold-start or empty response).",
       "recordedAt": <now>, "expiresAt": <now + 24h> } ] }
   ```
   at `$TMPDIR/pi-subagents-<scope>/model-exclusions.json`.
3. Start a **new** pi session in a different project and launch that agent. It fails with "No usable subagent models remain…" although the model is healthy and registered.

## Environment

- pi-subagents 0.61.0 (also reviewed 0.66.0 sources)
- pi (earendil-works pi-coding-agent) on Linux
- provider `openai-codex`, models `gpt-5.6-terra` (worker), `gpt-5.6-sol` (reviewer)
- consumer: patchmill run-once (unattended issue implementation)

nicobailon:
## Summary
- keep canonical empty/no-output attempts eligible for same-launch fallback
- prevent those exact transient failures from creating durable 24-hour model exclusions
- preserve useful exclusion behavior for other provider/model failures

Closes #2154.

Thanks to [@rochecompaan](https://github.com/rochecompaan) for the detailed reproduction.

## Validation
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/model-fallback.test.ts` (109 passed)
- `npm run typecheck`
- `git diff --check`
- fresh independent review: no findings, merge verdict OK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Keep canonical empty child responses eligible for same-launch model fallback without persisting them as 24-hour model exclusions. Thanks to [@rochecompaan](https://github.com/rochecompaan) for #2154.
 - Let Pi continue threshold and overflow compactions without an extra extension resume, while preserving manual re-drive for active async work. Thanks to [@mxp7064](https://github.com/mxp7064) for #2144.
 - Allow an explicit model request when a cached unavailable-model exclusion is contradicted by the current model registry, while preserving live health, auth, quota, and rate-limit exclusions. Thanks to [@xz-dev](https://github.com/xz-dev) for identifying the stale explicit-request cache symptom in #2145.
 - Preserve wrapped Pi core tools and explicitly requested non-core tools in child launches. Core slots still respect host availability; non-core tools are validated in the child's runtime after ceilings and exclusions (#2132, #2133, #2134, #2135, #2140). Thanks to [@carlesba](https://github.com/carlesba) for #2137 and [@clementprevot](https://github.com/clementprevot) for #2138.
--- a/src/runs/shared/model-fallback.ts
+++ b/src/runs/shared/model-fallback.ts
@@ -589,10 +589,15 @@ function messageError(message: unknown): string | undefined {
 	return typeof value === "string" ? value : undefined;
 }
 
+function isTransientNoOutputFailure(error: string | undefined): boolean {
+	return error === "Subagent produced no output (possible model cold-start or empty response)."
+		|| /^Subagent produced no output after terminal assistant stopReason "[^"]+"\.$/.test(error ?? "");
+}
+
 export function isRetryableModelFailureAttempt(input: { error: string | undefined; messages?: readonly unknown[]; toolCount?: number }): boolean {
 	if (!isRetryableModelFailure(input.error)) return false;
 	if ((input.toolCount ?? 0) > 0) return false;
-	if (input.error === "Subagent produced no output (possible model cold-start or empty response)." || /^Subagent produced no output after terminal assistant stopReason "[^"]+"\.$/.test(input.error ?? "")) return true;
+	if (isTransientNoOutputFailure(input.error)) return true;
 	if ((input.toolCount ?? 0) === 0 && (input.messages?.length ?? 0) === 0) return true;
 	const error = input.error?.trim();
 	return Boolean(error && input.messages?.some((message) => messageError(message)?.trim() === error));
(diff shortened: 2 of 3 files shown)


rochecompaan:
As always, thank you for the quick fix @nicobailon! Out of interest, do you automate monitoring issues and bug-fixes in some way?

PR #2166, opened 2026-09-11
fix(watchdog): surface provider error text in failed-review lastError

freezscholte:
## Problem

`runWatchdogReview` already reads `terminal.errorMessage` from the failed assistant message to decide whether the review is retryable, but drops it from the returned result. `MainWatchdogRuntime` can therefore only record:

```
Last error: Watchdog review ended with stop reason 'error'.
```

In practice this made a dead watchdog invisible. Observed on a project with `watchdog.enabled: true`, cadence every 10 tools, boundary reviews on: a ~14 h session accumulated `Failed reviews: 5` and produced **zero** findings across the whole project history, and there was no way to tell rate limit from rejected model id from auth from payload size — or from "every review was clean". Only after a `/reload` (fresh runtime) did reviews start passing, which points at a long-lived-runtime issue I can now only diagnose once the error text is visible.

## Change

- `review.ts`: return `errorMessage` next to `stopReason` when the terminal message carried one.
- `runtime.ts`: optional `errorMessage` on `WatchdogReviewResult`; `fail()` message appends a bounded (600 char, `boundWatchdogReviewText`) copy, so `/subagents-watchdog status` → `Last error:` shows the cause.
- Tests: runtime test asserts the text reaches `lastError`; review test asserts the context-overflow case returns it.
- CHANGELOG entry under Unreleased/Fixed.

No change to review, retry or fallback behaviour; this is observability only. `tsc --noEmit` clean; `watchdog-runtime` 41/41 and `watchdog-review` 22/22 pass.

## Follow-up (not in this PR)

Unclassified `error` outcomes are deliberately not retried on `fallbackModels` (per the existing "does not retry … unclassified error outcomes" test), so a configured fallback never engaged during the 5 failures above. That may be the right policy, but with the error text now visible I'd like to gather a few real failures before proposing anything there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Surface the provider error text of a failed watchdog review in `/subagents-watchdog status` `Last error` (bounded to 600 chars). Previously only `stop reason 'error'` was recorded, so a watchdog failing every review (rate limit, rejected model, auth) was indistinguishable from a clean one.
 - Reject invalid global `checkpointBeforeDeadlineMs` values during config loading instead of silently disabling the requested checkpoint.
 
 ### Added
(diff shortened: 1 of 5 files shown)


nicobailon:
Thanks @freezscholte — your watchdog provider-error fix landed via #2172 in 1828ee35a79794952b0c8bde1e1d73ce0868a85d, with commit co-author attribution and changelog credit. I used the maintainer replacement because this fork PR conflicted with the current changelog and could not be updated from the repository. Closing this superseded PR.

PR #2169, opened 2026-09-11
fix(slash): include runtime-registered agents in /subagents admin panel

mystery4f:
## Problem

The `/subagents` admin panel only lists statically discovered agents. `allVisibleAgents()` (`src/slash/subagents-admin.ts`) builds its list from `discoverAgentsAll()` (project / user / package / builtin) and never consults the runtime agent registry.

Runtime agents registered through the `pi-subagents:runtime-agent-register:v1` event are invisible there, even though every other path merges them via `mergeRuntimeAgents()`:

- delegation resolution (`effectiveAgentsForScope`, `agent-management.ts`)
- the `subagent` tool list action
- `/run` (`discoverSlashAgents`)

### Reproduction

Any extension that registers a runtime agent — for example [pi-flow-composer](https://github.com/mystery4f/pi-flow-composer), which auto-injects `templates/flow/*/agents/*.md` as runtime agents on session start. The agents are delegable (`subagent({ agent: "sdd-explore", ... })` works) but absent from `/subagents`.

## Fix

- `allVisibleAgents(pi, cwd)` merges the runtime registry after static discovery — same pattern as `effectiveAgentsForScope()` in `agent-management.ts`.
- `pi` is threaded through `selectAgent()` from `openSubagentsAdmin()`.
- `readOnlyAgentMessage()` treats `runtime` source as read-only: these agents are owned by the registering extension, and a settings override written by the panel would never apply to them. The message points the user to the source definition instead.

## Verification

- Live pi install with 6 runtime-registered agents (`sdd-*`): after the patch they appear in the `/subagents` selector labeled `[runtime]`, and attempting to edit one shows the read-only message.
- Delegation, `subagent` list, and `/run` behavior unchanged (runtime merge was already active on those paths).
- `mergeRuntimeAgents` collision semantics are preserved: a runtime name colliding with a configured name still surfaces the existing hard error, consistent with the other merge call sites.

### DIFF
--- a/src/slash/subagents-admin.ts
+++ b/src/slash/subagents-admin.ts
@@ -11,6 +11,7 @@ import {
 	type BuiltinAgentOverrideBase,
 } from "../agents/agents.ts";
 import { serializeAgent } from "../agents/agent-serializer.ts";
+import { mergeRuntimeAgents, type RuntimeAgentOwner } from "../agents/runtime-agent-registry.ts";
 import { editableAgentConfig, preservedAgentFrontmatterFields } from "../agents/agent-management.ts";
 import { findModelInfo, getSupportedThinkingLevels, toModelInfo } from "../shared/model-info.ts";
 import { SelectorComponent, type SelectorItem, type SelectorResult } from "./selector.ts";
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks @mystery4f. Your runtime-agent admin-panel contribution was preserved and landed through the corrected maintainer replacement #2171 in commit 9f112d05e6d2ecdd4db73a3dced39e2d4c1b8ab4. The landed version adds the disabled-definition collision fix, focused regression coverage, and visible changelog credit.

Issue #2168, opened 2026-09-11
RPC async-status snapshot lists workflow children twice since 0.66 (as the parent's step and as a root run)


niko-operal:
### Summary

Since 0.66 (#1964), a workflow's children run as tracked jobs of their own, each carrying `parentWorkflowRunId` and `workflowKey`. The RPC async-status snapshot (`PI_SUBAGENT_ASYNC_JSON:` widget line, kind `pi-subagents.async-status-snapshot`) projects every tracked job as a root run, so each live child is listed twice:

- as a `step` node under the workflow node (`id` = workflow key, `label` = lane label), and
- as a root `subagent` run labelled after its agent, with one `step` child repeating the agent name.

Nothing in the projected nodes links the two: the step node does not carry `runId`, and the child node carries neither `workflowKey` nor `parentWorkflowRunId`.

The TUI widget does not show the duplicate: `widgetJobTree` groups children under their parent by `parentWorkflowRunId`, and `materializedKeys` hides a lane already executed by a child. RPC hosts receive the raw projection and cannot do the same join.

### Reproduction

1. Run pi in RPC mode with pi-subagents 0.67.0 loaded.
2. Start a workflow whose `runs.all` launches several children on the same agent, e.g. five `reviewer` lanes keyed `rev-core`, `rev-analyse`, `rev-main`, `rev-r`, `rev-parity`.
3. Read the `subagent-async` widget while children are still running.

Observed in `runs[]` (abridged): the workflow node with its five `step` children, plus one root run per live child:

```json
{
  "id": "<child run id>",
  "kind": "subagent",
  "label": "reviewer",
  "state": "running",
  "children": [
    { "id": "step:0", "kind": "step", "label": "reviewer", "state": "running" }
  ]
}
```

Finished children leave the job list, so only running lanes are duplicated. A host rendering one row per node shows `rev-r` and `reviewer` side by side for the same work, once per live lane.

The tracked child job's `status.json` does carry the link (`parentWorkflowRunId`, `workflowKey`), and the parent's `steps[].runId` points back at the child, so the data to join on is available to the projector.

### Expected

One node per lane: the workflow's `step` node keeps its identity (workflow key and label) and takes the live facts from the child job (state, timing, `activity`), as the TUI's grouped widget already shows it. A child that matches no lane stays under its parent as an ordinary child. A child whose parent is not part of the snapshot stays a root run, unchanged.

### Proposed fix

[...]

Issue #1964, opened 2026-09-05
Stream bounded live child activity in synchronous workflow progress


nicobailon:
Split from #1952.

## Summary

Synchronous `workflowScript` children expose lifecycle state, but not the bounded live activity that single foreground children already stream through `details.progress`. This leaves RPC hosts and the TUI live card with only "running"-style state until a child settles.

## Scope

Forward bounded per-child activity for foreground children of synchronous workflows.

Preferred shape:
- reuse the existing single-run progress snapshot source;
- update `workflowChildren` rows by stable workflow key from the child execution callback;
- update the type, strict parser, and builder together;
- preserve #1961's lifecycle-progress contract and same-repo emission gating;
- avoid filesystem polling, full transcript forwarding, and async workflow persistence changes in this lane.

## Acceptance

- Mid-run workflow child rows can show bounded `currentTool`, timing, model/thinking, and counters comparable to single-run progress where available.
- Concurrent children keep stable identity and do not swap data between keys.
- Terminal states clear or freeze activity intentionally.
- Byte ceilings and emission coalescing prevent hot-path regressions.
- Existing #1961 workflow progress tests keep passing.

Thanks to [@yanqianglu](https://github.com/yanqianglu) for the RPC-host parity audit in #1952.

PR #2174, opened 2026-09-11
fix: group materialized workflow children in the async status snapshot

niko-operal:
Fixes both defects reported in #2168. Projection-only: nothing changes in what the tracker persists.

## Workflow children listed twice

Since 0.66 (#1964) a workflow's children run as tracked jobs of their own, carrying `parentWorkflowRunId` and `workflowKey`. The RPC async-status projection walked every tracked job as a root run, so each live child appeared twice — as a `step` under the workflow node, and as a root `subagent` run named after its agent, with one `step` child repeating the agent name. The projected nodes carried nothing to join on, so an RPC host could not de-duplicate the way `widgetJobTree` does for the TUI.

`groupMaterializedChildren` now groups jobs by `parentWorkflowRunId` when that parent is a `workflow` job present in the snapshot — the same rule the widget applies — and `projectMaterializedStep` projects a lane and its child as a single node:

- the lane keeps its identity: `kind: "step"`, its workflow key as `id`, its label;
- the child supplies the live facts — state, timing, activity — and its own descendants;
- a single-mode child's lone self-describing step is not repeated underneath, since it *is* that node;
- a child matching no lane is appended to its parent's children, unchanged;
- a child whose parent is absent from the snapshot stays a root run, unchanged;
- the `omitted` counters count a lane and its child once.

Two supporting changes fell out of this:

- `projectWorkflowGraphNode` becomes `workflowGraphStep`, returning the `AsyncJobStep` a planned stage stands for, so graph stages take the same lane-matching path as loaded steps.
- `projectRun` now carries `depth` explicitly and bounds children with `depth < maxDepth` instead of the former `maxDepth > 0`, so a materialized child's own subtree is bounded like any other child. `stepIdentity` factors out the identity both paths derive.

## Finished steps had no end time

A completed step is persisted with `startedAt` and `durationMs` but neither `endedAt` nor `lastActivityAt`, so a terminal lane's node had a start and no end. A host computing time worked from the node kept counting it on every republish — visible as four finished lanes out of five showing ✓ with their time still growing alongside the running one. `projectStep` now derives `endedAt = startedAt + durationMs` when the tracker recorded no end; `updatedAt` follows it as it already did.

## Tests

Four cases added to `test/unit/async-status-projection.test.ts`, on a workflow of three lanes (one finished, one running, one still planned):

- a matched lane: listed once, lane identity kept, live facts and activity taken from the child, no repeated self-step, and the finished lane's end derived from its duration;
- a child matching no lane kept under its parent, and a child whose parent is absent kept at the root with its pre-existing shape;
- omission accounting at `maxDepth: 0` — a lane and its child counted once;
- the derived end time on a plain step with no materialized child.

## Verification

`npm run typecheck` is clean and the full unit suite passes (3202 of 3206). The four failures are the `control-channel` inbox-diagnostics timing cases, which fail identically on a clean checkout of `main` in my environment. No integration test touches this projection. `npx oxlint` on both changed files reports the same 61 pre-existing findings as `main` — the change adds none.

### DIFF
--- a/src/runs/shared/async-status-projection.ts
+++ b/src/runs/shared/async-status-projection.ts
@@ -362,11 +362,51 @@ function groupMaterializedChildren(jobs: readonly AsyncJobState[]) {
 	return { roots, childrenByParent };
 }
 
-/** The child job executing a workflow step, matched by workflow key or by run id. */
-function materializedChildForStep(step: AsyncJobStep, children: readonly AsyncJobState[]): AsyncJobState | undefined {
-	return children.find((child) =>
-		(step.workflowKey !== undefined && child.workflowKey === step.workflowKey)
-		|| (step.runId !== undefined && child.asyncId === step.runId));
+/**
+ * Assigns the child job executing each workflow lane, one child per lane at most.
+ * Run ids are matched across every lane before any workflow key, because a declared
+ * key may legitimately be reused: a lane naming its exact child must keep that child
+ * rather than lose it to whichever sibling lane the scan reaches first.
+ */
+function assignMaterializedChildren(lanes: readonly AsyncJobStep[], children: readonly AsyncJobState[]) {
+	const childByLane = new Map<AsyncJobStep, AsyncJobState>();
+	const claimed = new Set<AsyncJobState>();
+	const claim = (lane: AsyncJobStep, child: AsyncJobState | undefined): void => {
+		if (!child || claimed.has(child)) return;
+		childByLane.set(lane, child);
+		claimed.add(child);
+	};
+	for (const lane of lanes) {
+		if (lane.runId !== undefined) claim(lane, children.find((child) => child.asyncId === lane.runId));
+	}
+	for (const lane of lanes) {
+		if (childByLane.has(lane) || lane.workflowKey === undefined) continue;
+		claim(lane, children.find((child) => child.workflowKey === lane.workflowKey && !claimed.has(child)));
+	}
+	return childByLane;
+}
+
+/**
+ * The freshest activity of a grouped tree. A workflow must not rank behind its own
+ * live children: a finished or paused parent keeps a frozen `updatedAt` while a
+ * detached child still works — cleanup is held back for exactly that case — and the
+ * root cap would otherwise drop the parent and hide the child with it.
+ */
+function treeUpdatedAt(job: AsyncJobState, childrenByParent: MaterializedChildren, seen = new Set<string>()): number {
+	let freshest = job.updatedAt ?? job.startedAt ?? 0;
+	if (seen.has(job.asyncId)) return freshest;
+	seen.add(job.asyncId);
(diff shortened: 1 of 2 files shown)


niko-operal:
Both review findings were real — I reproduced each against the first commit before touching anything. Fixed in 95e6e60, with a regression test per finding.

**Lane keys reused.** Confirmed: with two lanes keyed `writer` and children `child-1` / `child-2`, the second lane carried the first child's start and tool while `child-2` reappeared as a row of its own.

```
before: [{id:writer, label:First,  startedAt:1100, tool:read},
         {id:writer, label:Second, startedAt:1100, tool:read},    ← child-1 again
         {id:child-2, label:worker, startedAt:2100, tool:bash}]   ← unmatched
after:  [{id:writer, label:First,  startedAt:1100, tool:read},
         {id:writer, label:Second, startedAt:2100, tool:bash}]
```

I went further than matching the run id before the key inside the same scan: that still lets a lane carrying no run id claim a child another lane names exactly, depending only on scan order, and it does not stop one child from being claimed by two lanes. Lanes are now assigned in two passes — every run id first, then keys among the children still unclaimed — with each child claimed at most once. That assignment is also what the depth-capped branch counts against, so its omission count no longer double-counts a child two lanes matched.

**Children hidden behind a capped parent.** Confirmed, including the accounting hole: with 21 roots and `maxRuns: 20`, the frozen parent was dropped, its running child disappeared, and `omitted` reported `{ runs: 1, children: 0 }` — the child counted nowhere.

I took the second of the two suggested remedies. The first — keeping live children as roots when their parent is not running — would reinstate the duplication this PR removes: the child would be both its parent's lane node and a root run. So roots are ranked on the freshest activity in their grouped tree, and a dropped root is counted together with the runs grouped under it, for the root cap and for the byte-limit trim alike (the trim had the same hole).

One correction to the framing: a *paused* workflow is not the case that bites, since a running workflow's `updatedAt` is refreshed from its own status as its lanes advance. The case that does is a workflow that has **ended** while a detached child works on — the state `hasLiveNestedDescendants` deliberately holds cleanup back for. That is what the new test covers: a `complete` parent at `updatedAt: 1` holding a child at `updatedAt: 9999` now ranks first instead of falling off the cap.

`npm run typecheck` clean, 3191 of 3209 unit tests pass — the four failures are the pre-existing `control-channel` inbox-diagnostics timing cases, which fail identically on a clean `main` here. `npx oxlint` on both files still reports the same 61 findings as `main`.

nicobailon:
Thanks @niko-operal — your projection fix landed via #2177 in aa0d8af6ad6baa60ecee696f5b241e25efdfc8fc, preserving commit co-author attribution and changelog credit. The landed version adds one-to-one child assignment, live-root visibility and de-duplication, deterministic capped ordering, cycle handling, truthful omission accounting, and safe terminal end-time derivation. Closing this original PR as superseded.

PR #2192, opened 2026-09-12
fix: allow cold runners to reach resume readiness

qsgy-edge:
Direct resume currently gives a detached runner 10 seconds to publish `ready`, including loading its TypeScript/SDK dependency graph. On a Windows Node 24.18.0 host, dependency loading alone took 20.9 seconds; resume terminated the runner before it reached the handshake. Warm-cache retries succeeded.

Give the initial readiness phase a separate, bounded 60-second budget. After `ready`, the parent still requires `acknowledged` within 10 seconds and validates the token before authorizing `proceed`.

The integration regression delays a real Node runner by 12 seconds before module loading, then resumes the original session through the public executor. It reproduced the original `Timed out after 10000ms ... 'ready'` failure before the fix. A second real-process case suppresses the acknowledgement marker, verifies the 10-second failure and absence of child-session execution, waits for observed process close, then successfully resumes again. The retry waits for the parent's observed close proof, so lease checks run after termination has been processed on POSIX as well as Windows.

The readiness wait remains synchronous: a runner that never publishes readiness can take up to 60 seconds to reject. Healthy handshakes return as soon as their markers arrive. This change adds no startup configuration or protocol stage.

Validation for final revision `4e1d777`:

- [Upstream CI](https://github.com/nicobailon/pi-subagents/actions/runs/34684064145): all five jobs passed (Ubuntu, Windows, Bun workflow, Pi 0.85 clean-install, official standalone).
- Ubuntu: 3216 unit tests passed / 14 skipped; 1061 integration tests passed / 6 skipped.
- Windows: 3180 unit tests passed / 47 skipped; 1011 integration tests passed / 56 skipped.
- Both added real-process tests passed on Ubuntu and Windows; local typecheck and the updated close/retry case also passed.

The added tests give initial session creation up to 60 seconds as fixture setup; the injected delay and production readiness deadline are exercised only on the subsequent resume. Independent review of frozen `4e1d777` passed with no findings. It reran the updated close/retry test and checked the unchanged production logic against the previously executed no-ready, wrong-token, failed-proceed and lease-contention probes. Cross-platform results come from the CI above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Allow cold resumed runners up to 60 seconds to load and publish `ready`, while retaining the 10-second acknowledgement limit and the existing lease/token startup authorization.
+- Allow cold resumed runners up to 60 seconds to load and publish `ready`, while retaining the 10-second acknowledgement limit and the existing lease/token startup authorization. Thanks to [@qsgy-edge](https://github.com/qsgy-edge) for #2192.
 
 - Include retention-managed async, output-artifact, and structured-output retrieval paths in native completion notices. Thanks to [@peedrr](https://github.com/peedrr) for #2181.
 - Remove workflow-owned one-shot result payloads together with their child-local result indexes after successful consumption. Thanks to [@peedrr](https://github.com/peedrr) for #2182.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Fixed
 
+- Allow cold resumed runners up to 60 seconds to load and publish `ready`, while retaining the 10-second acknowledgement limit and the existing lease/token startup authorization.
+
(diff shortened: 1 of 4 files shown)


qsgy-edge:
The exact-head CI run for this PR has one failure on `test (windows-latest)` during the existing integration test `aborts a blocked Git baseline before launching the external process`.

The test reached 1010 passed, 56 skipped, and 1 failed; the failure occurred in its cleanup hook when Windows returned `EPERM` while removing a temporary directory. The PR's two new cold-start/acknowledgement regression tests both passed in that same job, and the other four CI jobs passed.

Could a maintainer please re-run the failed Windows job? I cannot trigger it with my current repository permissions (`gh run rerun` returns “Must have admin rights to Repository”).

qsgy-edge:
关闭此 PR。当前改动只是将 runner `ready` 等待从 10 秒延长到 60 秒，属于临时止血；后续需要先对真实启动链做分段计时，确认冷启动耗时来自 Node/Jiti、依赖加载、Pi 初始化还是文件系统，再提交根因修复。

感谢审查和 CI 验证。

PR #2199, opened 2026-09-12
Add opt-in unbounded local inference for native subagents

JordiPosthumus:
Local inference can legitimately remain queued or prefill a large context for more than five minutes. The detached runner's proxy dispatcher currently retains Undici's 300,000 ms header/body defaults, even when the operator sets much longer Pi idle and subagent execution timeouts. In an observed incident, each failed lane made four requests lasting approximately 301–302 seconds; retry backoffs made this look like one twenty-minute failure. Gateway logs showed cancellation while those requests were still queued.

This proposes an opt-in owner configuration, `localInference: { enabled: true }`, that removes automatic native child execution and inference deadlines while preserving operator cancellation. The invariant is that a model-supplied run deadline or a separate HTTP default cannot silently override that explicit owner policy. Default behavior remains unchanged.

Each native child receives its own proxy-aware dispatcher. The transport forwards Pi's cancellation signal instead of the provider SDK's separate deadline signal. Policy is visible in doctor output and child session records; invalid policy fails explicitly. Tool/control waits and usage budgets keep their existing semantics.

Validation:
- Typecheck and 53 focused tests pass on current upstream main plus this patch.
- The regression harness exercises the real Pi 0.85.1 CLI, public subagent tool, detached runner and OpenAI-compatible HTTP stack. It checks delayed headers/body, one request per lane, foreground execution, manual stop, HTTP disconnection, runner exit and capacity release.
- This upstream branch passed the full 310-second header/body regression on npm Pi 0.85.1, including foreground/background completion, manual cancellation and capacity release. The installed fork also passed the same long gate locally and in Linux CI. All fixtures used isolated configuration and a local HTTP server.

Keeping this as a draft for maintainers' review of the policy/API. The tested provider boundary is native OpenAI-compatible requests on npm Pi 0.85.1; other providers, standalone runtimes and external CLI clients need separate validation. A permanently stalled connection requires operator cancellation. No private infrastructure configuration or fork updater/branding is included.

### DIFF
--- a/.github/workflows/local-inference.yml
+++ b/.github/workflows/local-inference.yml
@@ -1,0 +1,25 @@
+name: Local inference regression
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+permissions:
+  contents: read
+jobs:
+  local-inference:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-node@v5
+        with:
+          node-version: 22.22.2
+      - run: npm ci --ignore-scripts
+      - run: npm run typecheck
+      - run: npm run test:local-inference
+      - name: Install isolated real Pi runtime
+        run: npm install --prefix "$RUNNER_TEMP/pi-host" --ignore-scripts @earendil-works/pi-coding-agent@0.85.1
+      - name: Exercise real detached HTTP waits beyond five minutes and manual cancellation
+        run: node tools/check-local-inference.mjs "$RUNNER_TEMP/pi-host/node_modules/@earendil-works/pi-coding-agent" --long
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ pi install npm:pi-subagents
 
 That is the only required step. Background children use the host's SDK: npm Pi keeps its detached Node runner; the official Pi 0.85.1 Linux x64 standalone release loads the same runner through Pi's embedded SDK, without a separate SDK install. See [Standalone background execution](docs/standalone-background.md) for the supported boundary and validation gate.
 
+For local inference with long queue or prefill waits, see the opt-in [unbounded local inference policy](docs/local-inference.md).
+
 ## Try this first
 
 You do not need to create agents, write config, or learn slash commands. After installing, ask Pi in plain language:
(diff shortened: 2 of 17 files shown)


nicobailon:
Thanks @JordiPosthumus — the incident evidence, explicit owner-policy framing, cancellation regression, and real-stack harness establish a credible problem. I am keeping this draft **held for rework/reduction**, not closing it. Current blockers on head `19c3869b`: (1) the declared peer/runtime range includes Pi 0.81, whose OpenAI transport does not forward `options.fetch`, so the policy can remove execution deadlines without installing the promised dispatcher; (2) deadlines are stripped before runner/provider selection, affecting external runners and unvalidated native providers beyond the documented OpenAI-compatible boundary; (3) the full unit suite has a deterministic doctor formatting regression because the policy line interrupts the Workflow script section; and (4) the head lacks current-main exact-head checks and the advertised long transport/cancellation evidence. Please narrow the next revision to a proven native OpenAI-compatible runtime/transport boundary, preserve normal deadlines or fail closed everywhere else, give doctor output its own coherent section, and keep the long 310s gate scheduled/manual unless CI-cost evidence supports every-push use. Rebase on current main and include one current-runtime long cancellation/transport proof. I will re-review that reduced head.

nicobailon:
Closing as **not planned**, not as invalid or fixed.

Thank you, @JordiPosthumus, for the incident analysis, explicit owner-policy design, and real-stack cancellation harness. The motivating problem is credible, but this draft removes execution deadlines before the effective runner/provider path is known. That can remove the capacity backstop from external runners and providers that do not forward the injected fetch. A callable `streamFunction`, provider label, or package version does not prove end-to-end transport enforcement; observed Pi runtimes already differ on whether the fetch seam is forwarded.

A safe reduction needs an affirmative public transport capability available before deadline resolution, bound to the effective runtime/provider path, and reevaluated across fallback/model changes. Unsupported runtimes/providers and external runners must retain normal deadlines. A future proposal should also preserve operator cancellation and terminal/capacity evidence, keep connection establishment bounded, give doctor output its own section, and keep the 310-second gate manual or scheduled.

Exact reviewed head: `19c3869b19d2e92a5576b938225d2050afe23514`. No contributor branch changes were made.

nicobailon:
Follow-up: the narrow fix is up as #2227.

@JordiPosthumus, thank you again for the incident trace and the real-stack harness. They pointed at the actual defect: the detached runner installs its own `EnvHttpProxyAgent` (from #1867) and never applies Pi's `httpIdleTimeoutMs` to it, while Pi's CLI and the SDK request clock both honour that setting. That is why only async lanes hit the 301 s wall and foreground children did not.

#2227 reads `httpIdleTimeoutMs` in the runner (project `.pi/settings.json` over `~/.pi/agent/settings.json`, `0` disables) and passes it as `headersTimeout`/`bodyTimeout`, matching Pi's dispatcher, with a regression test. No new config: raise or disable `httpIdleTimeoutMs` in Pi settings and set `timeoutMs` as long as the run needs. The changelog entry credits you for the diagnosis.

If your environment still shows a cut after that lands, please reopen with the trace and I will look again.

Issue #2145, opened 2026-09-10
model:thinking suffix breaks child launch (Pi SDK uses separate thinkingLevel) and poisons model-exclusion cache for 24h


xz-dev:
## Summary

`applyThinkingSuffix` appends a `:thinking` level to the model id passed to the spawned child's `setModel`, but the Pi SDK/core model registry does not parse that syntax — thinking is a **separate `thinkingLevel` field** in the SDK (`setModel(model)` + `thinkingLevel`), never part of the model id. The child therefore fails with `Model "provider/id:high" not found. Use --list-models`, and the failure is recorded in the model-exclusion cache (keyed on the correctly stripped id) — so every subsequent request for that model, **even without any suffix**, is refused for the cache TTL (24h).

## Reproduction

1. Any agent with a thinking default (e.g. `delegate` with `thinking: max`), or any dispatch passing `model: "provider/some-model"`:
   ```
   subagent({ agent: "delegate", model: "openai-api-extension/commandcode/xai/grok-4.6", task: "..." })
   ```
2. Child launch fails:
   ```
   Model "openai-api-extension/commandcode/xai/grok-4.6:max" not found. Use --list-models to see available models.
   ```
3. All later attempts — including `model: "openai-api-extension/commandcode/xai/grok-4.6"` with **no suffix** — fail with:
   ```
   Requested subagent model '...' is excluded and cannot be replaced by a fallback (reason: Model "...grok-4.6:max" not found...; expires: <now+24h>)
   ```

## Root cause

- `src/runs/shared/child-tool-plan.ts` — `applyThinkingSuffix(model, thinking)` returns `` `${model}:${thinking}` ``.
- Call sites (`src/runs/foreground/execution.ts:378`, `src/runs/background/async-execution.ts:931`, `src/api/preflight.ts:379`) pass that suffixed string as the child model.
- Pi SDK contract (`docs/sdk.md`): `setModel(model: Model)` takes a plain model; thinking is set separately (`thinkingLevel: "off"|"minimal"|...|"max"`, see the `{ model: opus, thinkingLevel: "high" }` example). Stock Pi does not parse `provider/id:level` in `--model` or `setModel`.

## Secondary damage

[...]

PR #2204, opened 2026-09-12
fix: cache provisioning failures with a short TTL instead of 24h

rnavarro:
### Symptom

I hit a child dispatch that was refused for a full 24h even though the model never ran:

```
Requested subagent model wafer/DeepSeek-V4-Flash-0731 is blocked and cannot be replaced by a fallback (reason: npm install pi-prompt-template-model --prefix /tmp/pi-packages-auto-policy-profile/.pi/npm --legacy-peer-deps failed with code 217; expires: ...
```

The recorded reason is a provisioning failure from the child-launch phase. Before the child starts, pi installs its skill/extension packages into the child prefix with `npm install ... --prefix <cwd>/.pi/npm`, and that install exited 217 (transient npm failure, classically two concurrent installs racing into the same prefix). No provider was contacted, yet the failure was cached as a model exclusion with the default 24h TTL, so every dispatch of that model failed during the whole window.

The text is only classified as a model-health failure by accident: `npm install pi-prompt-template-model ... failed with exit code 217` matches the broad `/model.*(?:load|fail|error)/i` pattern because the package name contains "model" and the sentence contains "failed". Any cold-start failure (tool count 0, messages 0) is treated as retryable and cached for the full default TTL.

### The change

Detect provisioning-shaped reason text (`npm install|ci|run`, a `failed with code N` or `exited with code N` line, `preflight`) and record those exclusions with a 15 minute TTL. A broken environment is still dampened across dispatches, but a healthy model is not blamed for a full day. Model-health failures keep the default 24h TTL.

New tests cover: the exact npm exit-217 text records with exactly 15 minutes; provisioning text carrying a retryable signal records with the short TTL; model-health failures keep the default; and npm text with no failure signal records nothing.

### Related work

This is the provisioning class that remains after #2145 / #2148. That PR narrowed the stale bypass to reasons matching the not-found/unavailable set; the npm/bootstrap class matches a different (accidental) pattern and is untouched there, and I found no existing issue covering it.


nicobailon:
Superseded by merged PR #2212 (`3c7f9b98f3d5cd206082451a3e2284a0f202fd8d`). The replacement preserves your original provisioning-failure change as an authored commit and keeps the changelog credit to @rnavarro, while adding the reviewed live-diagnostic and shorter-configured-TTL corrections. Thank you for the contribution.

PR #2219, opened 2026-09-13
fix: keep revival startup handshake non-blocking

qsgy-edge:
## Problem

A revival runner startup failure blocked the parent event loop with synchronous polling. When the runner exited during `ready` or `acknowledged`, the parent could wait for the full startup timeout, delay close and terminal evidence, retain workflow capacity, or publish a failed nested revival as complete.

## Change

- Poll revival startup state with `setTimeout`/Promises so the parent event loop remains responsive.
- Observe runner close before accepting a startup state and preserve runner-written startup errors.
- Defer revival close finalization until startup failure persistence has written the zero-writer candidate.
- Keep ordinary non-revival launches on their existing synchronous startup path.
- Await the async result at the three production call sites that can launch revival or async single runs.

## Validation

- `npm run typecheck` — passed
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/async-execution.part-3.test.ts` — 58/58 passed
- `git diff --check` — passed

The repository-wide test command still has unrelated Windows environment and fixture failures; no failure was reported by the affected async integration file.


nicobailon:
Superseded by #2222, which preserves @qsgy-edge's contributor commit, authorship, timestamp, and changelog credit while adding the lifecycle corrections required by exact-head CI and review. Thank you, @qsgy-edge.

PR #2220, opened 2026-09-13
fix: avoid jiti for native async runner startup

qsgy-edge:
## Problem and change

Cold source-layout runners evaluate their TypeScript dependency graph through Jiti before publishing startup readiness. This PR uses native TypeScript stripping where the Node runtime and runner layout support it, while preserving Jiti for older runtimes and runners installed inside `node_modules`.

- Gate native loading on supported Node/preload versions: 22.19+, 23.5+, or 24+ within their respective major lines. Node 22.14 selects Jiti, and its preload avoids importing the unavailable `registerHooks` export.
- In native mode, redirect only `@earendil-works/*` host peers and resolve source `.js` imports to `.ts`. Ordinary dependencies such as `typebox` retain their own resolution, so child provider extensions load correctly.
- Preserve the existing binary-host path and Jiti peer resolution.

Based on `f6d2135ec2ca4010b83b7d5ee46774176d49d1e6`, including #2222 (superseding #2219) and the subsequent canonical agent-name/home fixes. The merged startup handshake and stale-result reconciliation are unchanged.

## Validation

All checks below ran locally on final head **`13b68d7cd3723bbf970fb92764291caada947ac7` before pushing**:

- Typecheck passed. Linux full unit suite: **3346 passed, 14 conditional skips, 0 failures**. Linux full integration suite: **1065 passed, 6 conditional skips, 0 failures**.
- Windows affected unit tests: **68/68**. Windows `async-execution.part-3`: **59/59**, including exit-before-ready and exit-after-acknowledgment regressions from #2222.
- `async-spawn-preload` under actual **Node 22.14.0 and 22.19.0**, with explicit expected Jiti/native spawn arguments: **2/2 on each**. The old-host preload regression was first observed failing on Node 22.14 before the fix.
- Original `standalone-matrix.mjs` with the checksum-pinned official Pi 0.85.1 Linux binary: **18/18**, including revival.
- Original `pi085-clean-install.mjs` on Node 22.19: **Pi 0.85.0 and 0.85.1 passed**. Original `npm-background.mjs`: **passed**, including provider registration, public launch, observed exits and the network-unshared receipt.
- A separate probe imported the actual npm-installed artifact and captured its production spawn boundary: **Jiti**, `PI_ASYNC_NATIVE_RUNNER=0`, no stripping flag for the `node_modules` runner.

Linux verification used WSL, temporary Git 2.48.1 and bubblewrap 0.6.1. Full tests used `--test-concurrency=2`; smoke ran serially, with original scripts, timeouts, `--clearenv` and network isolation unchanged. An earlier simultaneous test/sandbox run hit an existing 30ms socket-fixture wait and a 20s smoke deadline; unchanged code passed the serial checks above.

All **869 candidate files** matched the tested Linux snapshot. Git tree: `572f5c7a0192086b4180b9c47263580bc5711426`.

Hosted validation of `13b68d7` is complete: [Tests run 34763237171](https://github.com/nicobailon/pi-subagents/actions/runs/34763237171) passed all five jobs (Ubuntu, Windows, Bun, Pi clean-install and official standalone). Greptile identified no blocking code issues on the same head. Downloaded hosted receipts confirm all 18 official cases and npm network isolation; all 304 recorded execution inputs match the local validation.


nicobailon:
Thanks, @qsgy-edge. The cold-start defect is real, and selecting native stripping for source-layout Node is the right direction. One compatibility blocker remains before merge:

- `nativeTypeScriptSupport` currently checks `"typescript" in process.features`. Node 22.14 exposes that property as `false`, so the native path is selected even though `node:module.registerHooks` is unavailable; `runner-peer-preload.mjs` then fails before the startup handshake.
- Please gate the native path on the actual supported capability/version (Node 22.19+ for this preload) and retain Jiti below it. Add coverage proving Node 22.14 takes the fallback and Node 22.19+ takes the native path, rather than repeating the property-presence predicate.

#2219 should land first because it fixes the event-loop-blocking handshake correctness issue. After that, this PR needs rebasing/composition with #2219 and exact-head validation, including `async-spawn-preload`, revival startup tests, npm-packed fallback, and standalone CI. The current `official-standalone` check is also red, so the final head must return green. No external branch changes were made.

nicobailon:
Landed through #2224 as `d7ed66b`, `135d2c7`, and `2d99331`, preserving @qsgy-edge’s two commits, authorship, timestamps, subjects, order, and visible changelog credit. The maintainer commit adds capability-based native selection, the older-Node alias loader, configured-extension Jiti boundary, and automatic Node 22.14/22.19 loader CI. Thank you @qsgy-edge.

PR #2223, opened 2026-09-13
fix(launch): inherit ambient tools in foreground children

alchemistklk:
## Summary

- let native foreground children discover ambient extensions when an agent omits `extensions`
- preserve `extensions: []`, explicit extension lists, and capability ceilings as authoritative overrides
- align missing-tool diagnostics and documentation with the shared foreground/background behavior

## Motivation

A foreground child (`async: false`) currently loses tools registered by normal Pi extensions even when the agent does not restrict `extensions`. The equivalent background child inherits those extensions. This makes otherwise shared tools unavailable based only on the execution host.

Foreground children still initialize extensions inside the parent process, against the child session registry. Extensions with process-global side effects can opt out with `extensions: []` or use an explicit list.

## Validation

- `npm run typecheck`
- focused child launch/tool-plan unit tests: 20 passed
- focused missing-tool foreground integration tests: 2 passed
- full integration suite: 1064 passed, 6 skipped; one deadline-checkpoint case timed out on the first run and all 3 cases passed on focused rerun
- full unit suite: 3349 passed, 13 skipped; one unrelated macOS process-group cleanup case fails locally with `kill EPERM`


nicobailon:
Closing as unsafe on the current Pi extension API, with thanks and credit to @alchemistklk for identifying the foreground/background tool-availability mismatch.

The proposed default re-runs every ambient extension factory inside the live parent process. In the concrete pi-subagents case, a foreground child has no detached-runner child marker, so ambient rediscovery registers the unrestricted root `subagent` runtime even when nested delegation is not authorized. That bypasses inherited depth, fanout budget, child management restrictions, and capability ceilings. Filtering only pi-subagents would not solve the broader issue: arbitrary ambient factories can still mutate parent-process hooks, providers, commands, timers, watchers, sockets, environment, and global registries, while tool allowlists do not sandbox those effects.

The supported API exposes tool metadata, not a safe executable-definition binding seam with child-specific context. Revisit when Pi provides one of:

- supported binding of selected executable tool definitions into a child without rerunning provider factories, with exact context and ceiling enforcement;
- an enforceable child-safe/repeatable extension manifest and lifecycle-isolation contract; or
- a separate process boundary.

Any future implementation must prove with real loader tests that the root delegation runtime cannot register in a child, authorized fanout retains depth/ceilings/required-tool/shared-budget rules, omitted/empty/explicit semantics remain distinct, and parent-only handlers/global state are not duplicated. If such a mechanism lands later, @alchemistklk should retain visible changelog credit for #2223.

PR #2228, opened 2026-09-13
fix: stop progress_update from waking the parent

moofone:
Edit Withdrawn, will resubmit


moofone:
Opened against upstream by mistake; not submitting this.

PR #2238, opened 2026-09-14
feat: propagate trace parent to child sessions

deestax:
## Summary
- capture an immutable trace-parent envelope for foreground and deferred child runs
- expose the envelope through a child session hook without coupling to a tracing SDK
- roll back a newly allocated async-capacity slot when owner persistence fails

## Validation
- `npm run typecheck`
- `npm run test:unit`
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/in-process-child.test.ts`

`npm run test:integration` has six pre-existing `/run` slash-command failures in this checkout; the focused child-session integration test passes.


nicobailon:
Thanks for the work here, @deestax. I read the branch against `VISION.md` and I'm closing this one as not planned. Reasoning, so it's checkable:

**Fit.** The envelope is captured from a Langfuse-specific global (`Symbol.for("pi.langfuse.contexts.v1")`) and exposed through a new per-child hook. It touches every launch path (foreground, async chain, async single, revival, runner steps) and adds per-launch capture, validation, freezing and hook registration. VISION asks for a demonstrated need before a niche integration, and for measured cost before adding work to launch paths. There is no open or closed issue asking for trace-parent propagation, and no measurement in the PR. That's the bar for a change of this shape, and it isn't met yet.

**Branch state.** The branch is 194 commits behind `main` with semantic conflicts in `child-launch.ts`, `subagent-runner.ts` and `active-async-capacity.ts`, so even a "yes" would need a fresh implementation rather than a rebase.

**The bundled capacity fix.** Rolling back a freshly created slot directory when `owner.json` fails to write is a reasonable standalone change. `main` currently keeps that directory on purpose (fail closed rather than risk double admission), but your argument holds: the directory was created exclusively by this process and no handle exists yet. If you want to send that as its own small PR against current `main` (with a test that proves the original error is preserved and the slot is reusable), I'll review it, and the changelog entry will credit you.

If someone opens an issue with a vendor-neutral trace context need and a way to bound the per-launch cost, this can be revisited from that issue.

PR #2255, opened 2026-09-14
fix: preserve rejected report content in diagnostic artifacts

pgoodjohn:
## Summary

Preserve the resolved report payload in foreground diagnostic output artifacts when explicit acceptance rejects a completed child with saved output.

The rejection path retains the canonical saved file and its receipt, but currently copies the compact `finalOutput` back into `artifactOutputByResult`. In file-only mode this overwrites the diagnostic report with an `Output saved to: ...` pointer. Removing that overwrite keeps diagnostic content separate from presentation without changing the rejection status, saved file, or output reference.

## Regression coverage

- File-only rejection retains canonical report bytes, the correct saved-output path/reference, and diagnostic report content.
- Inline rejection retains the resolved report content in the diagnostic artifact.
- Tests use scripted fixtures without model calls or network access.
- The new regression failed before the source fix (diagnostic output contained only the receipt) and passed afterward.

## Validation

This PR is a **draft** because full validation is not green in the local sandbox.

Passed:
- `npm run typecheck`
- `node --experimental-strip-types --import ./test/support/register-loader.mjs --test test/integration/acceptance-file-report.test.ts` — 13/13 passed
- `git diff --check`
- Independent read-only code review found no correctness or scope issues; publication/merge confidence remains limited by full-suite validation.

Full suites:
- Untouched upstream baseline and patched unit suites both produced 3,290 passed, 47 failed, 24 cancelled, and 13 skipped. Their failing-test-name sets matched. Baseline: `47bae7f78d4a6993fd620963a9ecc1aceff4802f`.
- Latest full integration run completed in 127.7 seconds: 1,041 passed, 28 failed, 0 cancelled, 6 skipped (exit 1).
- Errors include `spawnSync ps EPERM` and `Process-tree cleanup failed: verification-failed.` Integration failures have **not** been baseline-compared; this PR does not claim all are pre-existing.

Please use upstream CI or an environment with normal process visibility for the remaining validation before merge.

## Scope

One production-line removal plus regression coverage. No pointer-only response validation, path-following behavior, public API changes, or installed-package updates.


nicobailon:
Landed on `main` via #2257, which carries your commit with authorship preserved plus a maintainer changelog entry crediting you. Upstream CI on your exact head was green, so the local-sandbox limitation did not block it. Thanks, @pgoodjohn.

pgoodjohn:
my pleasure @nicobailon ! thx for always merging so swiftly :)

PR #2256, opened 2026-09-14
fix: Retry premature stream termination

ghostwriternr:
Pi reports `Anthropic stream ended before message_stop` when a stream begins but closes before its terminal event. Node's fetch stack can report the exact `terminated` message for the same request-level failure. Neither currently gives an unused fallback model a chance to recover.

Treat these as transient failures before tool activity, without persisting a 24-hour model exclusion. Keep the matching narrow so process-control messages, status-prefixed errors, and tool failures do not become model retries.

Pi error provenance: https://github.com/badlogic/pi-mono/blob/7dc1bed4785694f6c6d1d8e9c63e115802aed184/packages/ai/src/providers/anthropic.ts#L415-L417

Tests:
- `npm run typecheck`
- `npm run test:all`
- `node --experimental-strip-types --import ./test/support/isolated-temp-root.mjs --test test/unit/model-fallback.test.ts` after rebasing onto current `main`


nicobailon:
Landed on `main` via #2258, which carries your commit with authorship preserved plus the changelog credit. Thanks, @ghostwriternr.

PR #2261, opened 2026-09-14
fix: avoid model fallback for provisioning failures

moofone:
## Summary

Make `isRetryableModelFailureAttempt` reject the existing provisioning-error classification before advancing to another model. A preflight/npm failure can contain broad model-error wording without representing a provider failure; changing models does not repair that environment.

This is a one-line production guard shared by foreground and background model-attempt loops. Ordinary provider fallback retains its existing behavior.

## Regression coverage

- Unit test covers preflight and npm provisioning signatures.
- Integration test asserts that a failed preflight preserves its diagnostic, returns failure, and records exactly one attempt and one child launch even with a fallback configured.
- Independently reproduced behavioral RED without the guard: unit `true !== false`; integration `0 !== 1` because the fallback response ran.

After merging upstream `c5a473bf`:
- 213 unit tests passed (model fallback, async resume, stale reconciliation, run status, maintenance lifecycle, nested-control fanout).
- 258 integration tests passed (single-execution part 2 and async-execution part 3).
- `npm run typecheck` and `git diff --check` passed.

Async applicability is supported by the shared helper call site and existing background fallback regression tests; the new provisioning integration fixture is synchronous.

## Scope

This draft prevents an internal model-fallback retry for recognized provisioning failures. It does **not** prevent a parent from issuing a new task after runner failure, introduce persistent retry admission state, or change stale-run reconciliation/supervision. Unrecognized infrastructure diagnostics retain the previous classification behavior. No Pi core change is needed.

Please keep this PR as a draft; merging/deployment has not been authorized.


nicobailon:
Thanks for the focused report and implementation. I moved this to #2264 so the provisioning exclusion remains intact while same-launch model fallback is suppressed. Your original commit and changelog credit are preserved there.

PR #2262, opened 2026-09-14
fix: avoid routine intermediate workflow child wakes

moofone:
## Summary

Avoid triggering a parent inference turn for an ordinary successful intermediate workflow-child settlement while the workflow is still running. Deliver the same visible/context notice with `triggerTurn: false` instead.

Actionable failure/paused/stopped outcomes, terminal child barriers, standalone completions, and schedule-aware notification policy retain their existing behavior. No polling loop or Pi core change is introduced.

## Regression coverage

- Added an async workflow integration test that observes both actual child-notification sends, requires both to be false-trigger, observes one persisted workflow-completion wake through the notifier/result watcher, and replays completion to verify deduplication.
- Added policy tests for actionable failures and terminal child barriers.
- Reproduced a behavioral RED against pristine upstream: `ordinary running child settlements must not trigger provider turns`.

After merging upstream `c5a473bf`:
- 89 unit tests passed (notify, async-resume, maintenance lifecycle, nested-control fanout).
- 162 integration tests passed (single-execution part 1, result watcher, workflow result publication).
- `npm run typecheck` and `git diff --check` passed.

Separately characterized real Pi AgentSession behavior with a faux provider at `ceea48f5`: false-trigger messages do not initiate inference while idle or streaming and remain available to a later prompt. That temporary characterization is not part of this patch. During streaming, context-only messages are not included in the immediately following continuation request; this is existing Pi behavior.

## Scope

This draft covers routine intermediate success wakes only. It does not suppress actionable failures or claim measured live token savings. Please keep this PR as a draft; merging/deployment has not been authorized.


nicobailon:
Thanks @moofone. The credited maintainer replacement #2273 has merged, preserving Greg Tiller’s authorship and visible CHANGELOG credit for #2262 and #2265. Closing this original as superseded by the landed replacement.

PR #2264, opened 2026-09-14
fix: retain provisioning exclusions without model fallback

nicobailon:
## Summary

- avoid advancing to a fallback model when a child fails during recognized environment provisioning
- retain the existing 15-minute exclusion so later launches skip the failing model
- cover ordinary foreground, permitted workflow-child, and detached background launches

This replaces #2261 with the cache regression fixed. Greg Tiller's original commit remains attributed to him, and the changelog credits [@moofone](https://github.com/moofone).

## Validation

- model-fallback unit tests: 119 passed
- foreground provisioning integration tests: 2 passed
- detached background provisioning integration test: 1 passed
- `npm run typecheck`
- `git diff --check`
- fresh read-only review: clean


nicobailon:
Closing this in favor of a simpler hard cutover: remove automatic model fallback and persistent model exclusions entirely. A child will resolve one model, launch once, and return the real result. Greg Tiller’s contribution in #2261 helped expose how much state and cross-run coordination this subsystem requires; thank you for the careful work.

PR #2069, opened 2026-09-09
fix: wait for remembered detached foreground descendants before settlement

nicobailon:
This is a maintainer-owned replacement for #2051. It ports the reviewed nested-reviewer completion seam onto current `main` without pushing to the contributor branch.

Original reviewed head: `e4ac176644e379d9f45db017458aad320c5f049b` by [@shaharmor](https://github.com/shaharmor).

## Why this replaces #2051

The fork PR is unmodified and now conflicts with `main`. This branch started from that exact reviewed commit, then cut the extra launch/replay/cancel/reconciliation surface so only the settlement seam remains.

#2051 stays open. This PR does not close it.

## Accepted contract

- A parent does not complete while a remembered detached foreground descendant is still in flight.
- `hasOutstandingWork` and `bg_wait({ all: true })` see those descendants.
- A result-bearing child is not marked failed only because a descendant is still settling.
- `hasUI` does not skip drain for a normal in-process child session.
- Default-deny nested subagent grants are unchanged. Child registration still requires explicit fanout authorization.

## Runner grace window

Pi starts `FINAL_STOP_GRACE_MS` on the last assistant `message_end`. Plain `agent_end` now cancels that window, and the in-process runner reschedules it while `drainOutstandingWork` is held, so a result-bearing child is not aborted or force-finished while a remembered descendant is still settling.

## Credit

Thanks to [@shaharmor](https://github.com/shaharmor) for #2051.

## Validation

CI unit tests failed on `270a223e544e9bbb1ca6eb42c49b290e09a2227d` with one test on Ubuntu and Windows:

- `rejects runtime and wait accessors without invoking getters during admission or revalidation`

Cause: factory-time `{ ...config }` in `childHooks` invoked enumerable getters during admission. Hook factories now inherit the reviewed config and attach `runtimeState` / `holdFinalDrain` without spreading.

Local after the fix:

- Focused drain + readonly suite: 136 passed
- Full unit suite: 2999 passed, 14 skipped, 0 failed

Exact head: `5fb38df0e42a10fef0203d2b02179bb66cb72273`

[...]

Issue #2266, opened 2026-09-14
Nested async workflow children become unresolvable: mission-only directories shadow status, and bg_wait excludes nested runs


esguard:
## Summary

With `pi-subagents` **0.67.0** on Linux, a coordinator launching explicit async children through a mission-backed workflow receives valid child IDs, but cannot reliably retrieve or wait for those children.

The children actually run and produce reports. Their status files exist in the nested artifact directories, while public status lookup reports:

```text
Status file not found.
```

An exact-ID wait reports:

```text
No active run matched "<child-id>". Nothing to wait for.
```

This appears separate from #2265, which concerns parent wake notifications and result-publication failures.

## Observed launch pattern

- A coordinator has nested delegation enabled with sufficient depth and spawn budgets.
- It launches a workflow containing explicit `async: true` reviewer children.
- The workflow returns launch IDs and artifact directories.
- The coordinator waits for and inspects those exact child IDs.

Representative inner workflow shape:

```javascript
const results = await runs.all([
  {
    key: "review-a",
    label: "Review fixture A",
    agent: "reviewer",
    async: true,
    task: "Inspect the assigned local fixture without editing files.",
    output: "review-a.md"
  },
  {
    key: "review-b",
    label: "Review fixture B",
    agent: "reviewer",
    async: true,
    task: "Inspect the assigned local fixture without editing files.",
    output: "review-b.md"
  }
]);

return results.map(result => ({
  runId: result.runId,
  outputReference: result.outputReference,
  artifactPaths: result.artifactPaths
}));
```

This is a reduced outline of the observed setup, **not yet a separately validated minimal reproducer**.

## Actual behavior

For all five affected children, two directories existed:

```text
<temp>/async-subagent-runs/<child-id>/
└── mission.json

<temp>/nested-subagent-runs/<root-id>/<child-id>/
├── mission.json
├── status.json
├── process-terminal.json
└── ...actual runtime artifacts
```

The mission bindings in each pair were identical. The top-level directories contained only mission metadata.

All five children eventually produced reports. However, their coordinators could not obtain their actual completion through the intended wait/status path.

## Source findings

The relevant behavior also remains in the inspected `main` snapshot, `83af77e6dfd4bca8e884712dd4dfd252fd4c1c44`:

[...]

PR #2267, opened 2026-09-14
fix: let nested reviewers collect their async results

shaharmor:
## Summary

Nested reviewers can now wait for and consume their asynchronous persona reviewers' results instead of receiving “No active run matched” while status shows those runs running. Discovery stays limited to the immediate owning session, including when sibling coordinators share a root workflow.

This covers the storage-discovery gap left after the earlier nested-reviewer lifecycle work. Existing detached-foreground draining and supervisor handling remain intact. Automatic model reconciliation and cancellation changes are out of scope.

## Important boundaries

- The additional storage scope comes only from the validated inherited nested route; ordinary workflow storage remains supported.
- Exact-ID and prefix waits recover already-terminal runs, including personas that finish before the first wait. Result-file or replay-archive references are returned in model-visible content, not just tool details.
- Initial status is indexed before launch returns. The native test exposed a second race where aggregate wait and drain could otherwise miss a run until its runner's first update.
- Session ownership is checked before stale-run reconciliation. Session-scoped queries now leave foreign runs untouched in both ordinary and nested storage; unscoped status reconciliation is unchanged. This uniform boundary was explicitly approved during implementation.

Related: #2051, #2069, #2087

## Validation

Local validation on macOS with Node 24.15.0:

| Check | Result |
| --- | --- |
| Full unit suite, `--test-concurrency=2` | 3,400 passed, 13 skipped, 0 failed |
| Full integration suite | 1,075 passed, 7 skipped, 0 failed |
| Focused wait, completion, drain, runtime, status, and reconciliation suites | 179 passed |
| Native Pi 0.85.1 nested reviewer integration | Passed |
| Typecheck and diff check | Passed |
| Repository-wide lint compared with base | 0 new diagnostics; 11,414 diagnostics on both base and branch |

The native integration scripts only the model provider. Real AgentSessions, fanout registration, detached runners, waits, storage, and the read tool are exercised. One reviewer arm waits for two personas with `all: true`; another uses exact-ID and terminal-prefix waits. Both arms read persona findings, and the outer reviewer reads and consumes both arms' findings through ordinary workflow storage.

Focused coverage includes mixed storage, foreign sessions and siblings, invalid inherited routes, active-to-terminal transitions, pre-wait completion, result/replay references, timeout, supervisor attention, and session changes. The focused and native tests were rerun after final cleanup.

## Remaining validation

This is a draft pending review and platform CI. A native Pi 0.85.1 job is added for macOS and Windows; native Windows execution has not been verified locally. No live GPT/Claude cross-model review was performed. Earlier high-load timing failures passed on rerun; the missing local lint dependency was installed without changing package manifests or the lockfile.


nicobailon:
Thanks @shaharmor. The credited maintainer replacement #2272 has merged, preserving your authored implementation and visible CHANGELOG credit for #2267 and #2266. Closing this original as superseded by the landed replacement.

PR #2268, opened 2026-09-14
fix(test): reap owned background runners

nicobailon:
## Summary

- register every test-created detached runner root with its run ID, instance ID, process, and async directory
- accept exact `ChildProcess` close as natural runner-root cleanup; never signal its recorded PID or process group afterward
- when a registered runner root remains active, reap its owned process tree and require observed termination before deleting lifecycle evidence
- use identity-bound native process handles for exact Windows forced cleanup without PID-reuse races
- report actionable run and process evidence when forced cleanup cannot be proven

## Why

Integration fixtures could remove temporary directories while detached runner processes were still active. Those orphaned runners accumulated across broad test runs and consumed substantial CPU.

The fixture now owns every runner root it starts and proves root cleanup before deleting its artifacts. This scope does not claim containment of independently detached descendants after a runner root has already exited. Product `processTerminal` evidence remains separate and may truthfully report `unknown`; ordinary runtime behavior is unchanged.

## Validation

- `npm run typecheck`
- focused lifecycle unit tests: 18 passed, 1 Windows-only test skipped on macOS
- focused lifecycle integration tests: 2 passed
- intentional `failure-before-JSON` lifecycle test: passed with product terminal state still `unknown`
- focused external CLI tests: 34 passed
- exact registered runner-root survivor count: 0
- fresh bounded-scope review: clean

Windows CI exercises the real PowerShell/C# process-handle path and persistent parent/child forced-cleanup test.


nicobailon:
Closing this as not planned. The branch-specific Windows runs repeatedly reached the workflow timeout, and the failures showed that exact cross-platform test-runner ownership would require substantially more native process-supervision machinery than this test fixture justifies. The production cleanup protections already merged in #2074 and #2217 remain unchanged. We will use focused local validation and ephemeral CI runners for broad suites instead.

PR #2269, opened 2026-09-14
feat: gate inspector.open and project.open behind authorityPolicy

kevthedawg:
`inspector.open` and `project.open` both spawn an external process and neither checks `authorityPolicy`.

`inspector.open` at least has to get past a plugin's `available()`, which means being in Ghostty or inside a Herdr pane. `project.open` has nothing: the dispatch never passes plugins, `open()` in `project-panes.ts` calls `detectHerdr()` unconditionally, and that spawns `HERDR_BIN` or `herdr` from PATH on any platform in any terminal. The pane it opens hosts its own Pi session.

This adds `inspectorOpen` and `projectOpen` to `AUTHORITY_ACTIONS` and moves the existing policy gate in `subagent-executor.ts` above the two dispatches. Same `forbid`/`confirm`/`auto` handling and confirm prompt as `stop`, `steer` and `schedule.create`.

`inspectorOpen` defaults to `auto`. The plugin gate already limits where it can fire, and confirm-required actions fail closed without a UI, so `confirm` would break it in every headless session.

`projectOpen` defaults to `confirm`. That is a behavior change for anyone using `project.open` today, and I think it should be: it is the one action with no gate, and a pane running its own Pi session is closer to `spawnBudgetGrant` than to `stopRun`. `"projectOpen": "auto"` gets the old behavior back. If you want this strictly additive, I will flip it.

The fleet TUI is untouched; opening an inspector there is already a keypress.


kevthedawg:
Good catch, fixed in b3e873f.

The gate now refuses first for every action it covers. That also picks up `schedule.create`, which had the same ordering against its own check when `scheduleCreate` is set to confirm or forbid. `worktree.discard` was already in this order.

The three new tests fail on the previous commit.

nicobailon:
Landed in #2277 as 13f8f3286c7ce42d0aecfd72bca76b26912b1445 with your authorship preserved through the merge co-author trailer and visible changelog credit. Thank you, @kevthedawg.

PR #2271, opened 2026-09-14
feat: record structured delegations in a host-owned in-process ledger

tayiorbeii:
## Motivation

Structured delegation consumers currently cannot prove that the host actually ran a child. Everything a consumer can check lives in the project tree (`.pi-subagents/` artifacts, plan receipts, attestation files), and every one of those locations is writable by any tool the delegated session can reach. A bash-capable child (or a compromised prompt in the delegating session) can forge run-identity evidence — run id, agent, child index — and no file-based check can distinguish that from a real run.

We hit this concretely in pi-plan: review attestations were bound to a host-written persona-team `_meta.json`, and the binding was forgeable end-to-end by a planner with shell access.

## What this change adds

The prompt-template delegation bridge — the single seam every structured delegation passes through — now records delegation lifecycle into a **host-owned, in-process ledger** exposed as a new supported package export, `pi-subagents/delegation-ledger`:

- `recordDelegationStarted({ requestId, ownerRunId, nodeId, agent })` — called when a structured request is accepted (after duplicate/settled guards, so replays do not refresh records)
- `recordDelegationTerminal({ requestId, ownerRunId, nodeId, status, runId?, launchContractDigest? })` — called from `emitTerminal`, merging into the started entry; `agent` comes from the accepted request (kept per attempt key) so early terminals (`cancelled`, `unavailable_context`, `duplicate_node`) still carry identity
- `getDelegationRecordByRun(runId, { agent })` / `snapshotDelegationRecords()` — read API for consumers
- `DELEGATION_LEDGER_VERSION = 1`, `DELEGATION_LEDGER_KEY = "pi-subagents.delegation-ledger.v1"` — a well-known `Symbol.for` key so every module instance in the host process shares one registry, the same cross-extension pattern as the external-run registry

`invalid_request` responses are intentionally not recorded: a parse failure is not a delegation.

## Design decisions

- **Fail-closed validation** of every entry (bounded ids without line breaks, safe-integer timestamps, whitelisted terminal statuses), mirroring the external-run registry.
- **Bounded** (`maxEntries: 256`, oldest-evicted) so long-lived host processes cannot leak.
- **Best-effort recording**: the bridge wraps ledger calls so a ledger failure can never break a delegation in flight.
- **Standalone terminals** are recorded even without a started entry (bridge restart mid-run), so completed-run evidence cannot be silently erased.

## Threat model, honestly

The ledger lives only in host-process memory, which the delegated session's tools cannot write. That makes it an authoritative **save-time** root of trust: a consumer can ask "did the host really run this child with this identity?" and get an unforgeable answer at the moment evidence enters downstream state (e.g., a saved plan receipt). It is deliberately **not** a cross-process cryptographic log — the host process is the trust boundary here, same as the rest of the extension layer.

## Tests

- `test/unit/delegation-ledger.test.ts` (new): started+terminal merge, standalone terminals, agent-scoped lookup misses, fail-closed validation (empty ids, newlines, `running` as terminal status, negative child index), bounded eviction, malformed persisted registries (`Malformed`/`Unsupported`).
- `test/unit/prompt-template-bridge.test.ts`: completed delegations produce a completed ledger record with `runId` + `launchContractDigest`; early `unavailable_context` terminals record the requested agent without a `runId`; `invalid_request` never reaches the ledger.
- `test/unit/package-manifest.test.ts`: `./delegation-ledger` added to the expected exports map, plus entrypoint assertions (version, key, function surface) matching the other API exports.

Results: `npm run typecheck` clean; targeted suites green; full unit suite 3393 passed with 3 failures that also fail on clean `main` on this machine (Windows `APPDATA` npm-root fallbacks and a profiles test — environment-specific, reproduced with the patch stashed); `single-execution` integration parts 1–2: 307/307.

## Vision fit

This is delegation evidence infrastructure: it gives the operator (and downstream tools like pi-plan) a way to trust that delegation happened as recorded — "makes delegation more trustworthy for the person whose name it runs under" — with zero hot-path cost (a bounded Map insert per delegation) and no change to the wire contract.


nicobailon:
Thanks for the detailed proposal. Closing this as not planned. A second in-process delegation ledger would duplicate existing run lifecycle state without establishing a separate trust boundary: code running in the host process can reach the same process-global registry, while best-effort recording cannot be treated as authoritative proof. We do not want to add another lifecycle surface whose records can diverge from the run system it mirrors.

PR #2274, opened 2026-09-15
fix: inherit the parent's registered providers in foreground children

lallenlowe:
## Summary

A foreground child (`async: false`) never loads the parent's ambient extensions, so its `ModelRuntime` only knew Pi's built-in providers. Every model from an extension-registered provider (`pi.registerProvider`) failed with Pi core's `Model "…" not found`, including the builtin `worker`/`reviewer`/`delegate` agents on a parent whose active model comes from such a provider. #2249 explains that failure; this makes it not happen.

The extension records `ctx.modelRegistry` at `session_start`. A child launched without ambient extensions then registers any provider its runtime lacks, from the parent's registered config or native provider, and refreshes offline before `resolveCliModel` runs. Children that load the ambient extensions are unchanged, and a parent-side registration failure is reported through `onExtensionError` without blocking the launch.

The foreground diagnostic no longer sends operators to `async: true`, and the three docs that said foreground children cannot use provider-extension models now say the opposite.

## Reproduction

Pi 0.85.1, pi-subagents 0.67.0, parent model `bifrost/vertex/gemini-3.8-flash` from a provider registered by an ambient extension.

- `subagent({ agent: "worker", task: "…", async: false })` → `Model "bifrost/vertex/gemini-3.8-flash:high" not found. Use --list-models to see available models.` on every launch.
- Same call with `async: true` → succeeds.
- `pi --model bifrost/vertex/gemini-3.8-flash:high -p hello` → succeeds, so the id and suffix are fine.

With this change, foreground launches on that model succeed (4/4 single + parallel `runs.all`), and background is unchanged.

## Validation

- `npm run typecheck`
- `npm run test:unit` (3262 pass, 0 fail)
- `test/integration/model-resolution-diagnostic.test.ts` + `test/integration/in-process-child.test.ts` (31 pass)
- `git diff --check`


lallenlowe:
Thanks all three addressed.

**1. Parent overrides skipped (P1)**: Fixed. Dropped the `known.has()` skip and moved `inheritParentProviders` *before* the child's own `flushQueuedProviderRegistrations`. Precedence is now built-in < parent < child-explicit: a parent extension's override of a built-in id (e.g. `anthropic`) reaches the child, and a provider the child itself registers still wins on conflict (last write). New test asserts a parent `anthropic` override is registered even though the child runtime already knows that id.

**2. Refresh failure aborts launch (P1)**: Fixed. The `refresh({ allowNetwork: false })` is now wrapped; a rejection is reported via `onExtensionError` (`<parent-provider:refresh>`) and `resolveCliModel` still runs. New test drives a throwing refresh and asserts the error is reported and model resolution proceeds.

**3. Missing credit (P2)**: Added `Thanks to @lallenlowe for #2274` to the changelog entry.

typecheck + focused unit (17) + integration (`model-resolution-diagnostic`, `in-process-child`, 31) all pass.

nicobailon:
Landed in #2276 as e32028b07fd402ec9e70ab6750a55f28ce410f33 with both of your authored commits preserved through the explicit merge co-author trailer and visible changelog credit. Thank you, @lallenlowe.

PR #2294, opened 2026-09-16
Fix Orca progress-tab viewer crash under fish (single-quote backslash collapsing)

yourfriendaaron:
## Problem

The Orca progress-tab viewer is a single-quoted POSIX shell string, but Orca runs `terminal create --command` through the user's login shell. fish collapses `\\` inside single quotes to `\`, so the two backslash comparisons in the viewer script reached `node -e` as `char==='\'` — an unterminated string literal. Every progress tab therefore opened on a `SyntaxError` instead of the progress mirror. Affects any fish user with `orcaProgressTabs` enabled (present in 0.67.0 and 0.68.0).

## Fix

Remove all backslashes from `VIEWER_SCRIPT` by comparing against `String.fromCharCode(92)`:

```diff
+	"const BACKSLASH=String.fromCharCode(92);",
-	"if(state==='osc-escape'){state=char==='\\\\'?'text':code===27?'osc-escape':'osc';continue}",
+	"if(state==='osc-escape'){state=char===BACKSLASH?'text':code===27?'osc-escape':'osc';continue}",
-	"if(state==='string-escape')state=char==='\\\\'?'text':code===27?'string-escape':'string';",
+	"if(state==='string-escape')state=char===BACKSLASH?'text':code===27?'string-escape':'string';",
```

Those were the only backslashes in the script; the viewer command now contains none. The `'"'"'` quoting idiom (`shellQuote`) is unchanged and fish-safe, and its other use (the cleanup hint) only quotes plain paths.

## Verification

- fish collapses backslashes inside single quotes; POSIX shells do not. Same literal, different program:
  - `fish -c "printf '%s\n' 'a\\\\b'"` → `a\b`
  - `sh -c "printf '%s\n' 'a\\\\b'"` → `a\\b`
- Extract-and-run harness: built the viewer command from the patched source for a fake mirror, ran it under `fish -c` against a log with ANSI/OSC/ST sequences (including ESC `\` string terminators, the branches that changed) plus a `.done` marker. Result: script parses, sanitizer output is intact, both mirror files are deleted, exit code 0. The same command still passes under a POSIX shell, and the unpatched source reproduces the `SyntaxError` under fish.
- Regression test asserts the generated `--command` contains no backslash character.
- `npm run typecheck`, unit and Orca integration tests pass.


nicobailon:
Superseded by #2331, which preserved your exact contributor commit `7265b0f9195fc2494c92f0d86f20a7cf31a5a040`, added the required OSC/DCS coverage, passed the full exact-head CI and current-head review gates, and merged as `8744268fc7cd3aceb8e5478039a4b4ab289a479f`. Contributor credit remains in the changelog. Thank you, @yourfriendaaron.

nicobailon:
Correction: GitHub marked this original PR as merged when its preserved head became reachable through #2331. No closure was needed. The exact contributor commit and authorship remain in the landed history; #2331 supplied the follow-up coverage and gate evidence.

Issue #1666, opened 2026-08-28
subagent tool never reports child Usage — session totals undercount delegated turns


jf88888:
## Summary

Subagent token usage is never reported back to the parent session. The `subagent`
tool result carries no `usage`, so a delegated turn silently undercounts both
tokens and cost — in pi's footer, in `/session`, and in the RPC
`get_session_stats` totals.

`docs/extensions.md` states the contract this misses:

> **Usage accounting:** If a tool makes nested LLM calls, return their combined
> `Usage` as `usage`. Pi persists it on the tool result and includes it in
> footer, `/session`, and RPC session totals.

## Reproduction

- pi 0.84.3, pi-subagents 0.58.0
- provider `openai-codex`, model `gpt-5.4-mini`
- RPC mode (`pi --mode rpc`), full frame capture
- Prompt: *"Create 5 subagents in parallel. Each says one short random joke in a
  different language (French, Japanese, Malay, Spanish, German). Then list all 5
  jokes."*

## Observed

Every `tool_execution_end` for `subagent` / `subagent_wait`:

```
subagent      -> result.usage = null
subagent      -> result.usage = null
subagent_wait -> result.usage = null
subagent      -> result.usage = null
```

`get_session_stats` after `agent_end`:

```json
{ "tokens": { "input": 15750, "output": 585, "cacheRead": 48640, "total": 64975 },
  "cost": 0.018093, "toolCalls": 4, "assistantMessages": 5 }
```

The 5 children each wrote their own session file, which **does** contain full
assistant usage:

```
<sessionDir>/<sessionId>/<runId>/run-0/session.jsonl   x5
```

Summing those:

| | tokens | cost |
|---|---|---|
| Parent (reported) | 64,975 | $0.018093 |
| 5 children (not reported) | 14,012 | $0.014904 |
| True total | 78,987 | $0.032997 |

**17.7% of tokens and 45.2% of cost are missing** from every figure pi shows the
user. Five one-shot jokes is the cheap case; a `reviewer` or `oracle` pass on a
real diff makes the children dominate.

## Second, related problem: async results carry `usage: null`

On the async path the per-child usage is null in the tool details too:

```
subagent_wait -> details.completions[0].results[].usage = [null,null,null,null,null]
```

`buildSubagentCostReport` (`src/slash/slash-commands.ts:398`) reads
`result.usage` from session-entry details, so `/subagent-cost` also reports
nothing for an async run — the numbers never leave the child.

## Suggested fix

[...]

PR #2295, opened 2026-09-16
fix(foreground): reconcile child usage from session messages when live events carry none

riique:
Follow-up to #1666.

## Problem

A synchronous foreground subagent can do real work yet return a tool result with zero usage, so the parent footer and session totals never move. Reproduced end-to-end: single `scout` call, `async:false`, native agent — the child listed files and wrote its output, but the parent session gained only its own call overhead (+$0.00062) and the tool result carried no `usage`/`totalCost`. Same symptom on two different providers (Meta-ai and OpenRouter), which pointed away from model pricing and toward the usage pipeline.

Root cause chain:

- `withAggregatedToolUsage` covers every execution path, but only attaches `usage` when the summed child usage is non-zero.
- Foreground child usage is accumulated exclusively from live `message_end` events (`execution.ts` accumulator). For some providers those events carry no `usage` (in the reproduced case, not even `turns` was incremented), while the persisted session messages do carry usage — the parent footer itself is proof, since it aggregates persisted entries.
- The live accumulator also only reads pi's canonical field names, while the background path already tolerates provider variants (`input ?? inputTokens`, numeric `cost`).

## Fix

- After the child settles, reconcile `result.usage` from the live session messages (`session.messages`) when the session totals exceed the event-accumulated ones. A pre-prompt message-count baseline excludes history the child inherited (e.g. forked parent turns), and `progressSummary.tokens` is updated to match.
- Harden the live accumulator with the same tolerant field reading the background runner uses.

## Tests

- New `test/unit/foreground-usage-reconciliation.test.ts` drives the real `runSync` with a fake child session: (1) events without usage + session messages with usage → usage adopted; (2) events with usage → unchanged behavior; (3) inherited pre-prompt history → excluded. All 3 pass with the fix; case (1) fails without it (`{input:0, output:0, ..., turns:1}` — exactly the reported symptom).
- `tsc --noEmit` clean. Full unit suite: 3272 tests, only the 12 pre-existing environment failures (Windows symlink privileges, offline npm) — verified byte-identical failure set on pristine `main`, so no regressions.

## Local verification pending

Mirrored into the installed extension copy for a live re-test (requires Pi restart); will report back.


riique:
Live verification on the installed copy (Pi restarted after mirroring the patch): a synchronous single `scout` call now persists a `toolResult` entry with `usage` `{input: 3723, output: 573, cacheRead: 5602, cost.total: 0.000498104}` — the cost-breakdown-zeros/total-only shape produced by `toAgentToolUsage`, i.e. the reconciled child cost. Session delta for the test window was +$0.00100006 = parent call turn ($0.000228742) + reconciled child ($0.000498104) + report turn ($0.00027321). The pre-fix diagnostic on the same setup showed no `toolResult` usage at all and only parent overhead (+$0.00062). Models: Muse Spark 1.3 via Meta-ai (child inherits parent model).

riique:
Both greptile findings addressed in 99fc793d: terminal `progress.tokens`/`inputTokens`/`outputTokens` now sync with reconciled usage before the final snapshot, and the changelog entry carries contributor credit. `tsc` clean, reconciliation tests 3/3 pass.

nicobailon:
Superseded by the unified reconciliation implementation in #2335, merged as `8bd275bba0dc13273eff366e348378d41ad5535e`. The landed commit preserves @riique as author and includes changelog credit for both #2295 and #2296. It applies one field-wise post-baseline reconciler at the foreground and background terminal owners, with focused coverage and the full exact-head CI/review gates.

PR #2296, opened 2026-09-16
fix(background): reconcile step usage from the persisted child session at completion

riique:
Follow-up to #1662, #1666 and #2295.

## Problem

Background (async) children whose providers omit usage on live session events complete with `SingleResult.usage` at zero, so every collection path (`wait` terminal, `resume`-await, `workflowAwaitAsync`, workflow terminal) attaches zero and parent footer/session totals miss delegated turns — even though the runner already reconciles display tokens from the session file at completion (`parseSessionTokens`).

Verified by reading (not assumed): `resume`/`workflowAwaitAsync`/workflow-terminal returns all flow through `withAggregatedToolUsage`, so no change was needed there — the data they sum was simply zero. Likewise deliberately untouched: `status` (attaching would double-count on repeat polls with no consumption watermark), wait timeout/attention/yield partials (partial-then-terminal would double-count), the quiet completion notification, and FleetView/widget (display-only by design).

## Fix

- At background step completion, when event-accumulated step usage is empty, backfill `singleResult.usage` (in place — `results[]` holds the reference) and `totalCost` from a session-parse delta anchored at a step-start baseline, so inherited fork history stays out of the step delta. A null baseline (file absent at step start = fresh context) falls back to the full cumulative parse.
- New `parseSessionUsage` (full-fidelity sibling of `parseSessionTokens`: input/output/cache splits, cost, assistant turns, tolerant field names), `hasUsageValue` and pure/testable `backfillStepUsageFromSession` in shared utils.

## Tests

- New `test/unit/session-tokens.test.ts` (5 cases: sums, provider variants, top-level entry usage, malformed input, null cases) and `test/unit/background-usage-backfill.test.ts` (5 cases: fresh adopt, fork-history exclusion, events win, null/zero/negative handling) — 10/10 pass.
- `tsc --noEmit` clean. Affected suites (`subagent-wait`, `wait-completions`, `async-status-*`, `total-cost`, `async-resume`, `fleet-status`) show only the 2 pre-existing Windows-symlink failures, byte-identical set on pristine `main`; full-suite 13 failures likewise identical to pristine `main` (12 environmental + 1 stray-log artifact of the test run itself).


nicobailon:
Superseded by the unified reconciliation implementation in #2335, merged as `8bd275bba0dc13273eff366e348378d41ad5535e`. The landed commit preserves @riique as author and includes changelog credit for both #2295 and #2296. It applies one field-wise post-baseline reconciler at the foreground and background terminal owners, with focused coverage and the full exact-head CI/review gates.

PR #2302, opened 2026-09-16
fix(delegation): report blocked tool budgets as tool_budget_exhausted

kylerberry:
## Problem

A foreground structured-delegation request with `toolBudget: { hard: 0, block: "*" }` that requires a worker tool call produced a terminal response of `status: "completed"` instead of `"tool_budget_exhausted"`.

The runtime correctly blocked the attempted call (the child transcript shows `Tool budget hard limit reached after 1 tool call (hard 0). The 'bash' tool is blocked...`), the child ended normally, but the public terminal projection lost the blocked-call fact.

## Root cause

In `runSingleAttempt` (`src/runs/foreground/execution.ts`), the `tool_result_end` handler only marked `result.toolBudgetBlocked = true` when a `pendingToolResult` was set — and `pendingToolResult` is only set from a `tool_execution_start` event. When the runtime blocks the call before emitting an execution-start event, no `pendingToolResult` exists, so the blocked call was never recorded and `resolveSubagentDelegationStatus()` could not map the child to `tool_budget_exhausted`, even though the mapping (`child?.toolBudgetBlocked`) already exists.

## Fix

Detect the hard-budget block from the blocked tool-result text itself, independently of the execution-start bookkeeping:

- `result.toolBudgetBlocked = true` whenever a tool result contains the hard-limit message and a tool budget is configured.
- The blocked tool name falls back to the tool result's own `toolName` when `pendingToolResult` is absent.
- `toolBudgetState` is projected with `toolCount >= hard + 1` so `hardReachedAt` and `outcome: "hard-blocked"` are correct for the no-execution-start path.

This preserves the semantic rule: classify as `tool_budget_exhausted` only when the runtime actually blocked an attempted tool call — not merely because a budget was configured or the worker completed without attempting a tool. The child's final output is retained as internal partial evidence (`details.results[0].finalOutput`); no public response field is added. Existing terminal-status precedence (cancelled, timed_out) is untouched.

## Tests

- Unit (`test/unit/delegation-api.test.ts`): a foreground bridge result with `toolBudgetBlocked: true` maps to the sole terminal response `tool_budget_exhausted`.
- Integration (`test/integration/single-execution.part-1.test.ts`): a public structured-delegation run with `{ hard: 0, block: "*" }` and a blocked bash result **without any `tool_execution_start` event** yields `details.results[0].toolBudgetBlocked === true` with the final output preserved as partial evidence, and a normal no-tool delegation still completes without the flag.
- Cancellation and timeout paths are unchanged (existing suites).

## Verification

- `npm run typecheck` — pass
- `npm run test:all` — 1055 passed, 0 failed, 7 skipped

This is a bug fix, not a public API change: `tool_budget_exhausted` already exists in `SubagentDelegationStatus`, and `toolBudget` already accepts `{ hard: 0, block: "*" }`.


kylerberry:
Thanks — both comments addressed in a586ef7e.

**Comment 1 (substring match):** The foreground detector no longer trusts a substring. `isToolBudgetBlockedMessage` (in `tool-budget.ts`, beside the message generator it validates) requires the trimmed tool result to be exactly the runtime block message: full-format anchored match, embedded hard limit equal to the configured budget, embedded blocked count strictly greater than that limit, and the named tool equal to the result event's own tool name. A successful `grep`/log read that merely contains the phrase — even quoting the full line verbatim — no longer sets `toolBudgetBlocked` (regression test added). No tool name resolved fails closed.

On the larger point: I agree text inference is the weak link. The budget hook (`registerToolBudget` in `subagent-prompt-runtime.ts`) is the authority that performed the block, and a structured block signal from it through the child-session bridge would remove the spoofable/adventitious class entirely. That crosses the runtime event boundary, so I kept it out of this narrow fix — would you prefer it here or as a follow-up issue? One consistency note for that decision: the background runner (`src/runs/background/subagent-runner.ts:1394`) uses the same text heuristic today and should adopt whatever signal is chosen.

**Comment 2 (blocked tool attribution):** Fixed per your recommendation — `toolResultCompletion.toolName` is now authoritative, and `pendingToolResult?.tool` is only consulted when the result event carries no name. The regression test covers the exact overlap you described: a pending `read` call is open when the blocked `bash` result arrives without a start event, and the projection records `blockedTool: "bash"`.

Verification: `npm run typecheck` and `npm run test:all` (1056 passed, 0 failed, 7 skipped).

PR #2308, opened 2026-09-17
feat(fleet): show a live token average on active status rows

### LINKED ISSUE #2307: Show effective throughput (tok/s) on live Fleet status rows
### What

The below-editor Fleet status widget already renders `elapsed · tokens` on every live row, but not the rate between them.

Live rows show e.g. `worker-0 · running   11s · ↓ 13.1k tokens`. Nothing there tells the operator whether a child is moving or stalled, which matters most when several children run at once.

### Proposal

Derive the rate from data the row already carries and render it between the two existing fields:

```
worker-0 (fable-5 · thinking low) · running      11s · 1191 tok/s · ↓ 13.1k tokens
```

### Semantics (worth stating plainly)

The number is **cumulative tokens over wall-clock time**, so it includes thinking, tool calls, and provider waits. It reads below the model's decode rate. A true decode-rate metric needs per-request output-token timing, which nothing records today; this is the honest cheap version.

### Detail

- Hold the rate back until the row has run about five seconds. Tokens move once per turn, so earlier samples are noise.
- Skip it for external runs, project panes, and workflow wrapper rows, whose usage is attributed to child rows.
- Drop the rate rather than truncate the agent label on narrow terminals.

### Fit

No config value, flag, or existing primitive produces this, so it is not a second path to an outcome that already exists.

lamtran2601:
Closes #2307.

The below-editor Fleet status widget renders `elapsed · ↓ N tokens` on every live row but not the rate between them, so an operator watching several children cannot tell a moving row from a stalled one.

```
    worker-0 (fable-5 · thinking low) · running      11s · 1191 tok/s avg · ↓ 13.1k tokens
```

## What the number is

Cumulative input-plus-output tokens over wall-clock time. The numerator includes prompt tokens the provider read from cache rather than decoded, and the denominator includes thinking, tool calls, and provider waits, so it is a **lifetime average, not a decode rate**. The label says `tok/s avg` for that reason, and `docs/observability.md` states the same.

## Behavior

- Five-second floor before the rate appears. Tokens move once per turn, so earlier samples are noise.
- Running rows only, and never on workflow wrapper rows, whose usage is attributed to their child rows. External and project-pane rows carry no token totals, so the formatter's zero-token guard covers them.
- Under 0.1 tok/s the field is omitted rather than rendered as `0.0 tok/s`, which reads as stalled.
- The render drops the rate before it truncates the agent label. The unclipped measurement in the workflow coverage gate uses the same decision, so a narrow roster that drops the rate does not revoke coverage for a row that still fits.

## Evidence

- `npm run test:unit` — 3311 tests, +2 new, failing set byte-identical to baseline (14 pre-existing suites: Herdr inspector, supervisor channel, Orca/observer — unchanged by this diff).
- `npm run test:integration` — new coverage test passes; `result-publication` is flaky on this machine with and without the diff (I reproduced it on a stashed baseline).
- Mutation-checked: reverting the unclipped-path fix fails the new integration test; removing the rate from the rendered row fails the new unit test.

## Residual risk

Async step rows pair `step.startedAt ?? job.startedAt` with `step.tokens?.total`, so a step that restarts (retry or recovery) keeps its prior token total under a fresh `startedAt` and would show an inflated average. That pairing predates this diff — the elapsed column already uses it — so this reports it rather than changing it.


lamtran2601:
Closing this. I am keeping the change as a local patch on my install instead of shipping it upstream, so the branch and PR are no longer needed. Thanks.

